Issue #64, opened 2019-01-08
End to end checksum and bit rot.


kerneltime:
For upload and download of files we need to insure we are setting up the checksum correctly and validating any bit rot that might occur.

kerneltime:
Fixed in https://github.com/oneconcern/datamon/pull/104 and https://github.com/oneconcern/datamon/pull/145

kerneltime:
On upload we do a CRC for GCS and on download we check blake hash.

PR #102, opened 2019-04-03
cafs: verify hash functions on get

ransomw1c:
the idea here is that the`cafs` `Get` function should re-compute hashes upon retrieving data, just as they were computed on upload, and throw an error if the hashes don't match.


ransomw1c:
this might have to do with #64

ransomw1c:
fixed the `nil` error, moved the branch to this repo such that the circle ci passes, and opened #145 from that branch

PR #131, opened 2019-04-27
CSI support

kerneltime:
Adds

- [x] Support for read only bundles in k8s

- [x] Sample YAML to drive the implementation


deploy/controller.yaml
```diff
@@ -0,0 +1,75 @@
...
+            - "--csi-address=$(ADDRESS)"
+          env:
+            - name: ADDRESS
+              value: /csi/csi.sock
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+        - name: csi-external-provisioner
+          imagePullPolicy: Always
+          image: gcr.io/onec-co/csi-provisioner:v0.3.1
```
ransomw1c: noticed the provisioner is in the one concern docker repo while the attacher is using the upstream docker repo.. what's the reasoning here?  is provisioner the component that had the spurious logging output?

deploy/controller.yaml
```diff
@@ -0,0 +1,75 @@
...
+      serviceAccountName: datamon-csi-controller-sa
+      containers:
+        - name: csi-external-attacher
+          imagePullPolicy: Always
+          image: quay.io/k8scsi/csi-attacher:v0.4.1
+          args:
+            - "--v=4"
+            - "--csi-address=$(ADDRESS)"
+          env:
+            - name: ADDRESS
```
ransomw1c: as a further addition, what about using Ivan's [`envexpand`](https://github.com/oneconcern/depmon/blob/master/hack/envexpand.go) `hack/` [as in](https://github.com/oneconcern/depmon/blob/master/k8s/dev-pod.template.yaml) [`depmon`](https://github.com/oneconcern/depmon/blob/master/hack/dev-env/create_dev_pod.sh) to do simple templating?

this `ADDRESS` and the `CSI_ENDPOINT` used below could all use the same template value, and the env variable could be removed in favor of inlining the template variable with the `args` item.  inlining into `args` in particular would make it clearer that the environment variable isn't meant to be used elsewhere (if this is in fact the case).

deploy/controller.yaml
```diff
@@ -0,0 +1,75 @@
...
+            - "--nodeid=$(KUBE_NODE_NAME)"
+            - "--controller=true"
+            - "--log-level=debug"
+            - "--credential=/etc/datamon-creds/gcp-storage-admin.json"
+          env:
+            - name: CSI_ENDPOINT
+              value: unix:/csi/csi.sock
+            - name: KUBE_NODE_NAME
+              valueFrom:
+                fieldRef:
```
ransomw1c: note from [k8s docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#envvarsource-v1-core)

> `fieldRef` Selects  a field of the pod: supports metadata.name, metadata.namespace,  metadata.labels, metadata.annotations, spec.nodeName,  spec.serviceAccountName, status.hostIP, status.podIP.

pkg/csi/controller.go
```diff
@@ -10,45 +18,132 @@ import (
...
+	if repo == "" {
+		s.config.driver.config.Logger.Error("Repo not set", zap.String("name", req.Name))
+		return nil, fmt.Errorf("repo not set for req name:%s", req.Name)
+	}
+	hash := req.GetParameters()["release"]
+	if hash == "" {
+		s.config.driver.config.Logger.Error("Release not set", zap.String("name", req.Name))
+		return nil, fmt.Errorf("release not set for req name:%s", req.Name)
+	}
+	rd, err := core.GetRepo(repo, s.config.driver.metadataStore)
```
ransomw1c: so `CreateVolume` creates a volume corresponding to an _existing_ repo...

Makefile
```diff
@@ -34,12 +42,12 @@ help:
 ## Build datamon docker container (datamon)
 build-datamon:
 	@echo 'building ${YELLOW}datamon${RESET} container'
-	@docker build --pull --build-arg github_user=$(GITHUB_USER) --build-arg github_token=$(GITHUB_TOKEN) -t reg.onec.co/datamon:${GITHUB_USER}-$$(date '+%Y%m%d') -t reg.onec.co/datamon:$(subst /,_,$(GIT_BRANCH)) .
+	@docker build --pull --build-arg github_user=$(GITHUB_USER) --build-arg github_token=$(GITHUB_TOKEN) -t ${REPOSITORY}:${GITHUB_USER}-$$(date '+%Y%m%d') -t ${REPOSITORY}/datamon:$(subst /,_,$(GIT_BRANCH)) .
```
ransomw1c: the first tag param doesn't have the `datamon` image name

cmd/datamon/main.go
```diff
@@ -2,8 +2,20 @@
...
+)
 
 func main() {
+	f, err := os.Create("cpu.prof")
+	if err != nil {
+		log.Fatal(err)
+	}
+	_ = pprof.StartCPUProfile(f)
 	cmd.Execute()
+	defer pprof.StopCPUProfile()
```
ransomw1c: unclear what the intent is here:  either removing the `defer` or placing the `defer`red `Stop` immediately after the `Start` would be clearer.

pkg/core/fs_ro_ops.go
```diff
@@ -24,35 +24,130 @@ func (fs *readOnlyFsInternal) StatFS(
...
+	fs.l.Info("End",
+		zap.String("Request", reflect.TypeOf(op).String()),
+		zap.String("repo", fs.bundle.RepoID),
+		zap.String("bundle", fs.bundle.BundleID),
+		zap.Any("op", op),
+		zap.Error(err),
+	)
+}
+func typeAssertToFsEntry(p interface{}) *fsEntry {
+	fe, ok := p.(fsEntry)
```
ransomw1c: using the one-return-value form `fe := p.(fsEntry)` will panic without the need for the conditional on `ok`

pkg/core/fs_ro_ops.go
```diff
@@ -24,35 +24,130 @@ func (fs *readOnlyFsInternal) StatFS(
...
+	fs.l.Info("Start",
+		zap.String("Request", opType.String()),
+		zap.String("repo", fs.bundle.RepoID),
+		zap.String("bundle", fs.bundle.BundleID),
+		zap.Any("op", op),
+	)
+}
+func (fs *readOnlyFsInternal) opEnd(op interface{}, err error) {
+	opType := reflect.TypeOf(op)
+	switch op.(type) {
```
ransomw1c: i suspect it's more idiomatic to assign a value at the top of the `switch` with, for example, `v := op.(type)` rather than adding duplicate type assertions in each `case`.

pkg/csi/controller.go
kerneltime: Yes, it is not a create for the backend but a create operation for the reference in kubernetes for a volume external to it.

kerneltime:
Related to #156

pkg/csi/controller.go
```diff
@@ -2,53 +2,148 @@ package csi
...
+	if err != nil {
+		s.config.driver.config.Logger.Error("requested repo could not be found", zap.Error(err))
+		return nil, fmt.Errorf("requested volume not found. err:" + err.Error())
+	}
+	attrs := make(map[string]string)
```
ransomw1c: it's possible to initialize maps with literals as in

```
attrs = map[string]string{"repo": repo, "hash": hash}
```

with some slight advantages for readability (all init-like stuff in one statement) and maintainability (not necessary to repeat the name of the map so many times)

pkg/csi/controller.go
```diff
@@ -2,53 +2,148 @@ package csi
...
+func getVolumeEntries(repos []model.RepoDescriptor, max int32) ([]*csi.ListVolumesResponse_Entry, string) {
+	volumes := make([]*csi.ListVolumesResponse_Entry, max)
+	for _, repo := range repos {
+		if max == 0 {
+			return volumes, repo.Name
+		}
+		volume := csi.ListVolumesResponse_Entry{
+			Volume: &csi.Volume{
+				CapacityBytes: math.MaxInt64,
+				Id:            repo.Name,
```
ransomw1c: remain somewhat confused as to the difference between this `csi.Volume` and the use in `CreateVolume`.  in one function, CSI volumes to correspond to bundles (uniquely identified by repo name + root hash), whereas here, volumes correspond to repos.

pkg/csi/node.go
```diff
@@ -2,41 +2,192 @@ package csi
...
+	}
+
+	n.l.Info("Publish volume done",
+		zap.String("volume", req.VolumeId),
+		zap.String("id", downloadedBundle.bundleID))
+	return &csi.NodePublishVolumeResponse{}, nil
+}
+
+func getDownloadedBundleKey(repo string, bundle string) string {
+	return repo + bundle
```
ransomw1c: here, as in `CreateVolume`'s use of `csi.Volume.Id`, the `csi.NodePublishVolumeRequest.VolumeId` corresponds to a bundle, except there's no `-` delimiter between the repo and bundle.. is it that the k8s and CSI apis remove the delimiter, or is there a mismatch between formatting the volume ids?

pkg/core/fs_ro_ops.go
kerneltime: Thanks

pkg/csi/controller.go
kerneltime: Yup much cleaner. Thanks

pkg/csi/controller.go
kerneltime: The datamon model is made to fit into this. In some ways, the volumes are repos and exact instance of the volume (snapshot) is the hash. This could have been modeled as snapshots but snapshot APIs are still in the works and needs each snapshot to be created. We might revisit this modeling when CSI is more baked and after getting feedback from the k8s community.

pkg/core/fs_ro_ops.go
kerneltime: Thank you!

deploy/controller.yaml
kerneltime: Will revisit the yaml generation when integrating with Argo.. for now the controller yaml will not be undergoing changes as it will be deployed only once. The current scheme seems simple enough.

pkg/csi/controller.go
ransomw1c: > revisit this modeling when CSI is more baked and after getting feedback from the k8s community

it's a plan.  #160 is open to track `ListVolumesResponse` as it pertains to datamon.

Issue #135, opened 2019-04-30
FUSE on k8s demo and documentation


ransomw1c:
prepare the FUSE implementation as something to be delivered to k8s users, specifically, easing the cost of getting started with datamon use via, for example

* an `examples/sidecar.yaml` demonstrating how to use a FUSE filesystem as a shared volume
* documentation in the form of a task-specific README

PR #136, opened 2019-05-02
init ro sidecar mount

ransomw1c:
toward #135


ransomw1c:
opening this to pivot back to making the mutable filesystem available from the CLI.

as is, it demonstrates mounting the read-only filesystem from a k8s sidecar container.  the `make fuse-demo-ro` target builds docker images for both a datamon container as well as an additional container, deploys the images to kubernetes, and starts a shell in the container that doesn't contain datamon.  from that shell, `ls /tmp/mount` displays the contents of a bundle via a FUSE mount in the sidecar container.

hack/k8s/example-ro.template.yaml
```diff
@@ -0,0 +1,50 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: datamon-ro-demo
+    spec:
+      containers:
+      - name: demo-shell
+        image: fuse-demo-shell:latest
+        imagePullPolicy: "Never"
```
kerneltime: Why never?

hack/k8s/example-ro.template.yaml
```diff
@@ -0,0 +1,50 @@
...
+        tty: true
+        volumeMounts:
+        - mountPath: /tmp/mount
+          name: fuse-mountpoint
+          mountPropagation: "HostToContainer"
+      - name: datamon-sidecar
+        image: fuse-demo-sidecar:latest
+        imagePullPolicy: "Never"
+        command: ["datamon"]
+        args: ["bundle", "mount", "--bundle", "1Kbtk5Mo0M8ck2lcnuhMNYiV9DO", "--repo", "ransom-datamon-test-repo", "--destination", "/tmp", "--mount", "/tmp/mount"]
```
kerneltime: skip the hash and let it download the latest for demo

hack/k8s/example-ro.template.yaml
```diff
@@ -0,0 +1,50 @@
...
+        securityContext:
+          privileged: true
```
kerneltime: What is the minimum securityContext? Does SYS_ADMIN work?

hack/k8s/example-ro.template.yaml
```diff
@@ -0,0 +1,50 @@
...
+        stdin: true
+        tty: true
+        volumeMounts:
+        - mountPath: /tmp/mount
+          name: fuse-mountpoint
+          mountPropagation: "Bidirectional"
```
kerneltime: Do we need bidirectional for within pod sharing?

ransomw1c: yes, i have tried toggling this to [other strings](https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation) and re-running the `fuse-demo/` scripts to drop into the `demo-shell`, and `ls /tmp/mount` from the demo shell only displays the demo files with bidirectional mount propagation.

in particular, what was not obvious to me (at the level of k8s abstraction) is why there's host->container unidirectional propagation and not container->host.

hack/k8s/example-ro.template.yaml
ransomw1c: i've tried both

```
        securityContext:
          allowPrivilegeEscalation: true
          capabilities:
            add:
              - SYS_ADMIN
```
and
```
        securityContext:
          capabilities:
            add:
              - SYS_ADMIN
```

and neither allows access to fuse

kerneltime: ok

hack/k8s/example-ro.template.yaml
kerneltime: I guess the mounts need to go back to the os for it to share with the other container cause pod is a k8s abstraction and as far as the OS is concerned these are separate containers.

hack/k8s/example-ro.template.yaml
```diff
@@ -0,0 +1,50 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: datamon-ro-demo
+    spec:
+      containers:
+      - name: demo-shell
+        image: fuse-demo-shell:latest
+        imagePullPolicy: "IfNotPresent"
```
kerneltime: I think this is best set to always? From the documentation ```imagePullPolicy is omitted and either the image tag is :latest or it is omitted: Always is applied.``` so the policy here will be ignored cause we want to pick the latest?

Issue #138, opened 2019-05-03
daemonize mount command


ransomw1c:
by default, the `bundle mount` command ought to run a [daemonized](https://github.com/jacobsa/daemonize/blob/master/daemonize.go) process similar to [gcsfuse](https://github.com/GoogleCloudPlatform/gcsfuse).

unmounting the bundle ought to cause the process to exit.

further impedance matching with gcsfuse [suggests](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/mounting.md#mounting) that adding a `--foreground` flag to disable daemonization is desirable.

ransomw1c:
the main differences between daemonizing and running as a background process are:

* the process runs in a different process group and thus doesn't get signals sent to the terminal
* the process's stdio are disconnected from the terminal

so in particular, when running as a daemonized process, logs need be written to a file

kerneltime:
@kerneltime to finish up the code review.

ransomw1c:
[need to toggle switch semantics](https://github.com/oneconcern/datamon/pull/140#discussion_r286661734).

fredbi:
mount as a deamon is now supported and allegedly, works fine. Closing this now.

PR #140, opened 2019-05-07
mount daemonization

ransomw1c:
address #138


ransomw1c:
so far, daemonization applies to the read-only mount only

ransomw1c:
daemonization applies to both mutable and read-only mounts as of d48706b

cmd/datamon/cmd/bundle_mount.go
```diff
@@ -13,22 +19,94 @@ import (
 	"github.com/spf13/cobra"
 )
 
+func registerSIGINTHandlerMount(mountPoint string) {
```
kerneltime: Should this code go some place else? Either a new file or a new package?

cmd/datamon/cmd/bundle.go
```diff
@@ -66,6 +67,11 @@ func addBundleFileFlag(cmd *cobra.Command) string {
 	return file
 }
 
+func addDaemonizeFlag(cmd *cobra.Command) string {
+	cmd.Flags().BoolVar(&bundleOptions.Daemonize, daemonize, false, "Whether to run the command as a daemonized process")
+	return file
```
ransomw1c: `?file? c/daemonize/`

cmd/datamon/cmd/bundle_mount.go
ransomw1c: this (like much of the daemonization impl) is lifted out of [gcsfuse](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/main.go), which suggests that `registerSIGINTHandlerMount` is be in the same package as much of the CLI.

i'm open to concrete suggestion about where this should go, of course, and will put it in a new file in case that's clearer.

tangentially, the distinction between physical layout and logical layout is an interesting dichotomy in go:  unlike C, the language has a notion of logical dependencies -- i.e. packages.  like C (albeit in a implementation-dependent way), physical layout in go has a natural correspondence with logical deps -- i.e. [intra-package initialization](https://golang.org/ref/spec#Package_initialization).

that said, the opinionated note i want to make (and the reason that i'm generally not too concerned about physical layout as opposed to logical deps), is that i'm not particularly concerned about physical layout in go, because the correspondence between it and logical deps is a fairly weak one compared to packages .. something that probably only makes much difference in terms of `init()` functions and the like.

documentation (and api design and a bunch of other ways to make the codebase more user-friendly) are, of course, super important.  let's try to maintain the "design stamina" on these and related issues.

cmd/datamon/cmd/bundle.go
```diff
@@ -66,6 +67,11 @@ func addBundleFileFlag(cmd *cobra.Command) string {
 	return file
 }
 
+func addDaemonizeFlag(cmd *cobra.Command) string {
+	cmd.Flags().BoolVar(&bundleOptions.Daemonize, daemonize, false, "Whether to run the command as a daemonized process")
```
kerneltime: We can make it true by default

kerneltime:
Looks good, you need to rebase and resolve conflicts.

cmd/datamon/cmd/bundle.go
ransomw1c: added this to #138

PR #145, opened 2019-05-08
cafs: verify hash functions on get

ransomw1c:
opening this instead of #102 to get the source branch from within the same repo


pkg/cafs/reader.go
```diff
@@ -156,27 +173,61 @@ func (r *chunkReader) Read(data []byte) (int, error) {
...
-			if err == io.EOF { // we reached the end of the stream for this key
+			if errRead == io.EOF { // we reached the end of the stream for this key
 				r.idx++
 				r.rdr = nil
 				r.lastChunk = r.idx == len(r.keys)
-
+				if r.verifyHash {
+					nodeOffset := r.idx
+					isLastNode := false
+					/* ??? what? */
```
kerneltime: not very informative about the concern here.. can you please elaborate

Issue #142, opened 2019-05-07
ship executable-running tool rather than executable itself


ransomw1c:
the vision of this issue, as described in [this article](https://eng.heetch.com/self-updating-tools-in-go-lang-9c07291d6285) is to

> push updates toward the team rather than burdening them with mandatory polling

specifically, what this means is distributing a tool that's intended not to change (or to change infrequently) after the initial install.  that tool is a program responsible for updating and running the datamon program itself, which is expected to change over time.

implementation decisions such as whether the tool downloads a pre-built binary or builds the program for source are part of this issue.

existing libraries designed to accomplish this sort of thing include [`go-update`](https://github.com/inconshreveable/go-update) and [`demand`](https://github.com/tv42/demand)

possibly related issues include #22 and #15

kerneltime:
Treating this as a spike for now. There are few options
1. go-update
2. demand
3. Run docker image in a shell script
4. Brew
5. manual curl command. (no self upgrade)

ransomw1c:
[`goreleaser`](https://goreleaser.com) [has been suggested](https://oneconcern.slack.com/archives/CCMQ25SP5/p1557774363000300) as a way to get binaries into github releases, brew, etc. in particular, goreleaser integrates with both circleci and github actions.

an additional implementation idea related to the requirements of this ticket is to put a git hash in the binary as one means of describing versions.

fredbi:
goreleaser now takes care of releases, including the brew tap repository.
The datamon binary is now self-updating (as an optional CLI flag)
There is still some work that could be carried out on docker releases, but I'd say they would better fit into a fresher issue.

Considering this closed now.

PR #159, opened 2019-05-13
init rel download shell script

ransomw1c:
this is the "manual curl command" option (via github releases) to push updates to users and address #142


hack/github_release_run.sh
```diff
@@ -0,0 +1,32 @@
+#! /bin/sh
+
+# todo: replace release id with /latest after non-pre-release exists
+# REL_URL=https://api.github.com/repos/oneconcern/datamon/releases/latest
+REL_URL=https://api.github.com/repos/oneconcern/datamon/releases/16541641
```
kerneltime: We will cut a release soon that works with latest, you can flip this to latest. Will merge it in and cut the release in this week.

kerneltime:
The script can be in deploy and called datamon?

kerneltime:
We can also include the MD5 sum as part of releases and include them in the script.

ransomw1c:
yes, some checksum ought to be added with regard to security.

deploy/datamon.sh
```diff
@@ -0,0 +1,97 @@
+#! /bin/sh
+
+# set these to non-zero to enable
```
kerneltime: Please add some documentation for that this script is expected to do.

Issue #158, opened 2019-05-11
Caching CAFS for RO FS Fuse mount


kerneltime:
When a RO fuse FS is mounted, if the flag is set, the FS is made available as soon as the metadata is ready. A CAFS local cache storage is used to pass in pre downloaded blobs. Any misses in the CAFS cache result in it being fetched from the the backend before serving the read. 

An additional optimization is to only download the chunk which serves the read boundary.

ransomw1c:
from #162 , it emerges that having a `storage.Store` that uses local storage and does atomic writes via `mv` will be useful toward this iss.

fredbi:
Update: blob cache and prefetching work fine. Closing this now.

PR #162, opened 2019-05-14
cafs-cached ro FUSE fs

ransomw1c:
address #158


ransomw1c:
this is still very early progress.  opening the merge request now for communication purposes.

ransomw1c:
i still consider this very much work-in-progress with some amount of cleanup at all layers (examples: tee writes to caching cafs, design updates to creating rofs instances, testing of CLI .. and everywhere).

yet conceptually, this is ready for a test drive with
```
umount /tmp/mmp ; rm -r /tmp/mmfs /tmp/mmp ; mkdir /tmp/mmfs ; mkdir /tmp/mmp ; ./cmd/datamon/run_datamon bundle mount --repo ransom-datamon-test-repo --destination /tmp/mmfs --mount /tmp/mmp --cache
```
then observe that as files are read at `/tmp/mmp` the cache is populated at `/tmp/mmfs`.

ransomw1c:
similarly, the incantation for tests is

```
go build && umount /tmp/mmp; rm -r /tmp/mmp /tmp/mmfs ; go test -v -tags fuse_cli -run 'TestBundleMountCached$'
```

pkg/core/fs.go
```diff
@@ -50,7 +52,12 @@ type MutableFS struct {
 }
 
 // NewReadOnlyFS creates a new instance of the datamon filesystem.
+// todo: control flow regarding whether to cache pivots on this function.  ideas:
```
ransomw1c: this api design quandry is the only additional `todo` incurred with this diff .. would appreciate opinions regarding this code design decision in particular

ransomw1c:
particular questions upon letting this marinate and then revisiting:

* use `mv` for atomicity of (linux) fs operations?
*

pkg/cafs/cafs.go
```diff
@@ -252,3 +259,124 @@ func IsRootKey(fs storage.Store, key Key, leafSize uint32) bool {
...
+	}
+}
+
+type cachedFs struct {
+	fsImpl           Fs
+	cacheStore       storage.Store
+	cafsKey2StoreKey func(Key) (string, error)
+}
+
+func (fs *cachedFs) Put(ctx context.Context, src io.Reader) (int64, Key, []byte, bool, error) {
```
kerneltime: Do we need to implement Put? Also, we should not be using ioutil.ReadAll interface.

pkg/cafs/cafs.go
```diff
@@ -252,3 +259,124 @@ func IsRootKey(fs storage.Store, key Key, leafSize uint32) bool {
...
+		return 0, Key{}, nil, false, err
+	}
+	err = fs.cacheStore.Put(ctx, storeKey, &bytesBuf, storage.IfNotPresent)
+	if err != nil {
+		return 0, Key{}, nil, false, err
+	}
+	return written, key, keys, found, err
+}
+
+// ??? thread-safety
```
kerneltime: For thread safety, the cacheStore can write to a file and move into place. These are CAFS files so safe in terms of over writes.

pkg/cafs/cafs.go
ransomw1c: as of 17e09219ea37 , this falls back to the decorated `fsImpl`.

pkg/core/fs.go
kerneltime: The property for what kind of bundle it is makes sense. Is it pre fetched, or streamed.

pkg/cafs/cafs.go
ransomw1c: i've added this to #158 and will open a separate request to this effect

pkg/core/fs.go
ransomw1c: deferring this for now..

pkg/cafs/cafs.go
ransomw1c: i made a `storage.Store` implementation that relies on move of the underlying `afero.Fs` to provide atomicity and used it in `bundle_mount.go` as suggested here.

considering move as the atomic part of filesystem operations considered insightful.

pkg/core/fs.go
ransomw1c: i've gone with the option of consolidating the `NewReadOnlyFS` functions according to `BundleOption`s

Issue #153, opened 2019-05-10
Harmonize the usage of paths/keys/delimiters across model, core and store


kerneltime:
The logic for creating keys and choice of delimiters should belong entirely in the model package.
The core package should extract that information from model and use it to talk to store.
The store treats them as strings without any assumptions about the key generation logic.
This is also related to the list apis not returning pre formatted lists but the structs that the CLI can use to print to console.

PR #166, opened 2019-05-20
metadata store keys api boundary to model pkg

ransomw1c:
toward #153


ransomw1c:
unsure if all access to the metadata store's key naming convention is now filtered thru `model` or not ... this is the only instance i was aware of

Issue #152, opened 2019-05-10
Consistency of concurrency patterns in code base


kerneltime:
There are a few patterns used in the code base for concurrency for parallelism. The recent code changes reflect the approach we want to take for all code where concurrency is in play. 
This is important to avoid bugs we have seen when concurrency is in play.

PR #167, opened 2019-05-20
CLI bundle upload concurrency patterns

ransomw1c:
towards #152


ransomw1c:
this is mostly a pattern-matching effort between the `fs_rw_ops.go` and `bundle_pack.go`.

in retrospect, the opinionated change here that makes a difference to functionality is using unbuffered channels where possible.. the done channel notion is what makes using unbuffered channels without leaking goroutines possible.

PR #169, opened 2019-05-21
Limit the concurrency during uploads.

kerneltime:
For large file counts or large files, the concurrency needs to be
reduced to keep memory usage in check which is critical for k8s
deployment.

### DIFF
--- a/pkg/cafs/cafs.go
+++ b/pkg/cafs/cafs.go
@@ -148,8 +148,8 @@ func (d *defaultFs) writer(prefix string) Writer {
 		pather:        nil,
 		prefix:        prefix,
 		count:         0,
-		flushChan:     make(chan blobFlush, 100000),
-		errC:          make(chan error, 1000000),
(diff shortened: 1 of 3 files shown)


pkg/cafs/cafs.go
```diff
@@ -148,8 +148,8 @@ func (d *defaultFs) writer(prefix string) Writer {
 		pather:        nil,
 		prefix:        prefix,
 		count:         0,
-		flushChan:     make(chan blobFlush, 100000),
-		errC:          make(chan error, 1000000),
+		flushChan:     make(chan blobFlush, sizeOfFlushChan),
+		errC:          make(chan error, 1),
```
ransomw1c: in case of more than two errors, this'll leave goroutines parked on write to the error channel.

ransomw1c:
this contains changes (and the comment re. goroutine leaks is) regarding #171

pkg/cafs/cafs.go
kerneltime: Yes we need to fix #171

PR #172, opened 2019-05-23
Allow for root to be readable by all.

kerneltime:
Mount the FS as non root in the sidecar demo
Set root permission to 0777

### DIFF
--- a/pkg/core/fs.go
+++ b/pkg/core/fs.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/core/fs.go
```diff
@@ -28,7 +29,7 @@ const (
 	firstINode       fuseops.InodeID = 1023
 	dirDefaultMode                   = 0777 | os.ModeDir
 	fileDefaultMode                  = 0666
-	dirReadOnlyMode                  = 0755 | os.ModeDir
+	dirReadOnlyMode                  = 0777 | os.ModeDir
```
ransomw1c: c2ce08 is there to hopefully save some typing while preserving meaningful naming conventions on these various `*Mode` constants

ransomw1c:
#173 is open to build sidecar images for distribution.

pkg/core/fs.go
kerneltime: For now leaving this as two separate modes.

PR #177, opened 2019-05-24
Stream in CAFS data for read only FS mounted bundles

kerneltime:

### DIFF
--- a/pkg/cafs/reader.go
+++ b/pkg/cafs/reader.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"strings"
 	"sync"
 
 	"github.com/minio/blake2b-simd"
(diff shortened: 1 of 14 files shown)


pkg/storage/localfs/store.go
```diff
@@ -76,6 +76,7 @@ func (rc readCloser) Close() error {
 }
 
 func (l *localFS) Put(ctx context.Context, key string, source io.Reader, exclusive bool) error {
+	// TODO: Change this implementation to use rename to put file into place.
```
ransomw1c: see #162 for a decorator pattern approach

ransomw1c:
overall it looks quite good.  adding a `ReadAt` to `Bundle` was not an approach i considered.

pkg/cafs/reader.go
```diff
@@ -159,7 +161,46 @@ func (r *chunkReader) WriteTo(writer io.Writer) (n int64, err error) {
...
+		// Read first leaf
+
+		read, err = rdr.ReadAt(p[n:], offset)
+		n += read
+		index++
+		offset = 0
+		if err != nil && !strings.Contains(err.Error(), "EOF") {
+			return
+		}
+		if (len(p) == n) || (index >= int64(len(r.keys))) {
```
ransomw1c: what is the intent of `len(p) == n`?  when is this false?

ransomw1c:
global style comment: the "naked return" wherein return variables are specified in the function declaration is suggested to be used in terse functions only according to the Go Tour.

pkg/storage/localfs/store.go
kerneltime: Can you file an issue under harmonization? Don't want to miss this.

pkg/cafs/reader.go
kerneltime: len(p) can be greater than n till reads are complete.

kerneltime:
Moving this to https://github.com/oneconcern/datamon/pull/182

Issue #87, opened 2019-03-14
Labels


kerneltime:
Add ability to attach labels to bundles. 
1. Contributor defined labels
2. Labels generate for every commit for ease of use. Example <dev name>-<an easy to type word>
The namespace for labels is scoped per repo basis.

kerneltime:
@kerneltime to write a proposal for the functionality of label and how it will work for review.

ransomw1c:
### notes from conversation

✨ what a label is: a label is a key-value map with
* keys ~ human-readable text
* vals ~ machine generated hashes

##### cli requirements

* add a `--label` flag to bundle upload operation 🌀 possibly with multiple, comma-separated values
* commands to set labels and list bundles
* download and mount by label

##### implementation details

* labels will exist in their own `storage.Store` (i.e. bucket in the case of GCS), and each label gets its own serialized entry (e.g. a `.json` file or otherwise -- not a directory as in the case of the extant GCS buckets)
* add additional function to `cmd/` `Run:` implementations -- or possibly `uploadBundleDescriptor` -- that does label uploading

ransomw1c:
there's currently a bug on overwriting labels (using `bundle upload` or `label set` with an extant label) wherein

```
googleapi: Error 412: Precondition Failed, conditionNotMet
```

is the error reported out of the application.

fredbi:
Labels now work fine.

PR #179, opened 2019-05-28
initialize labels implementation

ransomw1c:
towards #87

### DIFF
--- a/pkg/model/label.go
+++ b/pkg/model/label.go
@@ -1,0 +1,25 @@
+package model
+
+import (
+	"fmt"
+	"time"
+)
+
+type LabelDescriptor struct {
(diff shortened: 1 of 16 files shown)


ransomw1c:
the current implementation allows one label per `--label` flag, not multiple, comma-separated values.

pkg/model/label.go
```diff
@@ -0,0 +1,22 @@
+package model
+
+import (
+	"fmt"
+)
+
+type LabelDescriptor struct {
+	BundleID string `json:"id" yaml:"id"`
```
kerneltime: Add timestamp

kerneltime:
LGTM
Will go over one more time after timestamp are added. Adding Contributor to label might also be useful.

PR #182, opened 2019-05-29
CAFS streaming

kerneltime:
This supersedes https://github.com/oneconcern/datamon/pull/177
Includes 
1. Caching of working set of Blobs
2. Fixes for passing CI.

### DIFF
--- a/pkg/cafs/reader.go
+++ b/pkg/cafs/reader.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"strings"
 	"sync"
 
 	"github.com/minio/blake2b-simd"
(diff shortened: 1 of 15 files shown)


pkg/cafs/reader.go
```diff
@@ -159,6 +170,85 @@ func (r *chunkReader) WriteTo(writer io.Writer) (n int64, err error) {
 
 }
 
+func (r *chunkReader) calculateKeyAndOffset(off int64) (index int64, offset int64) {
+	index = off / int64(r.leafSize)
+	offset = off - (index * int64(r.leafSize))
```
ransomw1c: `off % int64(r.leafSize)` instead?

pkg/cafs/reader.go
```diff
@@ -159,6 +170,85 @@ func (r *chunkReader) WriteTo(writer io.Writer) (n int64, err error) {
...
+				r.lru.Add(k.StringWithPrefix(r.prefix), b)
+			}(nextIndex)
+		}
+	}
+
+	for {
+		// Fetch Blob
+		var buffer []byte
+		key := r.keys[index]
+		if r.lru.Contains(key.StringWithPrefix(r.prefix)) {
```
ransomw1c: on the first pass of the outer `for` loop, it's possible that this `if` block will need to account for `offset`

pkg/cafs/reader.go
```diff
@@ -159,6 +170,85 @@ func (r *chunkReader) WriteTo(writer io.Writer) (n int64, err error) {
...
+
+	// Calculate first key and offset.
+	index, offset := r.calculateKeyAndOffset(off)
+	if index >= int64(len(r.keys)) {
+		return 0, nil
+	}
+
+	var read int
+
+	seekHead := func() {
```
ransomw1c: might this broken out as a separate function and/or could there be a comment regarding its intent?

in particular, i'm curious to hear more about the (presumed) analogy between what this function is doing and [`seek`](http://man.cat-v.org/plan_9/2/seek) calls on file-handles.

pkg/cafs/reader.go
kerneltime: that works too

pkg/cafs/reader.go
kerneltime: seek ahead implies that the reader is seeking ahead to the next blob in the sequence.
The goal here is to serve reads from memory as far as possible when streaming a file. This way a read request never blocks on read completing from backend. Will add the comments.

ransomw1c:
the overall idea i'm taking from this change is that it's an implementation of streaming from `gcs` thru `cafs` to the bundle...

... the in-memory cache is added to the stream at the `cafs` level in order to prevent "stutters" on streaming.

kerneltime:
> the overall idea i'm taking from this change is that it's an implementation of streaming from `gcs` thru `cafs` to the bundle...
> 
> ... the in-memory cache is added to the stream at the `cafs` level in order to prevent "stutters" on streaming.

Yes

Issue #163, opened 2019-05-16
CLI version support for Mac and CLI shipment for linux


kerneltime:
Need to implement a makefile target that will

- [x] Build binary for mac with version support

- [x] Build binary for linux in addition to container image

- [ ] as part of circle ci push of a release build and publish the binaries

- [ ] Support for brew tap

- [ ] Documentation for brew tap

ransomw1c:
`make build-datamon-binaries` outputs the files to upload in the github release web ui to `out/release`.

note that the mac build uses the internal go linker (the linker implemented in go), not the llvm stack present on the mac desktop, because the build occurs in docker.  the `build-datamon-mac` target exists in case it becomes desirable to use an external linker to build the mac binary.

fredbi:
Released artifacts are:
- brew tap for mac & linux
- docker images (linux built, may be used on any host)
- binaries for linux & mac

considering this done now. Closing.

PR #183, opened 2019-05-30
binary builds

ransomw1c:
toward #163


ransomw1c:
as is this is the beginnings of makefile targets to build binaries in a non-CI way:

* `build-datamon-mac` builds a binary for mac os x using an external linker
* `build-datamon-binaries` creates the contents of a github release suitable for upload in `out/release` (currently sans source archives)

kerneltime:
Please rebase

PR #184, opened 2019-05-30
Create directories pass into CLI.

kerneltime:


ransomw1c:
does what it advertises: creates directories if they're absent..

.. noting in particular that the `cmd/` package is responsible for creating the data directory, while the `core/` package is responsible for creating the mount point.

PR #187, opened 2019-05-31
Ritesh/filelist

kerneltime:


cmd/datamon/cmd/bundle_upload.go
```diff
@@ -63,9 +65,31 @@ var uploadBundleCmd = &cobra.Command{
...
+				}
+				lineScanner := bufio.NewScanner(file)
+				files := make([]string, 0)
+				for lineScanner.Scan() {
+					files = append(files, lineScanner.Text())
+				}
+				return files, nil
+			}
+		} else {
+			fn = func() ([]string, error) {
```
ransomw1c: note that function types can be nil, so it's possible to allow `core/` to specify the default behavior on `nil`

PR #188, opened 2019-05-31
Allow list of files being passed and skip file

kerneltime:
This change allows a list of files to be passed in and allows for
upload to continue on error.


pkg/core/bundle_test.go
```diff
@@ -93,9 +93,11 @@ func TestBundle(t *testing.T) {
 		ConsumableStore(consumableStore),
 		BlobStore(reArchiveBlob),
 	)
-
+	getKeys := func() ([]string, error) {
+		return bundle.ConsumableStore.Keys(context.Background())
+	}
 	require.NoError(t,
-		implUpload(context.Background(), archiveBundle2, reBundleEntriesPerFile))
+		implUpload(context.Background(), archiveBundle2, reBundleEntriesPerFile, getKeys))
```
ransomw1c: can pass `nil` here as well

kerneltime: yup

PR #193, opened 2019-06-03
Go modules update and minor fixes

kerneltime:


ransomw1c:
these changes fix the label over-write bug mentioned in #87 

there's also a change wherein errors a relayed to the calling function without risking `panic()` out of a type assert.

PR #198, opened 2019-06-06
'list' commands cleanup

ransomw1c:
mentioned in #153 , this is (at least the start of) a quick cleanup of some todos:

several commands are used to display metadata.  this is a cleanup to make the implementation incrementally more maintainable:  the `core/` api returns types rather than strings (this was part of the original todo), and the `datamon/cmd/` implementation uses `text/template` to format the strings (this is my bright idea 😒 ).

### DIFF
--- a/pkg/core/label.go
+++ b/pkg/core/label.go
@@ -14,7 +14,6 @@ import (
 )
 
 type Label struct {
-	Name       string
 	Descriptor model.LabelDescriptor
 }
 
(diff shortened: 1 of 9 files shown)


ransomw1c:
the diff also moves `Label.Name` to `Label.Descriptor.Name` for consistency with repo and bundle metadata

pkg/core/label.go
```diff
@@ -38,8 +37,8 @@ func NewLabelDescriptor(descriptorOps ...LabelDescriptorOption) *model.LabelDesc
 }
 
 func LabelName(name string) LabelOption {
-	return func(b *Label) {
-		b.Name = name
+	return func(l *Label) {
```
kerneltime: For existing labels should we get it via the file name?

kerneltime:
LGTM. add the support for existing labels to return a name for backward compatibility.

PR #203, opened 2019-06-12
optional --destination param on mount

ransomw1c:
following up on a request from @galvare2 , these changes allow omitting the `--destination` on `bundle mount` (and `bundle mount new`).

when omitted, a subpath of system's `/tmp` directory is constructed by golang's `io/ioutil`.

### DIFF
--- a/cmd/datamon/cmd/bundle_mount.go
+++ b/cmd/datamon/cmd/bundle_mount.go
@@ -5,6 +5,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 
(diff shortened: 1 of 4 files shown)


cmd/datamon/cmd/bundle_mount.go
```diff
@@ -77,12 +78,23 @@ var mountBundleCmd = &cobra.Command{
...
 
-		path, err := sanitizePath(bundleOptions.DataPath)
-		if err != nil {
-			log.Fatalf("Failed to sanitize destination: %s\n", bundleOptions.DataPath)
-			return
+		var err error
+
+		var consumableStorePath string
+		if bundleOptions.DataPath == "" {
+			consumableStorePath, err = ioutil.TempDir("", "datamon-mount-destination")
```
kerneltime: We can use the bundle hash and write the bundle descriptor last or write a done marker file. The next time we can skip downloading the files and reused the json files downloaded.. ?

kerneltime:
Needs conflict resolution..

cmd/datamon/cmd/bundle_mount.go
ransomw1c: the bundle descriptor and filelists are written to disk at the same time it's used to initialize the in-memory structures.  #218 contains [the change](https://github.com/oneconcern/datamon/pull/218/files#diff-7d60c86cd0267ea4660946771a11f969R62) to read `${destination}/.datamon/` folder from disk into in-memory structures.

since this implementation places the destination in a temporary directory that might vanish after the process exits, it's not clear where the reusable json files are to be stored... i suppose creating a subdirectory of `${HOME}/.datamon`, say `${HOME}/.datamon/meta-cache`, is a reasonable option?

creating a new issue to document the intent of these changes seems more likely to me than including them in this diff.

Issue #199, opened 2019-06-07
datamon visualization


ransomw1c:
following up on [a request from Zhuo](https://oneconcern.slack.com/archives/CBD5FLXDM/p1559256679005200) on the data science team, this issue is about making datamon more user-friendly via a web interface.

to start with, the task here is to list metadata information (repos, bundles, files, labels, etc.) in a web ui .. specifically, the expectation is a simple web 1.0 type experience (server-side formatting, perhaps some jquery- or mootools-style twine and ducttape).

in keeping with the "git for datascience" vision statement, we'll want to construct a more replete ui over time -- something like github -- despite the modest goals for this particular task.

in addition to implementation, this task might also include deployment of the web ui to an internal system at OneConcern.

fredbi:
web ui is working, and provides basic repo browsing capabilities. Considering this done.

PR #206, opened 2019-06-13
init webserver

ransomw1c:
begin to address #199


ransomw1c:
leaving the wip tag on this as there could still be some additional functionality by way of ui.

as is, #199 minimum requirements are done.

Issue #171, opened 2019-05-22
Change CAFS writer logic to be less memory hungry


kerneltime:
Currently the CAFS writer has to keep a record in the channel for each chunk that is uploaded until the flush is called. This leads to channel sizes that are large enough to handle the larges possible file. This needs to change to where the channels are being drained in parallel.

kerneltime:
This will also eliminate the max CAFS file size limitation

ransomw1c:
i'm grouping the initial metric-collection setup with this task, since the plan is to collect metrics specific to upload of large files.

my plan here to to create a separate executable that does the metric collection.

ransomw1c:
slight scope shift here as the plan is to use a `--concurrency-factor` flag for both `bundle upload` and `download` that can be set between `1` and the upper bound on `int`s with sensible defaults.

while the `--concurrency-factor` might initially map directly onto some particular value (such as number of concurrent uploads or downloads), the external view need only have some parameter to tune concurrency (and hence memory usage), nothing as specific as number of files uploaded or downloaded.

PR #212, opened 2019-06-17
configurable concurrency/memory bounds

ransomw1c:
towards #171 , this request is about making the number of goroutines (hence the amount of memory used by the program) during upload configurable at runtime rather than statically parameterized.

### DIFF
--- a/pkg/cafs/cafs.go
+++ b/pkg/cafs/cafs.go
@@ -59,6 +59,12 @@ func Backend(store storage.Store) Option {
 	}
 }
 
+func ConcurrentFlushes(concurrentFlushes int) Option {
+	return func(w *defaultFs) {
+		w.concurrentFlushes = concurrentFlushes
+	}
+}
+
 type hasOpts struct {
 	OnlyRoots, GatherIncomplete bool
 	_                           struct{} // disallow unkeyed usage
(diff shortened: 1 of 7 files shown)


ransomw1c:
there'll still be some constants expressed as ratios (for example, the ratio of cafs flushes during put to the number of files uploaded at a time)

pkg/cafs/cafs.go
```diff
@@ -154,6 +162,10 @@ func (d *defaultFs) GetAt(ctx context.Context, hash Key) (io.ReaderAt, error) {
 }
 
 func (d *defaultFs) writer(prefix string) Writer {
+	maxGoRoutines := d.concurrentFlushes
+	if maxGoRoutines < 1 {
+		maxGoRoutines = 1
```
kerneltime: 👍

ransomw1c: ^ defensive..

kerneltime:
LGTM. A few changes

- [ ] Extend to read/download path

- [ ] Rename to concurrency/concurrency-factor? open to suggestions.

pkg/cafs/cafs.go
kerneltime: yup

PR #215, opened 2019-06-20
unbuffered channels on bundle upload

ransomw1c:
this patch is comparable to diff #167 towards iss #152 regarding concurrency patterns, specifically the preference for using unbuffered channels.

unlike #167 , this implementation additionally allows for limits on concurrency (number of files being uploaded at a time).


ransomw1c:
this is still a sketch:  more needs to be done by way of error-handling in particular.

also, the implementation needs to be cleaned up with regard to memory usage, keeping #171 in mind.

PR #216, opened 2019-06-20
generic concurrency-factor flag on upload, download

ransomw1c:
this is to address the [latest requirements](https://github.com/oneconcern/datamon/issues/171#issuecomment-503753719) on #171 , namely

- [x] Extend to read/download path

- [x] Rename to concurrency/concurrency-factor? open to suggestions.

### DIFF
--- a/pkg/cafs/reader.go
+++ b/pkg/cafs/reader.go
@@ -14,6 +14,11 @@ import (
 	"github.com/oneconcern/datamon/pkg/storage"
 )
 
+type Reader interface {
+	io.ReadCloser
+	io.ReaderAt
+}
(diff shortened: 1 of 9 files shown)


pkg/cafs/reader.go
```diff
@@ -14,6 +14,11 @@ import (
 	"github.com/oneconcern/datamon/pkg/storage"
 )
 
+type Reader interface {
```
ransomw1c: this type is added (1) for consistency with the `Writer` type and (2) to avoid the type assertion that was formerly present in `defaultFs.GetAt()`

Issue #214, opened 2019-06-19
bundle diff and update


ransomw1c:
in order to move data from Argo data-science workflows to production services (that is, the programs more directly connected to an externally visible web program) via datamon, the plan is to use either bundle ids or labels (which resolve to bundle ids) to provide of referring to workflow output from the production services.  now, from one such reference (bundle id) to the next, there could in general be a fair amount of duplicate files in the data-science artifact that production needs to access.  production is using `bundle download` (not the FUSE fs) to access the data, so supposing that there's already a bundle downloaded to the production environment, there need be a way, given the next reference, to (1) describe the diff between the currently downloaded bundle (in terms of file lists, not file contents) as well as (2) updated the downloaded bundle on disk such that only the differences as resolved, not such that the entirety of the bundle is downloaded again.

as is, the result of a `bundle download` contains a `.datamon/` directory with the bundle's metadata (file lists with names and hashes).  so to provide the diff, it'll suffice to compare the metadata available on GCS to the local `.datamon/` directory.  then updating the bundle on disk will consist of concurrently iterating through the diff, (i) adding missing files, (ii) removing additional files, and (iii) replacing any file with a differing hash.  afterward, the `.datamon/` directory metadata will need to be updated such that the result of an update is the same as the result of fresh `bundle download`:  there is not a local history being stored within `.datamon/` as with `.git/`.

----

note that this iss is distinct from the similarly-titled #204 :  that issue has to do with updating a bundle stored in GCS via local changes (specifically, via the FUSE filesystem abstraction).  this issue is about updating a bundle stored locally (after a `bundle download`) with changes from GCS.

ransomw1c:
@jakedsouza @galvare2 here's a design sketch of the CLI to review

### `bundle diff`

the `bundle diff` command diffs a bundle on the local filesystem with a bundle stored in GCS.  it uses the same flags as `bundle download`, so in particular bundles in GCS can be specified with either a label or a bundle id.

output is a CSV list

```
<diff_type> , <filename> , <remote_size> , <remote_hash> , <local_size> , <local_hash>
```

where `<diff_type>` may be among the following

* `A` (added) file name has been added to remote, not present in local
* `D` (deleted) file name is not present on remote, is present on local
* `U` (updated) file name is present on both the local and remote with different hash values on each

the `<filename>` is always present while `<remote_*>` and `<local_*>` are omitted in the case of deleted and added diff types, respectively.

```
[...]

PR #218, opened 2019-06-21
[wip] init bundle diff

ransomw1c:
the `bundle diff` part of #214


pkg/core/bundle_unpack.go
```diff
@@ -13,14 +16,58 @@ import (
...
-	bundleDescriptorBuffer, err := storage.ReadTee(ctx,
-		bundle.MetaStore, model.GetArchivePathToBundle(bundle.RepoID, bundle.BundleID),
-		bundle.ConsumableStore, model.GetConsumablePathToBundle(bundle.BundleID))
+func setBundleIDFromConsumableStore(ctx context.Context, bundle *Bundle) error {
+	// todo: use KeysPrefix() after that's implemented.
+	keys, err := bundle.ConsumableStore.Keys(ctx)
 	if err != nil {
 		return err
 	}
+	// todo: this is re-encoding model.GetConsumablePathToBundle()
```
ransomw1c: getting the bundle id out of the `.datamon/` folder on disk, and in particular the duplication (in the sense that there are two implementations that can't be changed independently) of this function with `model` is one of the remaining pain-points in this implementation.

cmd/datamon/cmd/bundle_diff.go
```diff
@@ -0,0 +1,110 @@
...
+			core.ConsumableStore(destinationStore),
+		)
+		remoteBundle := core.New(core.NewBDescriptor(),
+			core.Repo(params.repo.RepoName),
+			core.MetaStore(sourceStore),
+			core.BlobStore(blobStore),
+			core.BundleID(params.bundle.ID),
+		)
+
+		if err = core.PublishMetadata(context.Background(), localBundle); err != nil {
```
ransomw1c: these `PublishMetadata` calls are probably better included in `core.Diff` for consistency

ransomw1c:
`core.PopulateFiles` used by `bundle list file` now partially duplicates `PublishMetadata`.

cmd/datamon/cmd/cli_test.go
```diff
@@ -688,6 +685,143 @@ func TestDownloadBundles(t *testing.T) {
...
+	//
+	runCmd(t, []string{"bundle",
+		"upload",
+		"--path", dirPathStr(t, files[0]),
+		"--message", "diff update",
+		"--label", label,
+		"--repo", repo1,
+		"--concurrency-factor", "20",
+	}, "upload bundle at "+dirPathStr(t, files[0]), false)
+	// todo: zero out params in runCmd(), not cleanup()
```
ransomw1c: in particular..

ransomw1c:
since #220 is a dep, i'll resolve conflicts after it lands

pkg/web/routes.go
```diff
@@ -172,13 +172,12 @@ func (s *Server) HandleBundleListFiles() http.HandlerFunc {
...
 		repoName := chi.URLParam(r, "repoName")
 		bundleID := chi.URLParam(r, "bundleID")
-		bundle := core.Bundle{
-			RepoID:           repoName,
-			BundleID:         bundleID,
-			MetaStore:        s.metadataStore(),
-			BundleDescriptor: model.BundleDescriptor{},
-		}
-		err := core.PopulateFiles(context.Background(), &bundle)
+		bundle := core.New(core.NewBDescriptor(),
```
galvare2: What is the rationale behind this type of syntactic change?

ransomw1c: since the `cmd/datamon/cmd` package uses `core.New` (admittedly sugar), the main idea here is consistency.

now, the `core.New` itself is something that we could view as idiomatic, a sort-of borrowed idiom from Java, etc. where constructor functions are used.  golang idioms are liable to change over time, yet in no small part because i buy into the utility of type inference, i suppose the constructor idiom is a helpful one:  we get to treat the return values of constructors as semi-loosely (i.e. duck) typed.

pkg/model/bundle.go
```diff
@@ -64,12 +65,76 @@ func (c *Contributor) String() string {
...
+	ConsumableStorePathTypeFileList
+)
+
+type ConsumableStorePathMetadata struct {
+	Type     byte
+	BundleID string
+	Index    uint64
+}
+
+func GetConsumableStorePathMetadata(path string) (ConsumableStorePathMetadata, error) {
```
kerneltime: Please add a line as to what this method is expected to do.

pkg/core/bundle_unpack.go
```diff
@@ -18,15 +20,60 @@ const (
 	fileDownloadsPerConcurrentChunks = 3
 )
 
-func unpackBundleDescriptor(ctx context.Context, bundle *Bundle) error {
-
-	bundleDescriptorBuffer, err := storage.ReadTee(ctx,
-		bundle.MetaStore, model.GetArchivePathToBundle(bundle.RepoID, bundle.BundleID),
-		bundle.ConsumableStore, model.GetConsumablePathToBundle(bundle.BundleID))
+func setBundleIDFromConsumableStore(ctx context.Context, bundle *Bundle) error {
```
kerneltime: Why not getBundleIDFromPath?

ransomw1c: i've added a `getBundleIDFromPath` function that returns a bundle id given the path to a bundle descriptor within the bundle (and the empty string when given another path such as a filelist path) and used it within `setBundleIDFromConsumableStore`.

it occurs to me that what `getBundleIDFromPath` could also mean is to get the bundle id from a path to a bundle on the local filesystem, except i suppose keeping the abstraction away from the local filesystem via `storage` is a desirable abstraction.

also, i could inline the contents of `setBundleIDFromConsumableStore` since it's only used once .. or not:  i'm ambivalent re. the aesthetics of whether to maintain `setBundleIDFromConsumableStore` as a separate function or not.

pkg/model/bundle.go
ransomw1c: have added

PR #219, opened 2019-06-21
distribute web static content in binary

ransomw1c:
this diff is to prepare the `web` command for distribution as a binary building static assets (templates, css, etc.) into the binary via [`packr`](https://github.com/gobuffalo/packr/tree/master/v2), part of #199


ransomw1c:
recalled while implementing this that [`fileb0x`](https://github.com/UnnoTed/fileb0x) is a similar library to `packr` .. depending on the experience with `packr`, we might opt to use `fileb0x` instead.

PR #222, opened 2019-06-24
init bundle update

ransomw1c:
the `bundle update` part of #214

### DIFF
--- a/pkg/core/bundle_unpack.go
+++ b/pkg/core/bundle_unpack.go
@@ -8,6 +8,8 @@ import (
 	"io"
 	"io/ioutil"
 
+	//	"regexp"
+
 	"github.com/oneconcern/datamon/pkg/cafs"
 	"github.com/oneconcern/datamon/pkg/model"
 	"github.com/oneconcern/datamon/pkg/storage"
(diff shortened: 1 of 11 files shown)


pkg/core/bundle_unpack.go
```diff
@@ -60,72 +164,305 @@ type errorHit struct {
...
+	if err != nil {
+		return err
+	}
+	reader, err := fs.Get(ctx, key)
+	if err != nil {
+		return err
+	}
+	var putParameterOverwrite bool
+	if overwrite {
+		err = bundle.ConsumableStore.Delete(ctx, bundleEntry.NameWithPath)
```
ransomw1c: all of the internal api in this file is kinda of a disaster area:  what's convenient to debug, isn't always convenient to maintain.

using `Delete` and `Put` calls rather than calling `Put` with a `storage.Overwrite` parameter is particularly egregious, however:  the `storage.Overwrite` parameter does something different in CI (linux) than local (mac), and this needs to be tested and fixed as a separate issue.

pkg/core/bundle_unpack.go
```diff
@@ -60,72 +164,305 @@ type errorHit struct {
...
 	select {
 	case eh := <-errC:
 		return eh.error
-	default:
-		return nil
+	case <-doneOkC:
+	}
+
+	// rewrite destination bundle metadata
+	if bundleDest != nil {
```
ransomw1c: this conditional could additionally check for any empty diff.

pkg/core/bundle_unpack.go
```diff
@@ -60,72 +164,305 @@ type errorHit struct {
...
-				return
+	} else {
+		bundle.l.Info("downloading diff entries",
+			zap.Int("num", len(diff.Entries)))
+		for _, de := range diff.Entries {
+			concurrencyControl <- struct{}{}
+			switch de.Type {
+			case DiffEntryTypeAdd:
+				bundle.l.Info("adding added entry",
+					zap.String("name Additional", de.Additional.NameWithPath),
```
ransomw1c: throughout this `switch`, `de.Name` can replace `de.Additional.NameWithPath` and `de.Existing.NameWithPath`

pkg/core/bundle_unpack.go
```diff
@@ -284,6 +463,29 @@ func downloadBundleEntrySync(ctx context.Context, bundleEntry model.BundleEntry,
...
+			zap.String("name", bundleEntry.NameWithPath),
+			zap.Error(err))
+		return err
+	}
+	bundle.l.Info("deleted bundle entry",
+		zap.String("name", bundleEntry.NameWithPath))
+	return nil
+}
+
+// dupe: deleteBundleEntry
```
ransomw1c: not focusing on deduping this immediately...

PR #225, opened 2019-06-28
[wip] init web deploy

ransomw1c:
further progress on #199 , this is about deploying the web ui to a pod browsing at a uri accessible within the One Concern network.

currently, users are advised to install from the releases page.


ransomw1c:
this functionality is off the radar in terms of currently desired features

Issue #226, opened 2019-07-01
RO FS FUSE dbg, perf


ransomw1c:
there are various and sundry debuggings and performance improvements to the `bundle mount` command as used to mount a read-only bundle and stream from GCS:

- [ ] memory leaks on streaming read-only mount [1]
- [ ] memory usage during `core.populateFS()` [2]
- [ ] `populateFS` could be faster if parallelized [3]

`backup-filestore-output` and `1LqWPRbwaV6JXKIxJrYUrUV6ZIp` are a suitable repo and bundle to use for benchmarking.

ransomw1c:
### additional notes

##### 1

move the benchmark bundle with `rsync` to duplicate this error

##### 2

there are several maps and/or radix trees attached to the read-only fs implementation struct.  some of these (e.g. `fsDir`) are only used during `populateFS`.  since these in-memory key-value stores are referenced by the implementation struct, they aren't GCed after `populateFS`.  hence, making members of the struct function-local variables will allow GC and reduce memory usage after `populateFS` returns.

### possible approaches

##### 1

consider `seekAhead` in `chunkReader.ReadAt()` in `cafs`.  particularly, the anonymous goroutine ought pass errors to an error channel.

review entire pathway from FUSE IOC to `(*Bundle).ReadAt()` (and perhaps elsewhere?) to `chunkReader.ReadAt`
.. are the bytes read counts correct? ..
.. is `copy(p[n:], buffer[offset:])` creating slices that aren't getting collected on loop iterations? ..
.. does removing _all_ `lru.Contains` usage matter? ..
.. and is there some synchronization issue with github.com/hashicorp/golang-lru usage?

##### 3

part and parcel of parallelization is further decomposition into simpler functions.

ransomw1c:
monitoring memory use via `htop` during reads from the streaming read-only FUSE fs as in #227 show a bounded memory footprint according to the number of parallel processes (either `cp` or `rsync` reading bytes out of the filesystem).  in addition to the memory use after `populateFS`, additional memory scales according to number of processes like
* 10 : 8GiB
* 20 : 10GiB
* 40 : 11GiB
so there's more memory required to carry out more parallel reads, which is what we'd expect given that there's some buffering required for each read.  the non-linear growth of memory could be due to details of gcs or other upstream libraries.
[...]

PR #228, opened 2019-07-08
[wip] bundle.ReadAt memory usage

ransomw1c:
toward #226 , this merge request is about reducing the surprisingly high (GiB) memory usage found on parallel reads of the Read-Only FUSE filesystem (`bundle mount`) under the hypothesis that the memory usage occurs due, at least in part, to implementation between the blob storage implementation and the `bundle.ReadAt` function called by the FUSE IoC mechanism.  in particular, the GCS storage and FUSE are not sources of memory usage to be measured and reduced here.

### DIFF
--- a/cmd/metrics/cmd/bundle_read_at.go
+++ b/cmd/metrics/cmd/bundle_read_at.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"runtime/pprof"
 	"time"
+	"strconv"
 
 	"github.com/oneconcern/datamon/pkg/core"
 	"github.com/oneconcern/datamon/pkg/dlogger"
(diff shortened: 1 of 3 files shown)


ransomw1c:
current usage is along the lines of
```
metrics% go build
datamon% ./cmd/metrics/metrics bundle-readat
datamon% go tool pprof -svg cmd/metrics/metrics mem.prof/bundlereadat-3.mem.prof && mv profile001.svg bundlereadat.svg
```
where `bundlereadat-3.mem.prof` may be one of several `bundlereadat-*.mem.prof` profiles
... and indicates that internal buffers aren't eating too much memory under the assumptions (no FUSE, no GCS) of this merge request.

ransomw1c:
archiving

PR #227, opened 2019-07-02
[wip] improve sidecar memory usage

ransomw1c:
to address either [1] or [2] of #226 , this draft is currently considered PoC and/or spike

PR #229, opened 2019-07-11
buffer pool cafs reader ReadAt

ransomw1c:
extract buffer pool implementation from #227 to address [1] on #226

### DIFF
--- a/pkg/cafs/reader.go
+++ b/pkg/cafs/reader.go
@@ -3,8 +3,8 @@ package cafs
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
-	"io/ioutil"
 	"strings"
 	"sync"
 
@@ -51,6 +51,12 @@ func ConcurrentChunkWrites(concurrentChunkWrites int) ReaderOption {
 	}
 }
 
+func SetLeafPool(leafPool *leafFreelist) ReaderOption {
+	return func(reader *chunkReader) {
+		reader.leafPool = leafPool
+	}
+}
+
 func newReader(blobs storage.Store, hash Key, leafSize uint32, prefix string, opts ...ReaderOption) (Reader, error) {
 	c := &chunkReader{
 		fs:                    blobs,
@@ -94,6 +100,7 @@ type chunkReader struct {
 	currLeaf              []byte
 	verifyHash            bool
 	lru                   *lru.Cache
+	leafPool              *leafFreelist
 	concurrentChunkWrites int
 }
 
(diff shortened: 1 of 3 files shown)


ransomw1c:
previously, when reading files from `bundle mount` with the default `stream=true` setting, there were some allocations at the CAFS layer that the go GC wasn't keeping track of:  `ioutil.ReadAll` allocated a buffer, and that buffer was then added to the [lru cache](https://godoc.org/github.com/hashicorp/golang-lru), sliced in order to `copy()`, and, in short, passed through enough machinery that the golang GC, according to the [the `"heap"` profile](https://golang.org/pkg/runtime/pprof/#Profile), lost track of the allocation (although diffing `"alloc"` profiles via `go tool pprof --base first_alloc.mem.prof datamon later_alloc.mem.prof` did reveal that this was the cause of the memory use reported by the operating system).

it might be possible to tune the existing implementation in such a way that the GC can keep track of the allocations, as there are [several techniques](https://go101.org/article/memory-leaking.html) to be aware of in this regard.  for instance, subslices keep around memory for the entire slice, so the `append([]int(nil), s1[len(s1)-30:]...)` idiom is sometimes necessary to use rather than a subslice itself `s1[len(s1)-30:]`.  it's also possible that the GC doesn't realize it needs to free the memory due to implementation details of the lru cache.

all considered, it was viewed as more convenient to opt into semi-manual memory management with the buffer pool implementation in `freelists.go`:  since we know what the lifespan of the buffers in CAFS should be, we're able to encode the memory management in a `malloc()` and `free()` style without needing to be familiar with GC arcania (that is, arcane facts).  later, after getting more hip to GC facts, we could try setting the memory pool to `nil`:  in this case, it will always allocate `new()` buffers rather than attempting to maintain a pool.

pkg/cafs/reader.go
```diff
@@ -200,73 +207,93 @@ func calculateKeyAndOffset(off int64, leafSize uint32) (index int64, offset int6
...
+		}
+		lb := r.leafPool.get()
+		var n int
+		for {
+			m, e := rdr.Read(lb.slice[n:cap(lb.slice)])
+			if m < 0 {
+				panic(fmt.Errorf("read negative number of bytes"))
+			}
+			n += m
+			lb.slice = lb.slice[:n]
```
kerneltime: I have some questions around this, not sure why you need to do ```lb.slice = lb .slice[:n]```

kerneltime:
The code change looks clean and nice. Have asked a question, will discuss in person.

pkg/cafs/reader.go
ransomw1c: plan to resolve this:   use `len(lb.slice)` instead of `n` .. this'll add some compute time while making the code more readable.

incidentally, this implementation was originally lifted from [the implementation](https://golang.org/src/bytes/buffer.go) of [`bytes.Buffer.ReadFrom`](https://golang.org/pkg/bytes/#Buffer.ReadFrom), which is what the original `ioutil.ReadAll` call it replaces uses, and the copyover was done without too much thought.

PR #230, opened 2019-07-11
parallelize bundle file list download

ransomw1c:
part of [3] on #226 , originally started as part of #227 , this diff parallelizes the download of filelists, the bulk of a bundle's metadata.  so although it's not strictly part of parallelizing `populateFS` as mentioned in [3], it's within the spirit of what that item is about:  decreasing the time from starting `bundle mount` to the FUSE mount occurring.

it will also speed up `bundle download`  somewhat as well as (wip at time of creating merge request) `bundle update`.

### DIFF
--- a/cmd/datamon/cmd/bundle_download.go
+++ b/cmd/datamon/cmd/bundle_download.go
@@ -10,7 +10,8 @@ import (
 )
 
 const (
-	fileDownloadsByConcurrencyFactor = 10
+	fileDownloadsByConcurrencyFactor     = 10
+	filelistDownloadsByConcurrencyFactor = 10
 )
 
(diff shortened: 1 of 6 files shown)


ransomw1c:
this implementation goes to some effort to ensure that the `BundleEntries` slice is stored in memory in the same order as it is on disk.  a possible future improvement is to drop the assumption that filelists on disk (or subslices or `BundleEntries`) are all of some fixed length, except perhaps the last filelist.  by dint of only slightly more effort when programming -- and computation when downloading bundle entries -- it would be possible to additionally allow for variable-length filelists on disk.  i could implement that instead if it's preferable.

Issue #231, opened 2019-07-15
label direct lookup


ransomw1c:
add a `label get` command that looks up bundle information by label.

the implementation details are distinct from piping `label list` to `grep` because directly accessing labels by name is different than filtering a list of labels:  labels are stored in a GCS bucket that allows directly access by a key that can be constructed from the label name.  so `label get` is going to be faster than filtering through all labels.

ransomw1c:
if a label doesn't exist, exit with `syscall.ENOENT`

fredbi:
this works now. Closing

PR #224, opened 2019-06-27
dedupe cmd Run funcs (stores)

ransomw1c:
nominally toward #165 , this diff deduplicates creation of objects based on parameters in the `Run:` functions of the various `cobra.Command` structs in the `cmd`/CLI package.

this harmonization might in particular have implications for implementing IPC as discussed in [this comment](https://github.com/oneconcern/datamon/issues/214#issuecomment-506448369) on #214 , the `bundle update` iss.

PR #233, opened 2019-07-16
init label get cmd

ransomw1c:
implement `label get` per #231

### DIFF
--- a/cmd/datamon/cmd/label_get.go
+++ b/cmd/datamon/cmd/label_get.go
@@ -1,0 +1,67 @@
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/oneconcern/datamon/pkg/core"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/sys/unix"
+)
+
+var GetLabelCommand = &cobra.Command{
+	Use:   "get",
+	Short: "Get bundle info by label",
+	Long: `Performs a direct lookup of labels by name.
+Prints corresponding bundle information if the label exists,
+exits with ENOENT status otherwise.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		remoteStores, err := paramsToRemoteCmdStores(params)
+		if err != nil {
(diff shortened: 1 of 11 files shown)


ransomw1c:
current demo

```shell
cmd/datamon% go build && ./datamon label get --repo ransom-datamon-test-repo --label asdf ; echo "$?" && ./datamon label get --repo ransom-datamon-test-repo --label testlabel ; echo "$?"
```

ransomw1c:
leaving the WIP tag on this to hopefully rebase and dedupe after #224 ...

... also, want to pause for particular opinions about what to print in the case of an extant label.  same format as `label list`?

kerneltime:
> #224



> leaving the WIP tag on this to hopefully rebase and dedupe after #224 ...
> 
> ... also, want to pause for particular opinions about what to print in the case of an extant label. same format as `label list`?
Sounds good.

cmd/datamon/cmd/label_get.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+var GetLabelCommand = &cobra.Command{
+	Use:   "get",
+	Short: "Get bundle info by label",
+	Long: `Performs a direct lookup of labels by name.
+Prints corresponding bundle information if the label exists,
+exits with ENOENT status otherwise.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if params.repo.ContributorEmail == "" {
```
kerneltime: Thought not enforced uniformly, we should allow read operations without contributor information being set.

ransomw1c: yes, this was removed altogether

pkg/storage/gcs/store.go
```diff
@@ -58,7 +58,7 @@ func (g *gcs) Has(ctx context.Context, objectName string) (bool, error) {
 	}
```
kerneltime: The method should be updated to use the client that is part of the [gcs instance](https://github.com/oneconcern/datamon/pull/233/files#diff-aaf1bc657e525c33e0d1422846c07b12R97)

PR #236, opened 2019-07-22
[wip] init CI image

ransomw1c:
this corresponds to some internal One Concern CI setup paradigms.

see [the jira iss](https://oneconcern.atlassian.net/browse/UN-468) rather than github in this case.


ransomw1c:
archiving

PR #240, opened 2019-07-29
debug desktop install shell script

ransomw1c:
debug `deploy/datamon.sh` and update the readme.

### DIFF
--- a/deploy/datamon.sh
+++ b/deploy/datamon.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
 
 ### datamon.sh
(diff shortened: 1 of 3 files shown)


deploy/datamon.sh
```diff
@@ -22,14 +22,38 @@ CONFIRM_INSTALL=
...
+        echo "required auxilliary program '$*' not installed" 1>&2
+        exit 1
+    fi
+}
+
+ensure_aux wget
+ensure_aux curl
+#
+ensure_aux bc
+ensure_aux md5sum
```
kerneltime: on mac this should use md5 which comes with the OS.

Issue #239, opened 2019-07-27
Bundle upload is broken


kerneltime:
Latest master does not upload a large directory correctly, none of the file lists get uploaded.

ransomw1c:
@kerneltime are we satisfied that this is resolved?

fredbi:
closing this now: it has been fixed long ago

PR #241, opened 2019-07-30
test filelist upload

ransomw1c:
this diff uses a mock destination (blob and metadata) `storage.Store` in `cmd/metrics` in order to debug #239 .

additionally, several debug-level logging statements are added to the bundle upload implementation.

### DIFF
--- a/pkg/core/bundle_pack.go
+++ b/pkg/core/bundle_pack.go
@@ -5,10 +5,8 @@ package core
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"hash/crc32"
 	"io"
-	"log"
 	"os"
 	"path/filepath"
 	"runtime/pprof"
@@ -34,6 +32,7 @@ type filePacked struct {
 	keys      []byte
 	size      uint64
 	duplicate bool
+	idx       int
 }
 
 func filePacked2BundleEntry(packedFile filePacked) model.BundleEntry {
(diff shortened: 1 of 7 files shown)


ransomw1c:
likely #239 duplication in this branch via

```
./cmd/metrics/metrics upload --num-files 100000 --filesize .01 &> ~/tmp/metricsupload00.log
cat ~/tmp/metricsupload00.log  |grep 'mock destination store Put'  | grep 'bundles/repo[^ ]*bundle-files[^ ]*json' |wc -l
```

should be 100, while runs on bare metal show that it's consistently less than that.  each test run produces a different value.

ransomw1c:
the previous comment appears to be an error out of the logger.  with

```
	if isFilelist {
...
		fmt.Println("mock dest store Put got filelist")
	}
	mds.l.Info("mock destination store Put()",
...
```

in the proxy `storage.Store`, there are log entries of both the form (for the first several filelist uploads in each log)

```
mock dest store Put got filelist
{"level":"info","ts":1564594789.889411,"caller":"core/bundle_pack.go:294","msg":"uploading file list","curr file index":64999,"total upload attempts so far":65,"BundleEntriesFileCount":64}
```

and (for every filelist upload thereafter)

```
mock dest store Put got filelist
{"level":"info","ts":1564594789.917655,"caller":"cmd/mock_dest_store.go:55","msg":"mock destination store Put()","key":"bundles/repo-as9e130mfaubs3r/1Omto0yD5JLibo2kWt7GB3lPowY/bundle-files-65.json","isFilelist":true,"store name":"meta","fileListUploadPutCnt":66}
{"level":"info","ts":1564594789.917686,"caller":"core/bundle_pack.go:294","msg":"uploading file list","curr file index":66999,"total upload attempts so far":67,"BundleEntriesFileCount":66}
```

in other words, the logger is not printing the messages outside the conditional, while the `fmt.Println` in the conditional is printing the correct number of filelist uploads.

pkg/core/bundle_pack.go
```diff
@@ -158,16 +166,25 @@ func uploadBundleFiles(
 				error: err,
 				file:  file,
 			}
+			break
```
ransomw1c: continuing through the loop in case of an error when `--skip-on-error` is not set causes a bug in error reporting.

although fixing this is a conceptually separate idea from the original intent this request, i've included it here because, overall, it does connect to debugging `bundle upload`.

kerneltime: bundle.skipFile is to skip internal files, not sure how this is related to --skip-on-error

PR #242, opened 2019-07-30
remove error signal channel from upload and download implementations

ransomw1c:
this diff removes the `doneC` from the implementations of `bundle upload` and `download`.

the purpose of this channel is to broadcast errors from the "main thread" to all goroutines such that channel operations expecting to sync with the main thread do not park indefinitely in the case of an error with the net result of goroutine leaks.

meanwhile, in the case of many goroutines being run at once, it's been observed that, without the `doneC`, uploads and downloads don't work.  we could speculate that the desired non-blocking behavior in the case of errors is occurring when there are not errors.

in any case, since the overall binary typically exits on the sort of error conditions that the `doneC` is used to signal, leaks are secondary to the issue #239 of correctness at scale:  we needn't worry so much about whether there are leaked goroutines when the system can clean up after the process holistically.

### DIFF
--- a/pkg/core/bundle_pack.go
+++ b/pkg/core/bundle_pack.go
@@ -49,8 +49,6 @@ type uploadBundleChans struct {
 	// recv data from goroutines about uploaded files
 	filePacked chan<- filePacked
 	error      chan<- errorHit
-	// broadcast to all goroutines not to block by closing this channel
-	done <-chan struct{}
 	// signal file upload goroutines done by writing to this channel
 	doneOk             chan<- struct{}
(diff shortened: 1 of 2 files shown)


kerneltime:
Have we tested this against a large repo for upload and download? Were you able to reproduce this ?

ransomw1c:
i've rebased #241 onto this patch, and still see similar results wherein some but not all `bundle-files-*` are `Put` to the mock store.

kerneltime:
So we still need a fix, can we find a release where this bug is not present and only host that on the released page

PR #243, opened 2019-08-01
init datamover image and demo yaml

ransomw1c:
as with the [sidecar](pod/datamon-datamover-test-5b87d6d949-8svcr) used in Argo workflows, this is an operationalization of datamon in the sense that it's a use of datamon in order to accomplish a specific task.  in addition to the datamon program itself, an operationalization consists of any necessary auxiliary programs (for example, the `migrate` program used to create and act on lists of files in this case) and a wrapper script (shell scripts here and in the sidecar, although Python or similar could be used as well), all packaged as a Docker image along with sample Kubernetes YAML, and documented such that datamon users get an interface specific to their needs.

the particular use-case here at One Concern that we're currently operationalizing is backing up and cleaning an NFS filestore shared among various data-science pipelines and ad-hoc experiments.  while datamon is increasingly useful as the primary storage mechanism for some data-science pipelines, it's not feasible to replace the NFS store wholesale.  additionally, for ad-hoc data-science scripts (those that aren't part of pipelines), `datamon`s archiving and tracking mechanisms can be more of a hindrance than an asset.  so we wind up with a large (approx 10^7 files totaling in 32-64TB) NFS store that slowly fills up with data.

the operation here is to backup all data in the NFS store to datamon as well as to remove (unlink) all files with a modify time before some specified time.

the current scope of the request is a PoC of the operationalization.  scalability appears (`./hack/datamover/create_demo_pod.sh -s <size_tb> -n <num_files>` in order for statistics of up to at least around 1 TB and 10^5 files.  additional debugging and optimization might be required to bring the operation up to the requirements NFS store backup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,25 @@ build-and-push-fuse-sidecar:
 		.
 	docker push gcr.io/onec-co/datamon-fuse-sidecar
 
+
+.PHONY: build-and-push-datamover
+## build sidecar container used in Argo workflows
+build-and-push-datamover:
+	@echo 'building fuse sidecar container'
+	docker build \
+		--progress plain \
+		--build-arg version_import_path=$(VERSION_INFO_IMPORT_PATH) \
+		--build-arg version=$(VERSION) \
+		--build-arg commit=$(COMMIT) \
+		--build-arg dirty=$(GITDIRTY) \
+		-t gcr.io/onec-co/datamon-datamover \
+		-t gcr.io/onec-co/datamon-datamover:${GITHUB_USER}-$$(date '+%Y%m%d') \
+		-t gcr.io/onec-co/datamon-datamover:$(subst /,_,$(GIT_BRANCH)) \
+		--ssh default \
+		-f datamover.Dockerfile \
+		.
+	docker push gcr.io/onec-co/datamon-datamover
+
 .PHONY: build-datamon
 ## Build datamon docker container (datamon)
 build-datamon:
--- a/README.md
+++ b/README.md
@@ -223,6 +223,37 @@ chmod +x datamon
 It's probably most convenient to have the wrapper script placed somewhere on your
 shell's path, of course.
 
+
+# Datamover container guide
+
+As with the [Kubernetes sidecar guide](#kubernetes-sidecar-guide), this section covers
+a particular operationalization of Datamon at One Concern wherein we use the program
+along with some auxilliary programs, all parameterized via a shell script and shipped
+in a Docker image, in order to periodically backup a shared block store and remove
+files according to their modify time.
+
+The docker image is called `gcr.io/onec-co/datamon-datamover` and is tagged with
+versions just as the Kubernetes sidecar, `v<release_number>`, where `v0.7` is the first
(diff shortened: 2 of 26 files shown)


ransomw1c:
here's a cursory exercise of the `filelist-actions` verb to the `migrate` program that'll be used to remove entries from NFS according to modify time and backup to datamon

```
#! /bin/zsh

BIN=${HOME}/projects/ransomw1c/datamon/cmd/backup2blobs/backup2blobs
TEST_DIR=samplefiles

if [[ -d $TEST_DIR ]]; then
    rm -rf $TEST_DIR
fi

mkdir $TEST_DIR

TS_A=$(date +%y%m%d%H%M%S)

sleep 2

touch $TEST_DIR/afile

sleep 3

TS_B=$(date +%y%m%d%H%M%S)

sleep 2

touch $TEST_DIR/bfile

sleep 2

TS_C=$(date +%y%m%d%H%M%S)

rm *.list

find $TEST_DIR -type f  | $BIN filelist-actions --out all.list
find $TEST_DIR -type f | $BIN filelist-actions --out filter_a.list --time-before $TS_A
find $TEST_DIR -type f | $BIN filelist-actions --out filter_b.list --time-before $TS_B
find $TEST_DIR -type f | $BIN filelist-actions --out filter_c.list --time-before $TS_C

print "found $(wc -l all.list) files unfiltered, expected 2"
print "found $(wc -l filter_a.list) files in filter a, expected 0"
print "found $(wc -l filter_b.list) files in filter b, expected 1"
print "found $(wc -l filter_c.list) files in filter c, expected 2"

print
print "filter_b.list consists of"
cat filter_b.list
print "expected /afile"
```

ransomw1c:
## usage notes

these are some notes on how the datamover is being used in development.  the plan is to coalesce these notes into documentation in the readme.

### metrics/benchmarking

metrics collection is done via a kubernetes Deployment for the convenience of`exec`ing into a persistent shell during and after the metrics collection to review logs, view OS status, etc.

*todo*

ransomw1c:
i've added a couple of shell scripts as a separate commit from the initial approval.

there's also a debug of the `filelist-actions` verb in `migrate`, creds from local added to the datamover job (to prevent the `googleapi: Error 403: Insufficient Permission, insufficientPermissions` noticed by @mkiadi2002 ), and the zsh version of `set -e` added to all in-container scripts.

using zsh is a deliberate choice here, by the way, at least for the in-container scripts, because the container environment guarantees that zsh is available.

### `access_nfs.sh`

provides access to a kubernetes shell with the NFS share, the datamover toolkit, and some additional niceties available.

##### opts
* `-c` creates a new deployment.  run without this option to drop into a shell supposing that the script has already been run with `-c`.

### `nfs_lister.sh`

allows listing files in the NFS share in the local (desktop) environment.

##### opts

* `-n` pvc name (defaults to `depmon-pvc`, not `pvc1`)
* `-p` path to list.  if relative, it's relative to the pvc root
* `-o` list output file on local.  if not set, writes filenames to standard out prefixed by `item:`.

----

i'll open a separate request to continue evolving the UI design of the overall NFS operationalization.

PR #245, opened 2019-08-06
[wip] [const] sidecar bundle id param

ransomw1c:
this adds an `-i` for bundleID parameter to the `wrap_datamon.sh` script used by the FUSE RO FS sidecar by request for the seismic Argo pipeline.

use of this parameter supposes precisely one `bundle upload` command is passed to `wrap_datamon.sh` via the `-d` flag and writes the bundle id of the uploaded file to the file specified as the `-i` parameter.  this file can then be used as the [output parameter](https://github.com/argoproj/argo/tree/master/examples#output-parameters) of the Argo DAG node using the sidecar setup.

### DIFF
--- a/hack/fuse-demo/wrap_datamon.sh
+++ b/hack/fuse-demo/wrap_datamon.sh
@@ -268,7 +268,7 @@ for upload_cmd in $UPLOAD_CMDS; do
     IFS="$ifs_orig"
     bundle_id=$(2>&1 run_datamon_cmd "$upload_cmd" | \
                     grep 'Uploaded bundle id' | \
-                    head -1 | \
+                    tail -1 | \
                     sed 's/Uploaded bundle id:\(.*\)/\1/')
     upload_status="$?"
     if [ "$upload_status" != 0 ]; then
--- a/hack/fuse-demo/wrap_datamon.sh
+++ b/hack/fuse-demo/wrap_datamon.sh
@@ -11,8 +11,9 @@ POLL_INTERVAL=1 # sec
 
 COORD_POINT=
 DATAMON_CMDS=
+BUNDLE_ID_FILE=
 
-while getopts c:d: opt; do
+while getopts c:d:i: opt; do
     case $opt in
         (c)
             COORD_POINT="$OPTARG"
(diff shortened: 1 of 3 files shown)


ransomw1c:
a more durable implementation across requirements of the sidecar setup would be to allow a post-processing script to be passed to the `wrap_application.sh` and a generalized log file to be passed to `wrap_datamon.sh`.  the Argo pipeline would then be responsible for munging a the log file into the any form of output file it expects.  this requirement could possibly be spun off as a separate issue.

hack/fuse-demo/wrap_datamon.sh
```diff
@@ -245,11 +266,22 @@ ifs_orig="$IFS"
 IFS=':'
 for upload_cmd in $UPLOAD_CMDS; do
     IFS="$ifs_orig"
-    if ! run_datamon_cmd "$upload_cmd"; then
-        upload_status="$?"
+    bundle_id=$(2>&1 run_datamon_cmd "$upload_cmd" | \
+                    grep 'Uploaded bundle id' | \
+                    head -1 | \
+                    sed 's/Uploaded bundle id:\(.*\)/\1/')
```
amarkovits-1C: hey so looks like the output of `run_datamon_cmd` looks like this: 
```
'{"level":"info","ts":1565158923.6535618,"caller":"core/bundle_pack.go:307","msg":"Uploaded
            bundle id","BundleID":"1P5LHRMo8EipZGg5FJW193e6SkS"}'
```
so the sed command will have to change a bit

ransomw1c: there are, conceptually, at least two bytestreams getting muxed into stdout/stderr:  the `zap` logger ( https://github.com/oneconcern/datamon/blob/master/pkg/core/bundle_pack.go#L307 ) and the stdlib `log` logger ( https://github.com/oneconcern/datamon/blob/master/cmd/datamon/cmd/bundle_upload.go#L85 ).

the stdlib logger, the one that prints `Uploaded bundle id:`, is the bytestream used in test ( https://github.com/oneconcern/datamon/blob/master/cmd/datamon/cmd/cli_test.go#L335 ), so that's the bytestream the regexp is written for.

amarkovits-1C: Got it, I think that makes sense.
Is there a way to pass the bytstream from the stdlib logger instead of the zap logger to the regex? At the moment the text being written to the 'bundle_id' file (use as an argo output parameter) is the same as the zap logger output

ransomw1c:
preempted by #246

PR #247, opened 2019-08-09
[wip] [const] Docker Desktop sidecar usage

ransomw1c:
try out (and document as necessary) the means of using the `fuse-ro-demo` (Makefile target) on the Docker for Desktop kubernetes cluster rather than Google Kubernetes Engine only.

### DIFF
--- a/sidecar.Dockerfile
+++ b/sidecar.Dockerfile
@@ -32,6 +32,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends fuse &&\
   apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 RUN echo "allow_root" >> /etc/fuse.conf
 
+# RUN echo "user_allow_other" >> /etc/fuse.conf
+
 COPY --from=base /stage /
 ENV ZONEINFO /zoneinfo.zip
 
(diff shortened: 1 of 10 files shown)


sidecar.Dockerfile
```diff
@@ -32,6 +32,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends fuse &&\
   apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 RUN echo "allow_root" >> /etc/fuse.conf
 
+# RUN echo "user_allow_other" >> /etc/fuse.conf
```
amarkovits-1C: Does this seem to do the trick? (of giving permission to the sidecar container to /dev/fuse in docker-for-desktop context)

ransomw1c: no, it does not.  still unclear about the linkage between https://github.com/jacobsa/fuse and [`fusermount`](http://man7.org/linux/man-pages/man8/mount.fuse.8.html) if any.  i have read much of the internal message-passing dataflow in jacobsa/fuse, yet the filesystem io to `/dev/fuse` is still unread.

it could be that some options need to be passed to jacobsa/fuse rather than set in `/etc/fuse.conf`.

overall, i am supposing that maintaining a single image with a non-root user rather than shipping two separate sidecar images is an option.  indeed, perhaps even an easier option.

ransomw1c:
these kubernetes-specific attempts at allowing access to `/dev/fuse` in Docker Desktop environments are supplanted by using `sudo` to change `/dev/fuse` perms from within the sidecar in #252

PR #249, opened 2019-08-13
init build-and-push-release-docker-images Makefile target

ransomw1c:
slight automation increment while setting [v0.7](https://github.com/oneconcern/datamon/releases/tag/v0.7)

the additional Makefile target included builds both [the Argo sidecar](https://github.com/oneconcern/datamon#kubernetes-sidecar-guide) and [the NFS datamover](https://github.com/oneconcern/datamon#datamover-container-guide) and deploys to the repo with a release tag.

the release tag is the same as that used by git when `HEAD` is on such a tag.  otherwise, it's `hash-<hash>`, where `<hash>` is the abbreviated hash out of git.

### DIFF
--- a/hack/release_tag.sh
+++ b/hack/release_tag.sh
@@ -1,0 +1,49 @@
+#! /bin/zsh
+
+setopt ERR_EXIT
+
+is_latest=false
+
+while getopts l opt; do
+    case $opt in
+        (l)
+            is_latest=true
+            ;;
+        (\?)
+            print Bad option, aborting.
+            exit 1
+            ;;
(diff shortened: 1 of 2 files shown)


hack/release_tag.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/zsh
+
+setopt ERR_EXIT
+
+typeset -a version_tags
+
+version_tags=($(git tag --points-at HEAD |grep '^v'))
```
ransomw1c: add `--format=` param?

PR #254, opened 2019-08-20
[wip] regexp filter on bundle download

ransomw1c:


pkg/core/bundle.go
```diff
@@ -208,17 +208,22 @@ func New(bd *model.BundleDescriptor, bundleOps ...BundleOption) *Bundle {
 
 // Publish an bundle to a consumable store
 func Publish(ctx context.Context, bundle *Bundle) error {
-	return implPublish(ctx, bundle, defaultBundleEntriesPerFile)
+	return implPublish(ctx, bundle, defaultBundleEntriesPerFile, func(s string) (bool, error) { return true, nil })
+}
+
+func PublishSpecificKeys(ctx context.Context, bundle *Bundle, keyCond func(string) (bool, error)) error {
```
kerneltime: rename to PublishSelectBundleEntries and the function can be SelectionFilter?

ransomw1c: i'm okay with using 'select' verbiage here, although the corresponding upload function uses the word 'specific'.

calling a function that returns a boolean value a filter is going to be confusing for me.  i'm used to filter being the function that does the filtering, and the predicate (the word i've used) or conditional being the thing that determines the truthiness of sequence elements.

Issue #256, opened 2019-09-07
postgres sidecar


ransomw1c:
we have a sidecar-like use-case wherein the source and destination is a postgres database rather than datamon.  the task here to provide, either in the sidecar container or a different container, a user interface to make such an integration seamless with argo.

in particular, to store data to datamon, we have two use-cases for backup:
* backup selected tables from a database using `pg_dump`
* backup an entire database closer to the [fs level](https://www.postgresql.org/docs/11/backup-file.html)

the latter is not to be done exactly as in the linked documentation, because we can't guarantee that the database is vacuumed before the backup operation.  so we'll need some way (auxiliary program) to remove any sparseness when streaming the files out of the filesystem as well as any blocks that the db's b-tree has marked as invalid, hence a vendor-specific way.

restore is simpler, since the destination database will not contain conflicts:  it'll be necessary to take the artifact (possibly one of several types, suggesting a tagged union alike for the data within datamon's blob store) from datamon and add it to an empty postgres datadir.

fredbi:
@ransomw1c 
> because we can't guarantee that the database is vacuumed before the backup operation. so we'll need some way (auxiliary program) to remove any sparseness when streaming the files out of the filesystem as well as any blocks that the db's b-tree has marked as invalid, hence a vendor-specific way.

with Postgres, this is carried out using the admin `CHECKPOINT` command under `psql`. See https://www.postgresql.org/docs/current/sql-checkpoint.html

fredbi:
ok this is done now

PR #258, opened 2019-09-17
Pgsidecar with an init script and a tear down script

divya1c:
Init script downloads data dump from datamon and initializes the db.
Teardown script makes a data dump from the db and uploads to datamon.

Also included is a sample k8s deployment file for how the image can be used.
This can be included as a sidecar in an argo template to spin up local dbs with a known input and storing the output in datamon for some other task to pick up.


pgsidecar/sampleK8s/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+          - containerPort: 5432
+            name: psql-port
+        resources: {}
+        stdin: true
+        tty: true
+        volumeMounts:
+        - mountPath: /etc/google/
+          name: gcs-credentials
+        lifecycle:
+          preStop:
```
kerneltime: Look into ```terminationGracePeriodSeconds```
Eventually wrapping the application logic to detect an exit and then stopping the DB will be more reliable.

ransomw1c:
for consistency with the rest of the repo, please use the following layout:

* dockerfiles corresponding to images that we ship as part of releases are given `{name}.Dockerfile` and located in the project root
* dockerfiles that are used for specific demos are located in`hack/{demo-name}/{name}.Dockerfile`
* all kubernetes yaml goes in `hack/k8s`, either of the form `{name}.yaml` if it's valid as-is or `{name}.template.yaml` if it's templated with environment variables
* shell scripts go in `hack/{demo-name}`

pgsidecar/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+ENV GITHUB_USER ${github_user}
+ENV GITHUB_TOKEN ${github_token}
+
+RUN apt-get update \
+  && apt-get install -y wget \
+  && apt-get install -y postgresql-11-postgis-scripts
+
+RUN set -e && \
+  wget https://github.com/oneconcern/datamon/releases/download/v0.7/datamon.linux.tgz &&\
```
ransomw1c: in order to avoid cyclical dependencies of a sort, build datamon using alpine and then copy it into the postgres image as in `sidecar.Dockerfile` or use the existing `gcr.io/onec-co/datamon-fuse-sidecar` image as a place to copy the binary from.  write a makefile target to tie the entire build together.

one way or another, in-repo dockerfiles needn't cyclically depend on github releases that are themselves created out of the repo.

pgsidecar/docker/initDB.sh
```diff
@@ -0,0 +1,20 @@
...
+echo Initializing DB
+
+echo Loading PostGIS extensions
+psql -U postgres -d postgres -c "CREATE EXTENSION IF NOT EXISTS postgis"
+
+if [ -z "$DATA_INPUTREPO" ]
+then
+  echo DATA_INPUTREPO environment variable must be set
+  exit 1
+else
```
ransomw1c: minor nitpick:  it's not necessary to have an `else` after an `exit`

ransomw1c:
overall, this appears to be a reasonable start.  please also start putting together some operating instructions in the form of comments here or shell-scripts and makefile targets:  notice how the `fuse-demo-ro` makefile target constitutes a full demo of the read-only sidecar, for instance.  it's understood that this is work-in-progress and usability is something it's evolving towards.

other things to start considering include
* multiple inputs and outputs:  a data-science script might need to access multiple databases and/or store the contents of multiple databases
* selecting specific tables for backup

finally, i'll link #256 , the corresponding issue, here.

kerneltime:
Closing this review for now. Will be revisiting this as part of the side car work.

PR #259, opened 2019-09-19
feat(bundle): added parallelism to bundle list

fredbi:


pkg/core/bundle_list_test.go
```diff
@@ -0,0 +1,289 @@
+package core
```
kerneltime: Why the different formatting?

ransomw1c:
the meaning of attaching the `concurrentBundleList` to the `Bundle` struct is hazy... as is, the options attached to a bundle are specific to that particular bundle -- i.e. these are options about per-bundle operations such as upload or download.

if i were implementing this, i probably would've added an additional positional parameter to `ListBundles`, although, upon review, the varadic `opts ...` form does appear to be a nice design decision.  my qualm is using `BundleOption` rather than a different type, something like
```
struct BundleListSettings{
   concurrentBundleList int
}

type BundleListOption func(s *BundleListSettings)
```
or even
```
type BundleListOption func(ctx context.Context) context.Context
```
for use in a loop
```
for _, bOpt := range opts {
    ctx = bOpt(ctx)
}
```

this last has to do with a design decision that's been percolating for some time but hasn't been reconciled and probably is the source of the confusion resulting in this comment:  i'm not too pleased with all of the mutable state that's currently attached to the `Bundle` struct as-is and want to replace it with an immutable map of settings, a context.

introducing any conventions about how to plumb `cmd/` options to the internal `pkg/` machinery would be welcome.  conversely, don't feel obliged to conform to existing patterns of how to attach `cmd/` to `pkg/` for the purposes of this patch.  if the
```
type SomeOpt func(s *SomeSettings)
```
pattern is actually something that jibes with your intuition, this is great:  create new option (function) and settings (struct) types specific to the bundle list.  if there is some entirely different pattern for attaching `cmd/` to `pkg/` don't hesitate to throw that out there, too.

the only definite qualm here is attaching an option to the `Bundle` struct that is not specific to operations a particular bundle.

pkg/core/bundle_list_test.go
kerneltime: Eh.. generated..

pkg/storage/mockstorage/store.go
```diff
@@ -0,0 +1,574 @@
+// Code generated by moq; DO NOT EDIT.
```
ransomw1c: i typically aver from checking generated code into version control, although i acknowledge that there's not a strict, dogmatic rule about this.  the thing is, i'm fairly scatterbrained:  if it's possible to forget to regenerate, i will eventually forget to regenerate.  so particularly, if CI could be responsible for keeping mocks in sync with the single source of truth, it would be much more convenient for me.

could we consider checking in `mockstorage/.gitkeep` only and changing `.circleci/config.yaml` to produce `mockstorage/store.go` on the fly?

meanwhile, for the use-case of running tests locally, we have a `Makefile` with a `test` target.  there could be a `generate` prereq for that target to generate mocks according to a `go:generate` comment like in the [suggested `moq` usage](https://github.com/matryer/moq#usage)?

yet, checking in the generated source could also be a considered and deliberate decision with some advantages that i'm not noticing and we can sync up via slack, zoom, or otherwise to get into the nitty-gritty of the tradeoffs involved.

ransomw1c:
attempting to run tests in the core package is producing

```
core% go test
--- FAIL: TestListBundles (0.00s)
    --- FAIL: TestListBundles/io_error (0.00s)
        bundle_list_test.go:280:
                Error Trace:    bundle_list_test.go:280
                Error:          "expected label basename wrong to match regexp ^(.*)\.json$" does not contain "io error"
                Test:           TestListBundles/io_error
```

on my local

kerneltime:
Did not dive too deep but 
```time ./out/datamon.mac bundle list --concurrency-factor 400 --repo backup-filestore-output```
takes longer (did not complete)
than 
```time ./out/datamon.mac bundle list --concurrency-factor 100 --repo backup-filestore-output
./out/datamon.mac bundle list --concurrency-factor 100 --repo   6.13s user 2.12s system 17% cpu 46.779 total
```

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,128 @@ import (
 )
 
 const (
-	maxBundlesToList = 1000000
+	maxBundlesToList  = 1000000
```
ransomw1c: naming conventions to mention, i'll start with the naming convention i added first:  this should've been `maxMetaFilesToProcess` or something similar.  it is _not_ the maximum number of bundles to list, and this name was based on my misunderstanding.  please choose a better name.

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,128 @@ import (
...
+	typicalBundlesNum = 1000
 )
 
+type bundleChan struct {
```
ransomw1c: i'd typically been using two channels, one for error and one for results rather than multiplexing both errors and results through one channel.  the pattern of using one channel for both errors and results is okay with me, except could we call these structs `*ChanVal` or something?  i would expect the `*Chan` to be the channel itself, not the value it contains.

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,128 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
-	bds := make([]model.BundleDescriptor, 0)
-	for _, k := range ks {
+
+	var wg sync.WaitGroup
+
+	result := make(chan bundlesChan)
+	bundle := make(chan bundleChan, concurrencyFactor)
```
ransomw1c: in light of the comment on the channel value naming, could we call this `bundleC`, `bundleChan` or something?

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,128 @@ import (
...
+
+	wg.Add(1)
+	go func(bundle <-chan bundleChan, result chan<- bundlesChan, wg *sync.WaitGroup) {
+		// coalesce result
```
ransomw1c: why not coalesce the result synchronously at the end of `ListBundles`?  the `bundle` channel closes when all `getBundleAsync` goroutines finish, and i get that it's necessary for `ListBundles` to not leak goroutines...

yet what's the harm in setting a `var errRes error` (or probably some better name) in the scope of `ListBundles` and possibly setting that to non-`nil` in a synchronous `for .. range bundles` at the end of the function?  then `return bds, errRes`, why not?  i don't suppose there's any harm in returning a non-empty slice of bundle descriptors in an error case, although a conditional around the `return` could resolve even that if it's a worry. 

it is at most marginally clearer imo to coalesce results in a separate goroutine, and additionally, it [costs a fair amount of time](https://syslog.ravelin.com/so-just-how-fast-are-channels-anyway-4c156a407e45) to push bytes through a channel.  in the case of the `bundle` channel, the time to push bytes is worth it because of the parallelism of reads from gcs and deserialization.  here, my opinion (at least at the moment) is that any code style improvement this goroutine might offer is not worth the computational time to push bytes through the `result` channel.

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,128 @@ import (
...
+	go func(bundle chan<- bundleChan, wg *sync.WaitGroup) {
+		// start a bounded number of workers fetching descriptors.
+		// Limitation: this currently spans the whole set of keys, even though an early error has been met
+		defer wg.Done()
+		var workers sync.WaitGroup
+		keys := make(chan string, concurrencyFactor)
+
+		for count, k := range ks {
+			if count < concurrencyFactor {
+				workers.Add(1)
```
ransomw1c: how about spinning up the `getBundleAsync` worker goroutines synchronously before starting to send tasks (`keys` channel values) out via this goroutine?  supposing a `MinInt(int, int) int` util, this could look like
```
keys := make(chan string)
var workers sync.WaitGroup
numWorkers := MinInt(concurrencyFactor, len(ks))
for i := 0; i < numWorkers; i++ {
  workers.Add(1)
  go getBundleAsync(repo, store, keys, bundle, &workers)
}
```
then, with the additional change to collect results synchronously, the buffers can be removed from both the `keys` and `bundle` channels.  the current opinion is that channels should be unbuffered if at all possible for now in order to favor correctness.  we can add buffering for performance later on as necessary.

ransomw1c:
in `cmd/datamon/cmd`, `go test -v -run 'TestListBundles$'` sometimes reports

```
=== RUN   TestListBundles
Using config file: /Users/ransomwilliams/.datamon/datamon.yaml
Using config file: /Users/ransomwilliams/.datamon/datamon.yaml
Using config file: /Users/ransomwilliams/.datamon/datamon.yaml
Uploading blob:0e66a0b9a2f41af5c5dffc8bf3ea4a14f596ccb537c1a456a74754933996401390abf9714228ce3cc45dd9748741eb68a42be83ad84f579c6f9e53f92238bd82, bytes:1024
{"level":"info","ts":1568932698.646773,"caller":"core/bundle_pack.go:302","msg":"uploaded filelists","actual number uploads attempted":1,"approx expected number of uploads":0}
{"level":"info","ts":1568932699.158282,"caller":"core/bundle_pack.go:310","msg":"Uploaded bundle id","BundleID":"1R4iJOS70nL0ButWW1bIs1HY9v0"}
Uploaded bundle id:1R4iJOS70nL0ButWW1bIs1HY9v0
Using config file: /Users/ransomwilliams/.datamon/datamon.yaml
Uploading blob:1af0f9a940deb1fa0194a0d8982801ba436c4ef341a08f0474691e2348594722a96cb71d057cd9f3afe875570d90bf0c82436174f21069e464bdcd19395ee279, bytes:1
Uploading blob:d44d00d09629ecfa8a5a00fe0d3dbc77af438b8148d52176bc1fc6e3893bda7e955d19bef08d9f2d10e8baccf41b3b84e600a2bc9abdd63791fc0ae76ae19cd7
Uploading blob:99ebddccddbcc29bfd69feaa887e3b335d76d6a9777bae1e9dd1ce4003d9b7083352bb838616620c5fd6b4deab693fbbd4fee091aa514ae0191abc850ae2dcdf, bytes:1
Uploading blob:0e9f8bb8fa6187347bb7dac1731ba49ff35db6d948603b7669841780c81cd401e3da972b9c0397046cc6f304365d93726ced256d399051d59d64634fe11d60e4
Uploading blob:5d0e795a65273beb7cc5bdc9676b46610a5bf720177ca0c8003b626277a728e700907291ee9b45a33fb13c0b860f7e3e8b7b53e3d9a5fdbd3d9501496eeb257d, bytes:2097151
Uploading blob:82e69f6d942e26c99f707f5822649eef47fc0b8512a14237b1ad00b73992a66176a57d321022d5e833f6a478842078d9f86071c90a82bb9c567f74fe427921c1
Uploading blob:b118444420008dc59d14dc1de6cd5e9af7979c183449f88c9bba4c59430953ff08c01c43709e13353276a5d02261a83aa8537670e34376b705ae15be9fce5356
Uploading blob:6dae1a84aeae4c798027571e788e28f57bed7e4e878bdda2f836faf5da6feec62d46ac160dc5a1afde96478cbbe2029bdbd4680bc04b6e7883815ac625a152f8
{"level":"info","ts":1568932720.6018949,"caller":"core/bundle_pack.go:302","msg":"uploaded filelists","actual number uploads attempted":1,"approx expected number of uploads":0}
{"level":"info","ts":1568932721.123538,"caller":"core/bundle_pack.go:310","msg":"Uploaded bundle id","BundleID":"1R4iK3HZVE9THuYGKHPgheswLMG"}
Uploaded bundle id:1R4iK3HZVE9THuYGKHPgheswLMG
Using config file: /Users/ransomwilliams/.datamon/datamon.yaml
Uploading blob:8339171ab8acd2c45e53df901d5693370c0e0758f37f239960f7f69254ae53cf48bf045aed4f20889082f195360abd6e3d5b5281356d706a91c2110a8208bfe3, bytes:200
Uploading blob:979390a49f1435365d2ca03dd76cf786e5c56d72d7c8be251c394efed8e151dde1147002bcb1a337ef9326d247647533dabe4c478ed3d0b0611f31a47c59bf6b, bytes:100
{"level":"info","ts":1568932725.920009,"caller":"core/bundle_pack.go:302","msg":"uploaded filelists","actual number uploads attempted":1,"approx expected number of uploads":0}
{"level":"info","ts":1568932726.8141031,"caller":"core/bundle_pack.go:310","msg":"Uploaded bundle id","BundleID":"1R4iN1U9GxwNSealczC748RckfG"}
Uploaded bundle id:1R4iN1U9GxwNSealczC748RckfG
--- FAIL: TestListBundles (46.20s)
    require.go:157:
                Error Trace:    cli_test.go:505
                                                        cli_test.go:533
                Error:          Not equal:
                                expected: "vgfvj8eyi45r4tg"
                                actual  : "7pivdi6inw6oz6k"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -vgfvj8eyi45r4tg
                                +7pivdi6inw6oz6k
                Test:           TestListBundles
                Messages:       bundle log message
FAIL
exit status 1
FAIL    github.com/oneconcern/datamon/cmd/datamon/cmd   47.247s
```
on my local.

this is because the bundle results are expected (by the tests) to be in the same order as the partially sorted (ksuid) bundle ids returned from `KeysPrefix` such that the most recently uploaded bundle is the last item in the list.  as far as i recall, sorting the list of bundles is not a requirement, so changing `cli_test.go` is not strictly required.

either the cli tests need to be changed to reflect the fact that bundles are no longer listed in any particular order or the implementation needs to be changed to return bundle descriptors in the same order as keys they correspond to.

this is an unforeseen difficulty.  apologies for not mentioning it in planning.

pkg/storage/mockstorage/store.go
fredbi: Yeah just wanted to avoid hacking through existing code with go generate stuff depending on your local env

fredbi: Yeah just wanted to avoid hacking through existing code with go generate stuff depending on your local env

pkg/core/bundle_list_test.go
fredbi: @kerneltime could please you elaborate a bit? I don't understand

pkg/core/bundle_list.go
fredbi: ok will do

pkg/core/bundle_list.go
fredbi: Extra cost of pushing through channels is a valid point I did not consider.
Initial design (2 days ago) was pushing keys and coalescing results synchronously but then I met a blocking situation. Basically,  your advice of launching the workers _before_ is probably the right solution (only justification was to avoid looping twice).

pkg/core/bundle_list.go
fredbi: Yes I am considering that.

fredbi:
@ransomw1c  Thank you Ransom for all this much interesting insight.

I admit that I've been goaded away from my initial cleaner design because of side aspects of the channel workings like blocking sends (hence some undue extra async levels) and randomized select (having a separate error channel could help there).

Totally agree with the Option design. I admit that I just lazily patched the existing one (I do realize that reusing `BundleOption`  is not really the intended way to use that) better than squarely declaring a new type for it. Will definitely do that.

pkg/core/bundle_list.go
fredbi: dual channel is an interesting option. Giving it a try

pkg/storage/mockstorage/store.go
ransomw1c: > if `// go:generate` is better than `hack/generate-xxx.sh`

no strong opinion here.  i was only repeating the documentation, and this is not a strong opinion compared to experience with actual usage.

pkg/core/bundle_list_test.go
kerneltime: tab space here seems at 8? vs 3

fredbi: @kerneltime there are no spaces to indent in gofmt golang source, only tabs.

pkg/core/bundle_list.go
fredbi: ok. So:
- [x] Going unbuffered
- [x] Spinning workers beforehand
- [x] Removing extra data copy to channel
- [ ] "synchronous" push/fetch: well, I am sticking to "watcher" single goroutines to prevent blocking sends

pkg/core/bundle_list.go
fredbi: * [x] removed channels of struct (`bundleChan chan<-model.Descriptor`, èrrorChan chan<-error`, `keysChan <- chan string`)
* [x] used 2 channels instead of one as struct

pkg/storage/mockstorage/store.go
fredbi: * [x] removed mock from git
* [x] chose go generate / removed hack script
* [x] added target in Makefile
* [x] included that in CI

pkg/core/bundle_list_test.go
kerneltime: @fredbi i opened up the file in IDE and the formatting looks fine but here on GitHub the tab width seems to be set at 8..

fredbi: Ok

ransomw1c:
there are yet more `cli_test.go` tests that can fail as a result of the change in assumption that the last bundle listed is the most recent bundle created

```
--- FAIL: TestDownloadBundles (17.27s)
    require.go:347:
                Error Trace:    cli_test.go:728
                                                        cli_test.go:759
                Error:          Should be false
                Test:           TestDownloadBundles
                Messages:       no filesystem entry at destination path 'bundle-dl-1RFPs5zaMmxtkEpzAJ1iWMf7LKD' before bundle upload
```

```
--- FAIL: TestListBundlesFiles (16.45s)
    require.go:157:
                Error Trace:    cli_test.go:1155
                                                        cli_test.go:1187
                Error:          Not equal:
                                expected: 2
                                actual  : 6
                Test:           TestListBundlesFiles
                Messages:       file count in bundle files list log
```

to fix these (and perhaps others) at the level of `cli_test.go` rather than changing `ListBundles` to include a sort, search for all uses of `listBundles` (for example, `ll, _ := listBundles(...)`) where the last element in the result is accessed (that is, `ll[len(ll)-1]`) and make a change to the test that drops the assumption that this (`ll[len(ll)-1]`) corresponds to the most recently created bundle.

one idea is to call `llBefore, _ := listBundles(...)` and `ll, _ := listBundles(...)` on either side of the command that typically preceeds `listBundles` in test.  for example,

```
	rllBefore, err := listBundles(t, repo1)
	runCmd(t, []string{"bundle",
		"upload",
		"--path", dirPathStr(t, files[0]),
		"--message", msg,
		"--repo", repo1,
		"--concurrency-factor", concurrencyFactor,
	}, "create second test repo", false)
	rll, err := listBundles(t, repo1)
```

in `testListBundleFiles` and then have a helper like

```
func getAdditionalBundleListEntry(t *testing.T, bles []bundleListEntry, blesBefore []bundleListEntry) bundleListEntry {
	blesM := make(map[string]bundleListEntry)
	for _, ble := range blesBefore {
		require.IsNil(t, blesM[ble.hash])
		blesM[ble.hash] = ble
	}
	var bleNew bundleListEntry
	for _, ble := range bles {
		if blesM[ble.hash] == nil {
			require.Equal(t, "", bleNew.hash)
			bleNew = ble
		}
	}
	require.NotEqual(t, "", bleNew.hash)
	return bleNew
}
```

and use the result of this helper instead of `ll[len(ll)-1]` everywhere.

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,139 @@ import (
...
-	for _, k := range ks {
+
+	var (
+		workers, wg sync.WaitGroup
+		werr        error
+	)
+
+	bundleChan := make(chan model.BundleDescriptor)
+	keyChan := make(chan string)
+	errorChan := make(chan error)
```
ransomw1c: i kinda liked the pattern where both errors and results got sent through the same channel via a struct .. having a separate error channel is okay, too.  was there a reason that the original design was changed from multiplexing both errors and okay results through the same channel to using two separate channels?

fredbi: I kinda liked your suggestion, as I had always made it multiplexed before... :)

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,139 @@ import (
...
+
+	wg.Add(1)
+	go func(errorChan <-chan error, wg *sync.WaitGroup) { // watch for errors
+		defer wg.Done()
+		for err := range errorChan {
+			werr = err
+		}
+	}(errorChan, &wg)
+
+	for _, k := range ks { // distribute work
```
ransomw1c: was there advantage/disadvantage type reasoning involved in deciding to distribute work synchronously and coalesce results asynchronously?

like, this distribute work loop could probably be included in the goroutine that's waiting for workers to complete?

pkg/core/bundle_list.go
```diff
@@ -12,52 +13,139 @@ import (
...
+		workers.Add(1)
+		go getBundleAsync(repo, store, keyChan, bundleChan, errorChan, &workers)
+	}
+
+	bds := make([]model.BundleDescriptor, 0, typicalBundlesNum) // preallocate some typical number of bundles, e.g. 1000
+
+	wg.Add(1)
+	go func(bundleChan <-chan model.BundleDescriptor, wg *sync.WaitGroup) {
+		defer wg.Done()
+		for bd := range bundleChan { // watch for results and coalesce
```
ransomw1c: regarding the use of `select` mentioned on Slack, yes, there could be a

```
for {
	select {
		case bd, bundleChanMore := <- bundleChan:
			bds = append(bds, bd)
		case err := <- errChan:
			werr = err
	}
	if !bundleChanMore { 
		for {
			err, errChanMore := <- errChan
			if err != nil { werr = err }
			if !errChanMore { break }
		}
		break
	}
}
```

or some similar at the end of `ListBundles`.  with this, there's no need for the `WaitGroup` declared as `wg`.  closing the `bundleChan` and `errorChan` (in that order) is detected via the [receive operator](https://golang.org/ref/spec#Receive_operator).

kerneltime:
@fredbi take a look at https://github.com/oneconcern/datamon/pull/261 it should help reduce processing unwanted/duplicate keys.

pkg/core/bundle_list.go
fredbi: I am avoiding blocking situations in which all workers would be waiting on sending their result and work distributor would be blocked on pushing keys. 

So one of these tasks has to be run in the background. Doing it the other way around is I believe equivalent.

I’ve made some attempts at doing it all in one single loop. This requires the use of select and much extra care with avoiding dropped messages. In the end, this version came up to be much more complex and difficult to read so I decided to drop it and adopt straightforward looping, regardless of possible wasted work whenever a failure occurs.

ransomw1c: > So one of these tasks has to be run in the background. Doing it the other way around is I believe equivalent.

as the implementation currently exists, i agree: scheduling tasks or coalescing results in the background is equivalent.

> doing it all in one single loop ... this version came up to be much more complex

_totally_ agree here.

----

i kinda suspect that the (for now) equivalent coalescing results synchronously and scheduling tasks in the background will prove to be more durable over time, particularly upon parallelizing the read of keys out of gcs.

yet for the purposes of this patch, it's not worth finagling over in case the review cycle has already become somewhat of a drag.

ransomw1c:
```
datamon% go generate ./...
Couldn't load mock package: go [list -e -json -compiled=true -test=false -export=false -deps=false -find=true --]: exit status 1: build .: cannot find module for path .

moq [flags] destination interface [interface2 [interface3 [...]]]
  -out string
        output file (default stdout)
  -pkg string
        package name (default will infer)
pkg/storage/store.go:10: running "moq": exit status 1
```
on `make mocks` via `test` target

fredbi:
@ransomw1c just fixed the Makefile target. 
I am following up the fix the hack/script if working around at the moq repo

PR #260, opened 2019-09-23
base docker img, wrap_datamon.sh to zsh, dbg dest store parsing

ransomw1c:
this patch clears the ground a bit for #256 and other datamon kubernetes "batteries":

* all docker images that require building binaries from golang source are based on `binaries.Dockerfile`, and this is the only docker image that includes `go build` steps
* `wrap_datamon.sh` is converted to zsh.  `wrap_application.sh` is still POSIX sh so that it can be run within any application container from a shared volume.  the battery-specific scripts, on the other hand, run within their respective images and can use any interpreter made available within those docker images
* 3dbeae5d introduced a bug wherein it wasn't possible to specify `--destination` on the `bundle mount` command.  this commit both fixes that bug as well as changes the demo to omit the `--destination` flag such that a temporary directory is used, probably a saner default to recommend to users.

### DIFF
--- a/hack/fuse-demo/create_ro_pod.sh
+++ b/hack/fuse-demo/create_ro_pod.sh
@@ -68,9 +68,8 @@ else
         echo 'bundle id and label are mutually exclusive' 1>&2
         exit 1
     fi
-    bundle_id=$("$DATAMON_EXEC" label list --repo "$repo_name" 2>&1 | \
-                    grep '^'"$label" | \
-                    cut -d',' -f2 | sed 's/ //g')
+    bundle_id=$(2>&1 "$DATAMON_EXEC" label get --repo "$repo_name" --label "$label" | \
+                    cut -d ',' -f 2 | tr -d ' ')
     if [ "$bundle_id" = '' ]; then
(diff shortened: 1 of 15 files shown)


hack/fuse-demo/create_ro_pod.sh
```diff
@@ -68,9 +68,8 @@ else
         echo 'bundle id and label are mutually exclusive' 1>&2
         exit 1
     fi
-    bundle_id=$("$DATAMON_EXEC" label list --repo "$repo_name" 2>&1 | \
-                    grep '^'"$label" | \
-                    cut -d',' -f2 | sed 's/ //g')
+    bundle_id=$(2>&1 "$DATAMON_EXEC" label get --repo "$repo_name" --label "$label" | \
+                    grep -v '^Using config file' | cut -d ',' -f 2 | tr -d ' ')
```
kerneltime: We can just stop printing this in the CLI.. might be a cleaner approach

ransomw1c:
additional changes include:

* better error handling (error messages printed to logs when datamon commands fail)
* removed the 'Using config file' log message

PR #261, opened 2019-09-23
Use delimiter to reduce the number of keys for bundle list

kerneltime:


pkg/core/bundle_list.go
```diff
@@ -22,7 +23,7 @@ func ListBundles(repo string, store storage.Store) ([]model.BundleDescriptor, er
 	if e != nil {
 		return nil, e
 	}
-	ks, _, err := store.KeysPrefix(context.Background(), "", model.GetArchivePathPrefixToBundles(repo), "", maxBundlesToList)
+	ks, _, err := store.KeysPrefix(context.Background(), "", model.GetArchivePathPrefixToBundles(repo), "/", maxBundlesToList)
```
ransomw1c: haven't pulled apart https://godoc.org/cloud.google.com/go/storage details in full.  instead, verified with the debugger that after this change, elements of `ks` are such as

```
"bundles/ransom-test-repo-20190408/1Jbb3SicFGoKB7JQJZdCCwdBQwE/"
```

so that archive path components, `apc` below, are

```
model.ArchivePathComponents {
        Repo: "ransom-test-repo-20190408",
        BundleID: "1Jbb3SicFGoKB7JQJZdCCwdBQwE",
        ArchiveFileName: "",
        LabelName: "",}
```

rather than

```
        "bundles/ransom-test-repo-20190408/1Jbb3SicFGoKB7JQJZdCCwdBQwE/bundle-files-0.json",
        "bundles/ransom-test-repo-20190408/1Jbb3SicFGoKB7JQJZdCCwdBQwE/bundle.json",
```

kerneltime: Yes, the GCS underlying API returns the prefix if the path is not an object but just the delimited key.

PR #263, opened 2019-09-25
init postgres sidecar

ransomw1c:
#256


ransomw1c:
as of bc3ff8c ,

```
make pg-demo-coord-build-app
make build-and-push-pg-sidecar
./hack/fuse-demo/create_coord_pg_pod.sh
```

followed by
```
./hack/fuse-demo/run_coord_pg_shell.sh -s
```
and
```
./hack/fuse-demo/run_coord_pg_shell.sh
```
to enter shells in the _sidecar_ and _application_ containers, respectively,

and then, first, in the application container
```
/tmp/coord/.scripts/wrap_application.sh -s -c /tmp/coord -b postgres -- ./mock_application_pg.sh 5430
```
followed by
```
./wrap_datamon_pg.sh -c /tmp/coord -x -r ransom-datamon-test-repo -l pg-coord-example -m 'postgres coordination example' -p 5430
```
in the sidecar

is a minimal demo of creating a database and coordinating an upload to datamon.

hack/fuse-demo/coord-app-pg.Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM debian
+
+RUN apt-get update &&\
+    curl -sL https://deb.nodesource.com/setup_10.x  | bash &&\
+    apt-get install -y \
+        curl \
+        postgresql \
```
kerneltime: We need to lock down the version of postgres and the build itself. We can store the deb file within datamon.

hack/fuse-demo/coord-app-pg.Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM debian
```
kerneltime: We should use a release so that the image does not change. This is important for the PG version management.

hack/fuse-demo/coord-app-pg.Dockerfile
kerneltime: https://www.postgresql.org/docs/current/upgrading.html

kerneltime:
The approach looks fine. Will try to give it a try tomorrow.

ransomw1c:
POC demo is

```
./hack/fuse-demo/demo_pg_coord.sh
```

* `-s` skip container builds (after the first use)
* `-b` build datamon source

ransomw1c:
##### note to self

to resolve in parallel: `i256-tinicache--wip` workaround for rate-limiting error from the host (not a distro pkg, perhaps use a distro pkg?) where the pid 0 bin resides.

sidecar-pg.Dockerfile
```diff
@@ -0,0 +1,75 @@
+FROM datamon-binaries as base
+
+# dist-alike during development/debug
+RUN cp /stage/usr/bin/datamon /usr/bin/datamon
+ADD ./hack/fuse-demo/datamon.yaml /root/.datamon/datamon.yaml
+
+# Build the dist image
+# ??? do not need fuse and sudo in case of bundle download instead of mount
```
ransomw1c: main question off top:  use `bundle mount` or `download` for existing bundles?

fredbi:
See my comment on #256 and https://www.postgresql.org/docs/current/sql-checkpoint.html

.circleci/config.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - run:
           name: Run golang linter
           command: |
-            golangci-lint run --disable golint,funlen,godox,whitespace,stylecheck --build-tags fuse_cli --max-same-issues 0 --verbose
+            golangci-lint run --disable golint,funlen,godox,whitespace,stylecheck,unparam --build-tags fuse_cli --max-same-issues 0 --verbose
```
fredbi: Why don't we summarize this under `.golangci.yml` and allow contributors to share the same linting rules as CI?

README.md
```diff
@@ -119,28 +119,71 @@ multiple labels can refer to the same bundle via its commit hash.
 
 Current use of Datamon at One Concern with respect to intra-Argo workflow usage relies on the
```
fredbi: Just a question here: I do understand that our use case is with ARGO, but I believe datamon is quite independent from ARGO, isn't it? (no change requested for that)

README.md
```diff
@@ -119,28 +119,71 @@ multiple labels can refer to the same bundle via its commit hash.
...
+Docker images that practicably constitute sidecars.
+While there's precisely one application container per Argo node,
+a Kubernetes container created from an arbitrary image,
+sidecars are additional containers in the same Kubernetes pod
+-- or Argo DAG node, we can say, approximately synonymously --
+that concert datamon-based data-ferrying setups with the application container.
+
+_Aside_: as additional kinds of data sources and sinks are added,
+we may also refer to "sidecars" as "batteries," and so on as semantic drift
+of the shell scripts shears away feature creep in the application binary.
```
fredbi: _aside_ : the semantic drift is such that I got eventually lost in the middle of this sentence. Probably because I've no idea what a battery could be in this context.

README.md
```diff
@@ -119,28 +119,71 @@ multiple labels can refer to the same bundle via its commit hash.
...
+spec:
+  ...
+  containers:
+    - name: datamon-sidecar
+    - image: gcr.io/onec-co/datamon-fuse-sidecar:v0.7
+  ...
+```
+would be the corresponding Kubernetes YAML to access the sidecar container image.
+
+_Aside_: historically, and in case it's necessary to roll back to an now-antient
```
fredbi: ancient?

README.md
fredbi: Reading ahead a bit, I understand that what you are basically saying here is that the postgres-specific wrapper written in shell will gradually evolve toward native golang. Am I correct?

sidecar-pg.Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+FROM postgres:11.5
```
fredbi: I am not following clearly here why we build a postgres from scratch (and a lot of pg-related scripting) when we can probably be content with a standard postgres or postgis image, possibly installing some extra postgres addons. Does this come from a constraint on the version?
Using a standard docker image, the use-case corresponding to "mounting from a dump" becomes trivial, since the image startup script automatically provision any dump(s) given in its init.d.

sidecar-pg.Dockerfile
fredbi: I might have missed some nuances in datamon terminology, but "mount" would better correspond to "make the database available", while "download" would correspond to "make the database _storage_ available". Just saying.

README.md
```diff
@@ -119,28 +119,71 @@ multiple labels can refer to the same bundle via its commit hash.
...
-both the main application Dockerfile as well as the YAML.  When using a block-storage GCS
-product, we might've specified a data-science application's Argo DAG node with something
-like
+of each of the sidecar containers within the main container.
+This
+[can be accomplished](https://github.com/oneconcern/datamon/blob/master/hack/k8s/example-coord.template.yaml#L15-L24)
+via an `initContainer` without duplicating version of the Datamon sidecar
+image in both the main application Dockerfile as well as the YAML.
+When using a block-storage GCS product, we might've specified a data-science application's
+Argo DAG node with something like
```
fredbi: This part is most interesting. I believe we could use an example here (or with a link), demonstrating the typical workflow of a user.

ransomw1c:
consider Postgres `CHECKPOINT` along with vacuuming.

https://www.postgresql.org/docs/current/sql-checkpoint.html

https://github.com/oneconcern/datamon/issues/256#issuecomment-537918810

sidecar-pg.Dockerfile
fredbi: Oops. Was mislead here. Please ignore my comment

sidecar-pg.Dockerfile
kerneltime: Mount is specifically used when the bundle is mounted as a filesystem (no data downloaded locally) vs download is where all the data is downloaded to the local node.

.circleci/config.yml
kerneltime: Makes sense.

README.md
ransomw1c: (direct-message channel)

there's a small is-ought gap here:  datamon should be independent of argo, but practically, most of our users are folks writing argo pipelines, and we want to provide conveniences.
https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain
is meaningful here.
if datamon ever gets to be a popular thing, i'd like to have a [clearly delineated] porcelain layer, possibly in scheme or s'th.

README.md
ransomw1c: see above re https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain

README.md
ransomw1c: :/ for further consideration at office today..

README.md
ransomw1c: recursively

.circleci/config.yml
ransomw1c: and done

hack/k8s/example-coord-pg.template.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - mountPath: /tmp/coord
+          name: container-coord
+      - name: datamon-sidecar
+        image: gcr.io/onec-co/datamon-pg-sidecar:latest
+        imagePullPolicy: "$PULL_POLICY"
+
+        # command: ["/bin/tini"]
+        # args: ["--", "/bin/zsh"]
+
+        command: ["./wrap_datamon_pg.sh"]
```
kerneltime: This should be made declarative and hide the implementation details within the container.

ransomw1c:
```
brew tap petere/postgresql
brew tap-pin petere/postgresql
brew install petere/postgresql/postgresql@12
```

hack/k8s/example-coord-pg.template.yaml
ransomw1c: 4c1e9bf includes a serialization format that can be specified in the YAML in a declarative fashion (environment variables).

ransomw1c:
status on landing is that internals have a reasonable amount of stress testing, while user-interface is to be significantly revisited.

PR #266, opened 2019-10-04
feat(bundle): paginated bundle list

fredbi:
# Paginated bundle list

### Feature
This PR adds pagination support to the `bundle list` command. This means that we may query
large sets of keys while streaming back the output to the user.

### Requirements
Returned bundle metadata are sorted by key ID (ksuid).

### Description
I've built on top of PR #259 (parallel bundle metadata retrieval) and now decouple the retrieval of keys (which is sequential) from the retrieval of metadata (which is parallel).
Both queries progress asynchronously.
Each page (batch) of keys is submitted for parallel retrieval of bundle metadata, which are sorted before returning this batch.

### Main changes
The size of a page (batch size), expressed in # of keys, is an optional CLI  parameter.

The existing `pkg/core/ListBundle`  func keeps its interface (it is used by `pkg/web`).
I've added  the `pkg/core/ListBundleApply` func (now used by the CLI) to run any function asynchronously (that is in a non-blocking way) while fetching bundles.

Since this applied function may return an error, a signaling mechanism is put in place to abort the goroutines.

Some additional helper functions are now exported. The `ListBundleApply` func is constructed out of those.

I refactored error handling from storage (introducing sentinel errors in separate package) to avoid the string based error check in bundle list (en passant, made the adaptation for afero fs implem too).
### Testing
Besides CLI testing, the new feature is rather well covered by mocked tests, checking for no leaking goroutines, checking for most error paths.

### Suboptimal aspects
I am aware of the following suboptimal implementation issues:
* default values (batch size: 1024 or 20 kB, parallelism: 100) may not be the best ones. If users or benchmarkers find better defaults, I'll change them.
* the worker pool is spinned then wound down for every page. Although probably negligible for a few pages, this could add-up to some significant time with very large number of pages. However, this will be difficult to improve, given the sorting requirement.
* the error handling in the goroutines code is complex and required a lot of testing for blocking corner cases. This can be greatly improved by multiplexing values and errors on the same channel. At the moment, I am hesitating between  `chan struct{value: value_t, err: error}` and `chan func() (value_t, error)`. Your advice is welcome on this choice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ mocks:
 
 .PHONY: runtests
 runtests: mocks
-	@go test ./...
+	@go test -v ./cmd/datamon//cmd
 
 .PHONY: gofmt
 ## Run gofmt on the cmd and pkg packages
--- a/cmd/datamon/cmd/bundle_list.go
+++ b/cmd/datamon/cmd/bundle_list.go
@@ -7,14 +7,16 @@ import (
 	"text/template"
 
 	"github.com/oneconcern/datamon/pkg/core"
+	"github.com/oneconcern/datamon/pkg/model"
 
 	"github.com/spf13/cobra"
 )
 
+// BundleListCommand describes the CLI command for listing bundles
 var BundleListCommand = &cobra.Command{
 	Use:   "list",
 	Short: "List bundles",
-	Long:  "List the bundles in a repo",
+	Long:  "List the bundles in a repo, ordered by their key",
 	Run: func(cmd *cobra.Command, args []string) {
 		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}}`
 		ctx := context.Background()
@@ -23,17 +25,18 @@ var BundleListCommand = &cobra.Command{
 		if err != nil {
 			logFatalln(err)
 		}
-		bundleDescriptors, err := core.ListBundles(params.repo.RepoName, remoteStores.meta, core.ConcurrentBundleList(params.core.ConcurrencyFactor))
-		if err != nil {
-			logFatalln(err)
-		}
-		for _, bd := range bundleDescriptors {
+		err = core.ListBundlesApply(params.repo.RepoName, remoteStores.meta, func(bundle model.BundleDescriptor) error {
 			var buf bytes.Buffer
-			err := listLineTemplate.Execute(&buf, bd)
-			if err != nil {
-				log.Println("executing template:", err)
+			e := listLineTemplate.Execute(&buf, bundle)
+			if e != nil {
+				log.Println("executing template:", e)
+				return e
 			}
 			log.Println(buf.String())
+			return nil
+		}, core.ConcurrentBundleList(params.core.ConcurrencyFactor), core.BundleBatchSize(params.core.BatchSize))
+		if err != nil {
+			logFatalln(err)
 		}
 	},
 }
@@ -45,6 +48,7 @@ func init() {
 	addBucketNameFlag(BundleListCommand)
 	addBlobBucket(BundleListCommand)
 	addCoreConcurrencyFactorFlag(BundleListCommand)
+	addBatchSizeFlag(BundleListCommand)
 
 	for _, flag := range requiredFlags {
 		err := BundleListCommand.MarkFlagRequired(flag)
--- a/cmd/datamon/cmd/cli_test.go
+++ b/cmd/datamon/cmd/cli_test.go
@@ -463,7 +463,7 @@ func (b bundleListEntries) Len() int {
 	return len(b)
 }
 func (b bundleListEntries) Less(i, j int) bool {
-	return b[i].time.Before(b[j].time)
+	return b[i].hash < b[j].hash
 }
 func (b bundleListEntries) Last() bundleListEntry {
 	return b[len(b)-1]
@@ -493,12 +493,15 @@ func listBundles(t *testing.T, repoName string) (bundleListEntries, error) {
 		}
 		rle := bundleListEntry{
 			rawLine: line,
-			hash:    strings.TrimSpace(sl[0]),
+			hash:    strings.TrimSpace(sl[0]), // bundle key ID
 			message: strings.TrimSpace(sl[2]),
 			time:    t,
 		}
 		bles = append(bles, rle)
 	}
+	// bundles are ordered by lexicographic order of keys (ksuids).
+	require.True(t, sort.IsSorted(bles))
+
 	sort.Sort(bles) // sort test result by timestamp
 	return bles, nil
 }
@@ -521,7 +524,6 @@ func testListBundle(t *testing.T, file uploadTree, bcnt int) {
 	require.NoError(t, err, "error out of listBundles() test helper")
 	require.Equal(t, bcnt, bundles.Len(), "bundle count in test repo")
 
-	// ordering of bundles is not strictly required
 	found := false
 	for _, b := range bundles {
 		if msg == b.message {
--- a/cmd/datamon/cmd/flags.go
+++ b/cmd/datamon/cmd/flags.go
@@ -52,6 +52,7 @@ type paramsT struct {
 	}
 	core struct {
 		ConcurrencyFactor int
+		BatchSize         int
 	}
 }
 
@@ -143,9 +144,17 @@ func addCoreConcurrencyFactorFlag(cmd *cobra.Command) string {
 	// this takes the usual "concurrency-factor" flag, but sets non-object specific settings
 	concurrencyFactor := concurrencyFactorFlag
 	cmd.Flags().IntVar(&params.core.ConcurrencyFactor, concurrencyFactor, 100,
-		"Heuristic on the amount of concurrency used by core operations (e.g. bundle list).  "+
+		"Heuristic on the amount of concurrency used by core operations (e.g. bundle list). "+
+			"Concurrent retrieval of bundle metadata is capped by the 'batch-size' parameter. "+
 			"Turn this value down to use less memory, increase for faster operations.")
 	return concurrencyFactor
+}
+
+func addBatchSizeFlag(cmd *cobra.Command) string {
+	batchSize := "batch-size"
+	cmd.Flags().IntVar(&params.core.BatchSize, batchSize, 1024,
+		"Number of bundle keys fetched per rountrip to storage -- keys are 20B long, defaults corresponds to 20kB pages")
+	return batchSize
 }
 
 func addWebPortFlag(cmd *cobra.Command) string {
--- a/pkg/core/bundle.go
+++ b/pkg/core/bundle.go
@@ -180,13 +180,19 @@ type BundleListOption func(*CoreSettings)
 // CoreSettings defines various settings for core features
 type CoreSettings struct {
 	concurrentBundleList int
+	bundleBatchSize      int
+	doneChannel          chan struct{}
 }
+
+const (
+	defaultBundleBatchSize = 1024
+)
 
 var (
 	defaultBundleListConcurrency = 2 * runtime.NumCPU()
 )
 
-// ConcurrentBundleList sets the max level of concurrency to retrieve bundles. It defaults to 4 x #cpus.
+// ConcurrentBundleList sets the max level of concurrency to retrieve bundles. It defaults to 2 x #cpus.
 func ConcurrentBundleList(concurrentBundleList int) BundleListOption {
 	return func(s *CoreSettings) {
 		if concurrentBundleList == 0 {
@@ -194,6 +200,31 @@ func ConcurrentBundleList(concurrentBundleList int) BundleListOption {
 			return
 		}
 		s.concurrentBundleList = concurrentBundleList
+	}
+}
+
+// BundleBatchSize sets the batch window to fetch bundle keys. It defaults to 100.
+func BundleBatchSize(batchSize int) BundleListOption {
+	return func(s *CoreSettings) {
+		if batchSize == 0 {
+			s.bundleBatchSize = defaultBundleBatchSize
+			return
+		}
+		s.bundleBatchSize = batchSize
+	}
+}
+
+// WithBundleDoneChan sets a signaling channel controlled by the caller to interrupt ongoing goroutines
+func WithBundleDoneChan(done chan struct{}) BundleListOption {
+	return func(s *CoreSettings) {
+		s.doneChannel = done
+	}
+}
+
+func defaultCoreSettings() CoreSettings {
+	return CoreSettings{
+		concurrentBundleList: defaultBundleListConcurrency,
+		bundleBatchSize:      defaultBundleBatchSize,
 	}
 }
 
--- a/pkg/core/bundle_list.go
+++ b/pkg/core/bundle_list.go
@@ -4,18 +4,21 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"strings"
+	"sort"
 	"sync"
+
+	"errors"
 
 	"gopkg.in/yaml.v2"
 
 	"github.com/oneconcern/datamon/pkg/model"
 	"github.com/oneconcern/datamon/pkg/storage"
+	"github.com/oneconcern/datamon/pkg/storage/status"
 )
 
 const (
 	maxMetaFilesToProcess = 1000000
-	typicalBundlesNum     = 1000
+	typicalBundlesNum     = 1000 // default number of allocated slots for bundles in a repo
 )
 
 func minInt(a, b int) int {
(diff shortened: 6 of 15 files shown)


ransomw1c:
> the retrieval of keys (which is sequential)

are we sure that what was intended by "pagination" wasn't in the sense of or according to the [page tokens](https://godoc.org/google.golang.org/api/iterator#PageInfo) such that that the list of pages of keys could be traversed sequentially while the keys themselves are being retrieved in parallel?

fredbi:
@ransomw1c
Trying to clarify here: the pages of keys are retrieved sequentially. Bundle metadata attached to each key in a page (batch) are retrieved in parallel.

fredbi:
See #267 and #268 for versions implementing channel multiplexing w/ errors.

Makefile
```diff
@@ -192,7 +192,7 @@ mocks:
 
 .PHONY: runtests
 runtests: mocks
-	@go test ./...
+	@go test -v ./cmd/datamon//cmd
```
ransomw1c: there are tests other than those in `cmd`

fredbi: Yeah. Blunder. I’ve also seen it today. I unduly committed a local change just for me. Will revert

fredbi: Yeah. Blunder. I’ve also seen it today. I unduly committed a local change just for me. Will revert

fredbi: There is also an undue change in pkg/store errors: added an unused constant there.

fredbi: There is also an undue change in pkg/store errors: added an unused constant there.

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+//   err := core.ListBundlesApply(repo, store, func(bundle model.BundleDescriptorOption) error {
+//				fmt.Fprintf(os.Stderr, "%v\n", bundle)
+//				return nil
+//			})
+func ListBundlesApply(repo string, store storage.Store, apply ApplyBundleFunc, opts ...BundleListOption) error {
+	var (
+		err, applyErr error
+		wg            sync.WaitGroup
+		workDone      bool
+		mutex         sync.Mutex
```
ransomw1c: [the documentation](https://golang.org/pkg/sync/) mentions

> Other than the Once and WaitGroup types, most are intended for use by low-level library routines.

is there a way to communicate other than the shared `workDone` and lock?

fredbi: Mutex is commonplace.

I admit this comes a bit out of the nice full-channel picture... I thought about usIng a signalling channel again... in the end mutex expressed well my intent: first-past-the-post wins

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+			if !workDone {
+				doneChan <- struct{}{}
+			}
+			mutex.Unlock()
+			for range bundleChan {
+			} // wait for close
+			break
+		}
+	}
+	wg.Wait()
```
ransomw1c: with the block on `close(bundleChan)`, the `wg sync.WaitGroup` is now superfluous

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+
+	// collect bundle metadata asynchronously
+	wg.Add(1)
+	go func(bundleChan chan<- model.BundleDescriptor, doneChan chan struct{}, wg *sync.WaitGroup) {
+		defer func() {
+			close(bundleChan)
+			wg.Done()
+		}()
+
+		bundlesChan, errChan, workers := ListBundlesChan(repo, store, append(opts, WithBundleDoneChan(doneChan))...)
```
ransomw1c: `listBundlesChan` and `doSelectBundles`, perhaps?

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+
+		err = DoSelectBundles(bundlesChan, errChan, func(bundleBatch model.BundleDescriptors) {
```
ransomw1c: supposing this is, in fact a package local function -- and perhaps even if it's not -- i'd favor discarding the callback passing and give write access to `bundleChan`

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+
+	workers.Wait()
+
+	return bundles, err // we may have some batches resolved before the error occurred
+}
+
+// ListBundlesChan returns a list of bundle descriptors from a repo. Each batch of returned descriptors
+// is sent on the output channel, following key lexicographic order.
+//
+// Simple use cases of this helper are wrapped in ListBundles (block until completion) and ListBundlesApply
```
ransomw1c: let's remove the cruft:  if there's no dependency chain of `ListBundles` percolating to the ui, there's no need for `ListBundles`

ransomw1c: .. although it could be reasonable to add another commit to the mainline for the sake of having the bifurcated (sync/async) functionality in history

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+				bundleChan <- bundle // transfer a batch of metadata to the applied func
+			}
+		})
+
+		mutex.Lock()
+		workDone = true // nothing left to interrupt
+		close(doneChan)
+		mutex.Unlock()
+
+		workers.Wait()
```
ransomw1c: again, prefer waiting on channels (with this `for range c {}` idiom) to a waitgroup when the two are equivalent.

reason:  less mutable state is simpler than less procedural data.

ransomw1c:
note to self on "signalling channels" -- i.e. channels with type `struct{}` with a buffer size of one is that these channels convey two types doneness as encoded in `watchForInterrupts`
* "interrupts" wherein the channel is closed and empty
* non-"interrupts" wherein the channel has a value written to it

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+		close(outputChan)
+		wg.Done()
+	}()
+
+	var (
+		err                            error
+		errorChan1Open, errorChan2Open bool = true, true
+	)
+	for errorChan1Open || errorChan2Open { // wait on both channels to close
+		select {
```
ransomw1c: [recall](https://golang.org/ref/spec#Select_statements)

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection

so to prevent the enclosing `for` loop from being stochastic, i suggest adding a `switch` to read from precisely one or the other channel once one of the two is closed.

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+	)
+
+	for {
+		// get a batch of keys
+		ks, next, err = store.KeysPrefix(context.Background(), next, model.GetArchivePathPrefixToBundles(repo), "/", settings.bundleBatchSize)
+		if err != nil {
+			select {
+			case errorChan <- err:
+				break
+			case <-doneChan:
```
ransomw1c: here, we don't preserve the meaning of "interrupt" in the sense of `watchForInterrupts` via, e.g.,

```
case _, interrupt <-doneChan:
	if interrupt { errorChan <- ErrInterrupted }
```

for some reason?

pkg/core/bundle_list.go
```diff
@@ -25,29 +28,339 @@ func minInt(a, b int) int {
...
+		close(batchChan)
+		close(errorChan)
+		wg.Done()
+	}()
+
+	for {
+		select {
+		case <-doneChan:
+			errorChan <- ErrInterrupted
+			return
```
ransomw1c: nitpick or possibly stylistic elucidation:  what `return` here where elsewhere `break`?

why this could be stylistic is that, for instance, i use double quotes for human readable strings in scripting languages like js and py and single quotes for keyword-like values.  is this a similar idiosyncratic convention where `return` consistently connotes a specific meaning wrt. programming patterns?

cmd/datamon/cmd/bundle_list.go
```diff
@@ -7,14 +7,16 @@ import (
...
 
 	"github.com/spf13/cobra"
 )
 
+// BundleListCommand describes the CLI command for listing bundles
 var BundleListCommand = &cobra.Command{
 	Use:   "list",
 	Short: "List bundles",
-	Long:  "List the bundles in a repo",
+	Long:  "List the bundles in a repo, ordered by their key",
```
kerneltime: ```ordered by their bundle ID``` there is no entity in the model which is a key, that is an internal implementation.

cmd/datamon/cmd/bundle_list.go
```diff
@@ -23,17 +25,18 @@ var BundleListCommand = &cobra.Command{
 		if err != nil {
 			logFatalln(err)
 		}
-		bundleDescriptors, err := core.ListBundles(params.repo.RepoName, remoteStores.meta, core.ConcurrentBundleList(params.core.ConcurrencyFactor))
-		if err != nil {
-			logFatalln(err)
-		}
-		for _, bd := range bundleDescriptors {
+		err = core.ListBundlesApply(params.repo.RepoName, remoteStores.meta, func(bundle model.BundleDescriptor) error {
```
kerneltime: I think this works in go but it is ugly to read.. can we define a function before we pass it?

cmd/datamon/cmd/cli_test.go
```diff
@@ -493,12 +493,15 @@ func listBundles(t *testing.T, repoName string) (bundleListEntries, error) {
...
 		rle := bundleListEntry{
 			rawLine: line,
-			hash:    strings.TrimSpace(sl[0]),
+			hash:    strings.TrimSpace(sl[0]), // bundle key ID
 			message: strings.TrimSpace(sl[2]),
 			time:    t,
 		}
 		bles = append(bles, rle)
 	}
+	// bundles are ordered by lexicographic order of keys (ksuids).
```
kerneltime: Drop ksuid here it is an internal implementation of the core package.
Also, use Bundle ID

cmd/datamon/cmd/flags.go
```diff
@@ -143,11 +144,19 @@ func addCoreConcurrencyFactorFlag(cmd *cobra.Command) string {
...
+		"Heuristic on the amount of concurrency used by core operations (e.g. bundle list). "+
+			"Concurrent retrieval of bundle metadata is capped by the 'batch-size' parameter. "+
 			"Turn this value down to use less memory, increase for faster operations.")
 	return concurrencyFactor
 }
 
+func addBatchSizeFlag(cmd *cobra.Command) string {
+	batchSize := "batch-size"
+	cmd.Flags().IntVar(&params.core.BatchSize, batchSize, 1024,
+		"Number of bundle keys fetched per rountrip to storage -- keys are 20B long, defaults corresponds to 20kB pages")
```
kerneltime: ```Number of bundles streamed to out together as a batch. This can be tuned for performance based on network connectivity```

pkg/core/bundle.go
```diff
@@ -197,6 +203,31 @@ func ConcurrentBundleList(concurrentBundleList int) BundleListOption {
 	}
 }
 
+// BundleBatchSize sets the batch window to fetch bundle keys. It defaults to 100.
```
kerneltime: Drop ```It defaults to 100```.. ``` defaults to defaultBundleBatchSize```

cmd/datamon/cmd/bundle_list.go
```diff
@@ -23,17 +25,18 @@ var BundleListCommand = &cobra.Command{
...
 			var buf bytes.Buffer
-			err := listLineTemplate.Execute(&buf, bd)
-			if err != nil {
-				log.Println("executing template:", err)
+			e := listLineTemplate.Execute(&buf, bundle)
+			if e != nil {
+				log.Println("executing template:", e)
```
kerneltime: This was not caught earlier but we should use fmt or change the logger to print to STDOUT. Right now this is printing to STDERR which makes it hard to script the output.

pkg/storage/status/status.go
```diff
@@ -0,0 +1,12 @@
+// Package status declares error constants returned by the variou
```
kerneltime: typo

pkg/storage/status/status.go
```diff
@@ -0,0 +1,12 @@
...
+// implementations of the Store interface.
+package status
```
kerneltime: Why not have it in the storage package?

pkg/core/bundle_list.go
fredbi: Once will do it better.

pkg/core/bundle_list.go
fredbi: Ok.

pkg/core/bundle_list.go
fredbi: Used by web utility

pkg/storage/status/status.go
fredbi: @kerneltime because it creates a new constraint re not having cyclical dependencies. With a separate case package i am not adding any such implicit constraint on thid pkg

pkg/core/bundle_list.go
fredbi: I disagree: although the bulk of the work is done, there may be some goroutines pending (waiting for doneChannels etc).

**However**, I agree that the Wait here is optional and we can explicitly allow for goroutine leak. This was not my initial choice (see test for leaked goroutines).

NOTE: usage is not strictly limited to CLI, I discovered we have a web server too. This one must not leak goroutines.

pkg/core/bundle_list.go
fredbi: These funcs are typically what you would use to build the _porcelain_ layer you have been talking about. Agree that we can defer making them public to a more appropriate time.

pkg/core/bundle_list.go
fredbi: ok to make DoSelect a private func.
You'll see in #267, #268 that it can be simplified to the point it could even disappear as a valuable helper.
Don't totally adhere to removing the callback. Need time to think about that. Thanks for raising this point.

pkg/core/bundle_list.go
fredbi: ok with the principle but this is not equivalent

pkg/core/bundle_list.go
ransomw1c: at this time, not pre-optimizing for lispification

cmd/datamon/cmd/bundle_list.go
fredbi: ok

cmd/datamon/cmd/bundle_list.go
fredbi: ok if that is your preference.

cmd/datamon/cmd/cli_test.go
fredbi: ok

cmd/datamon/cmd/flags.go
fredbi: ok

pkg/core/bundle.go
fredbi: ok

pkg/core/bundle_list.go
fredbi: break would just get us out from the `select`, not the loop as intended.

pkg/core/bundle_list.go
```diff
@@ -57,37 +370,42 @@ func ListBundles(repo string, store storage.Store, opts ...BundleListOption) ([]
...
 
-	for i := 0; i < minInt(settings.concurrentBundleList, len(ks)); i++ {
+	// spin workers pool
+	for i := 0; i < minInt(settings.concurrentBundleList, len(keys)); i++ {
 		workers.Add(1)
 		go getBundleAsync(repo, store, keyChan, bundleChan, errorChan, &workers)
 	}
 
-	bds := make([]model.BundleDescriptor, 0, typicalBundlesNum) // preallocate some typical number of bundles, e.g. 1000
+	bds := make(model.BundleDescriptors, 0, settings.bundleBatchSize)
```
ransomw1c: noticed `len(keys)` instead of `bundleBatchSize` in #267 .. if this is to land as a separate commit, use `len(keys)` everywhere for consistency within `fetchBundleBatch`

cmd/datamon/cmd/bundle_list.go
fredbi: sure. Will do.

fredbi:
Closing this and taking the raised points in #267

pkg/storage/store.go
```diff
@@ -25,6 +25,7 @@ const (
 	ErrNotSupported errString = "not supported"
 	ErrExists       errString = "exists already"
 	ErrObjectTooBig errString = "object too big to be read into memory"
+	ErrNotExists    errString = "object doesn't exist"
```
fredbi: This one is superfluous.

pkg/core/bundle_list.go
fredbi: ok understood. Removed unnessary waitgroup and wait here. Also removed explicit close of internal doneChannels, letting the gc do the cleaning work (documented this quirk as a comment).

pkg/core/bundle_list.go
fredbi: taking advantage of this to move event types private in #267 (thus reducing the exposed symbols)

PR #267, opened 2019-10-06
feat(bundle): Channel simplified struct

fredbi:
Improves #266, using the ` chan struct {...}`  pattern. Specifically trying to address:
> The error handling in the goroutines code is complex and required a lot of testing for blocking corner cases. This can be greatly improved by multiplexing values and errors on the same channel. At the moment, I am hesitating between chan struct{value: value_t, err: error} and chan func() (value_t, error). Your advice is welcome on this choice.

See #268 to compare different approaches to multiplexing.

* same functionality with less code thanks to channel w/error multiplexing
* enhanced parallel bundle retrieval: starve workers immediately on error


cmd/datamon/cmd/bundle_list.go
```diff
@@ -7,14 +7,16 @@ import (
...
 
 	"github.com/spf13/cobra"
 )
 
+// BundleListCommand describes the CLI command for listing bundles
 var BundleListCommand = &cobra.Command{
 	Use:   "list",
 	Short: "List bundles",
-	Long:  "List the bundles in a repo",
+	Long:  "List the bundles in a repo, ordered by their key",
```
kerneltime: ordered by their bundle ID. There is no concept of keys in the CLI model for datamon, that is internal.

Issue #192, opened 2019-06-03
standard io selection 


ransomw1c:
currently, the output of list and get commands (`repo list`, etc.) is sent to stderr rather than stdout.

this issue is to make a thorough-going over of which of the standard output channels is used to print various types of messages from the executable.

this issue could probably be re-named.

ransomw1c:
upon addressing this iss., also remove the "Using config file" message.

fredbi:
fixed with #375

PR #269, opened 2019-10-11
Linting & errors reporting

fredbi:
* moved linting rules settings from CI config to .golangci.yml
* applied a few linting
* revisited error reporting

### DIFF
--- a/cmd/datamon/cmd/root.go
+++ b/cmd/datamon/cmd/root.go
@@ -49,6 +49,19 @@ var logFatalln = log.Fatalln
 var logFatalf = log.Fatalf
 var osExit = os.Exit
 
+// infoLogger wraps informative messages to os.Stdout without cluttering expected output in tests.
+// To be used instead on fmt.Printf(os.Stdout, ...)
+var infoLogger = log.New(os.Stdout, "", 0)
+
+func wrapFatalln(msg string, err error) {
+	logFatalf("%v", fmt.Errorf(msg+": %w", err))
+}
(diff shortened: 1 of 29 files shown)


ransomw1c:
intersects #192

cmd/datamon/cmd/root.go
```diff
@@ -49,6 +49,19 @@ var logFatalln = log.Fatalln
 var logFatalf = log.Fatalf
 var osExit = os.Exit
 
+// infoLogger wraps informative messages to os.Stdout without cluttering expected output in tests.
+// To be used instead on fmt.Printf(os.Stdout, ...)
+var infoLogger = log.New(os.Stdout, "", 0)
+
+func wrapFatalln(msg string, err error) {
```
ransomw1c: can this replace `logFatalf` everywhere?

fredbi: @ransomw1c, well it could if `wrapFatalf` interprets a `nil` as `err` like `logFataln` does. I can do that if you prefer (and yes this would make it only 1 func to mock). Deal?

ransomw1c: to clarify, the motivating principle is that of consistency:  i'd prefer not to have two functions with the same semantic meaning of 

> report an error and exit

i'm confident that your choices about api design are reasonable.

fredbi: @ransomw1c ok will merge both funcs then

PR #270, opened 2019-10-14
[wip] FUSE, PG sidecar reparameterization

ransomw1c:
corresponding UI for FUSE sidecar to PG sidecar

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,59 +218,21 @@ where
   [conventional UNIX syntax](http://zsh.sourceforge.net/Guide/zshguide02.html#l11)
   for stating that options to a command are done being declared.
 
-Meanwhile, each sidecar's datamon-specific batteries have their corresponding usages.
-
-##### `gcr.io/onec-co/datamon-fuse-sidecar` -- `wrap_datamon.sh`
-
-Provides filesystem representations (i.e. a folder) of [datamon bundles](#data-modeling).
-Since bundles' filelists are serialized filesystem representations,
-the `wrap_datamon.sh` interface is tightly coupled to that of the self-documenting
-`datamon` binary itself.
-
-```shell
-./wrap_datamon.sh -c <coord_dir> -d <bin_cmd_I> -d <bin_cmd_J> ...
-```
-
-* `-c` the same coordination directory passed to `wrap_application.sh`
-* `-d` all parameters, exactly as passed to the datamon binary, except as a
-  single scalar (quoted) parameter, for one of the following commands
-  - `config` sets user information associated with any bundles created by the node
-  - `bundle mount` provides sources for data-science applications
-  - `bundle upload` provides sinks for data-science applications
-
-Multiple (or none) `bundle mount` and `bundle upload` commands may be specified,
-and at most one `config` command is allowed so that an example `wrap_datamon.sh`
-YAML might be
-
-```yaml
-command: ["./wrap_datamon.sh"]
-args: ["-c", "/tmp/coord", "-d", "config create --name \"Coord\" --email coord-bot@oneconcern.com", "-d", "bundle upload --path /tmp/upload --message \"result of container coordination demo\" --repo ransom-datamon-test-repo --label coordemo", "-d", "bundle mount --repo ransom-datamon-test-repo --label testlabel --mount /tmp/mount --stream"]
-```
-
-or from the shell
-
-```shell
-./wrap_datamon.sh -c /tmp/coord -d 'config create --name "Coord" --email coord-bot@oneconcern.com' -d 'bundle upload --path /tmp/upload --message "result of container coordination demo" --repo ransom-datamon-test-repo --label coordemo' -d 'bundle mount --repo ransom-datamon-test-repo --label testlabel --mount /tmp/mount --stream'
-```
-
-##### `gcr.io/onec-co/datamon-pg-sidecar` -- `wrap_datamon_pg.sh`
-
-Provides Postgres databases as bundles and vice versa.
-Since the datamon binary does not include any Postgres-specific notions,
-the UI here is more decoupled than that of `wrap_datamon.sh`.
-The UI is specified via environment variables
-such that `wrap_datamon.sh` is invoked without parameters.
-
-The script looks for precisely one `dm_pg_opts` environment variable
-specifying global options for the entire script and any number of
-`dm_pg_db_<db_id>` variables, one per database.
+Meanwhile, each sidecar's datamon-specific batteries have their corresponding
+usages.
+Despite their differences, both are parameterized via environment variables.
+[12 factor](https://www.12factor.net/config) principles are loosely part
+of the overall ideology with the strong opinion being that of the virtue of
+declarative specification.
+The environment variables for sidecars both use the same serialization
+format to represent dictionaries (aka associative arrays) with strings.
 
 ----
 
 _Aside on serialization format_
 
-Each of these environment variables each contain a serialized dictionary
-according the the following format
+Each of environment variable parsed by the sidecars contains a serialized
+dictionary according the the following format
 
 ```
 <entry_sperator><key_value_seperator><entry_1><entry_seperator><entry_2>...
@@ -286,7 +248,7 @@ So for example
 ;:a;b:c
 ```
 
-expresses something like a Python map
+expresses something like a Python `dict`
 
 ```python
 {'a': True, 'b' : 'c'}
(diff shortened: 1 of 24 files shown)


README.md
```diff
@@ -218,59 +218,21 @@ where
...
+The environment variables for sidecars both use the same serialization
+format to represent dictionaries (aka associative arrays) with strings.
 
 ----
 
 _Aside on serialization format_
 
-Each of these environment variables each contain a serialized dictionary
-according the the following format
+Each of environment variable parsed by the sidecars contains a serialized
```
kerneltime: If the dictionary is fixed list of strings, can they be their own environment variable vs packing them into the value of an environment variable.

hack/fuse-demo/wrap_datamon.sh
```diff
@@ -8,98 +8,232 @@
...
+
+# deserialize an associate array from a scalar
+# example usage
+# typeset -A opts_global_dict
+# opts_global_dict=($(deserialize_dict $opts_global))
+# todo: implement via set variables by reference as in
+# deserialize_dict $opts_global opts_global_dict
+#   .. access by reference is via ${(P)var_name}.
+#   unsure how to set.
+deserialize_dict() {
```
kerneltime: All of this can be avoided if we have a set of known environment variables.
$DM_REPO
$DM_BUNDLE_LABLE
$DM_FUSE_STREAM

kerneltime: @galvare2 would you prefer to generate environment variables as dictionaries or have independent environment variables?

README.md
```diff
@@ -300,6 +262,47 @@ or shell option args
...
+
+* `c` the same coordination directory passed to `wrap_application.sh`
+* `n`, `e` name and email for datamon contributor record
+* `S` without an `<arg>` is causes the wrapper script to sleep instead
+  of exiting, which can be useful for debug.
+
+The per-bundle options (one per `<id>`,
+a unique string among all bundle options) determine whether the bundle
+is a _source_, data that exist in datamon and accessed by sidecar,
+or a _destination_, data that are created by the sidecar.
```
galvare2: > created by the sidecar
Aren't they created by the main container, not the sidecar?

hack/fuse-demo/wrap_datamon.sh
galvare2: It's unimportant to me, but I will say that this serialization scheme seems a bit complicated, so whichever implementation reduces the complexity is fine with me. Perhaps it would be preferable to use a standard, well-known serialization such as json or similar?

ransomw1c:
toggle back to **wip** pending 

* golang sidecar parameter `struct` definition
* golang serialization/deserialization routines of said `struct`
  - YAML
  - json
  - string dict format
* zsh wrap of golang parameter parsing

where the last includes

* ingestion of parameters
  - from precisely one environment variable that may contain either a file path or the serialized parameters
  - according to any format via fallback on errors
* bridge to internal zsh variables required by existing implementation over either
  - environment variables
  - `.conf` files

----

in particular, the `struct` de/serialization routines are to be versioned along with the rest of the datamon, so while there's no need to direct product the sidecar _version_ as 

```
(datamon) ⊙ (serialization) ⊙ (zsh wrapper)
```

these will remain components of the planned implementation

README.md
ransomw1c: > created by the main container and stored in datamon by the sidecar

is a more full explain

> created by the main container

is also a possible re-phrase...
... erring on the side of terseness with the readme, overall.

could change to the more verbose explain or `s/sidecar/main container/`.  no particularly strong opines here.

README.md
ransomw1c: this is the *bridge to internal zsh variables* section of [the batch comment](https://github.com/oneconcern/datamon/pull/270#issuecomment-543291247).

there are multiple options for the bridge.  the list in the comment is probably not even exhaustive.

hack/fuse-demo/wrap_datamon.sh
ransomw1c: to be an internal implementation detail occluded by pending golang mentioned in [the comment](https://github.com/oneconcern/datamon/pull/270#issuecomment-543291247) following this morning's discussion

sidecar.Dockerfile
```diff
@@ -15,6 +15,27 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 RUN echo "allow_root" >> /etc/fuse.conf
 
+## BEGIN tini
+
+ENV TINI_VERSION v0.18.0
```
fredbi: Do we really need to install tini? Recent docker includes a `docker  run --init` [option](https://docs.docker.com/engine/reference/commandline/container_run/) . As per the author of tini himself (https://github.com/krallin/tini#using-tini) we should no more have to install this. To be tested, of course...

hack/fuse-demo/run_coord_shell.sh
```diff
@@ -15,14 +15,16 @@ while getopts s opt; do
...
-pod_name=$(kubectl get pods -l app=datamon-coord-demo | grep Running | sed 's/ .*//')
+print -- "waiting on pod start"
 
-if [[ -z $pod_name ]]; then
-	echo 'coord demo pod not found' 1>&2
-	exit 1
-fi
+while [[ -z $pod_name ]]; do
+    sleep 1
+    pod_name=$(kubectl get pods -l app=datamon-coord-demo | \
```
fredbi: Just saying here... From my experience, having a pod in a running state does'nt mean its service is ready. Don't we have a ready probe that could be queried here? Like:
```
until printf "." && curl -fs --insecure --connect-timeout 2 https://localhost:8443/auth >/dev/null ; do
  sleep 2;
done
```
Or use `nc` instead of `curl`

hack/fuse-demo/coord-app.Dockerfile
```diff
@@ -46,5 +47,7 @@ RUN chmod a+x mock_application.sh
 
 USER developer
 
+RUN touch ~/.zshrc
+
 ENTRYPOINT [ "/tmp/coord/.scripts/wrap_application.sh"]
```
fredbi: I am just curious about why we set the entrypoint on `/tmp`. This fs should normally be mounted as `tmpfs` and hence be empty when the container starts. The fact that it just works is bugging me ....

hack/fuse-demo/verify_coord_bundle.sh
```diff
@@ -11,12 +11,16 @@ start_timestamp=$(cat /tmp/datamon_fuse_demo_coord_start_timestamp)
 
 verify_datamon_timestamp() {
     timestamp="$1"
-    timestamp_to_parse=$(echo "$timestamp" | sed 's/\.[^ ]*//g' | sed 's/ *[^ ]*$//')
+    timestamp_to_parse=$(echo "$timestamp" | \
+                             sed 's/\.[^ ]*//g' | sed 's/ *[^ ]*$//')
     epoch_timestamp=$(date -jf '%Y-%m-%d %H:%M:%S %z' "$timestamp_to_parse" '+%s')
-    sec_from_start=$(echo "$epoch_timestamp - $start_timestamp" |bc)
+    start_timestamp_local=$(date -jf  '%s' "$start_timestamp" \
```
fredbi: mmh I don't know the `-j` flag for date. This is most likely non standard.

fredbi: When it comes to parsing dates into timestamps, I usually do something like `date --date "${timestamp}" '+%s'`(GNU date). Wouldn't that help clarifying this section?

sidecar.Dockerfile
ransomw1c: the tini use is a copy 'n paste job throughout.

i can consider it technical debt, yet i'm pressed to consider it part of the scope of this patch.

hack/fuse-demo/run_coord_shell.sh
ransomw1c: understood.

this is more stuff that "just works."  moreover, for the purposes of starting a shell within the pod, the intent of this script, having the pod in a Running state could be precisely what's necessary?

conversely, it should be understood that the "service layer" is the cross-container coordination mechanism, which is currently implemented as files that could be stat'ed.  and yes, there could be an additional event(s) added to `wrap_application.sh` for health checks, etc.  this is another item i'm willing to file under technical debt and cleanup as part of this merge request or a later one, depending on how much time is left after the requirements management and functionality is pieces are settled.

hack/fuse-demo/coord-app.Dockerfile
ransomw1c: i used `/tmp` based on familiarity with *nix systems.  unsure what tradition `tmpfs` corresponds to.

hack/fuse-demo/verify_coord_bundle.sh
ransomw1c: unclear why gnu is more or less of a standard than freebsd

cmd/sidecar_param/cmd/root.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "sidecar_param",
```
kerneltime: just param?

ransomw1c:
this to merge request needs to be redirected to a different branch in the forthcoming git branch model mirroring semver

ransomw1c:
v2.0 requires FUSE only, not PG.

there are still some changes from this branch that'll be useful for FUSE.  do not delete branch.

PR #271, opened 2019-10-17
[WIP] Datamon 2.0 WAL

kerneltime:
This is a WIP commit for the WAL.
ToDo:
1. Add test for parallel listing of WAL
2. Add documentation


kerneltime:
The build failure is due to go mod updates, will fix it.

pkg/wal/wal.go
```diff
@@ -0,0 +1,345 @@
...
+		option(&wal)
+	}
+	// Check if token generator object exists
+	_ = mutableStore.Put(context.Background(), wal.tokenGeneratorPath, strings.NewReader(""), storage.OverWrite)
+	return &wal
+}
+
+// Gets a token such that tokens are K-sortable.
+func (w *WAL) getToken(ctx context.Context) (string, error) {
+	err := w.updateTokenTimestamp(ctx)
```
ransomw1c: i think that K-sortable (aka partially sortable) is due to the total sort on time.  hence, i'm unsure why it's not necessary to, for example, acquire locks from remote. 

----

if there are two locals, A and B, at least two things can happen

- A tells the remote to look at the time (t1)
- A makes ksuid time from remote (w/ t1)
- B tells the remote to look at the time (t2)
- B makes ksuid time from remote (w/ t2)

in this case, sequential times lead to sequential ksuids

- A tells the remote to look at the time (t1)
- B tells the remote to look at the time (t2)
- A makes ksuid (k1) time from remote (w/ t2)
- B makes ksuid (k2) time from remote (w/ t2)

in this case, is there still a meaningful relation between k1 and k2?

pkg/wal/wal.go
```diff
@@ -0,0 +1,345 @@
...
+	if err != nil {
+		return nil, "", err
+	}
+	if max > 1000 {
+		max = 1000
+	}
+
+	// Go back in time to include the keys that might have been written before the token was generated.
+	ksuidOld := k.Time().Add(-w.GetExpirationDuration() * 2)
+	tokens, next, err = w.walStore.KeysPrefix(ctx, ksuidOld.String(), "", "", max)
```
ransomw1c: i'm slightly confused... `ksuidOld` is a `Time` instance.

b/c both the `Time` here and the `KSUID` in `getToken` that's used in the corresponding`walStore.Put` in `Add` are used _via casts to string_, some info that might otherwise be decipherable via the type system is probably being lost on me.

at the possible expense of DSLification, could there be some functions added to cast ksuids and times to strings?  or perhaps some comments on the relation between the time string and the ksuid string would suffice?

pkg/wal/wal.go
kerneltime: KSUID is a combination of a timestamp and a random string. Within a second granularity, the uniqueness depends on randomness. [KSUID how do they work](https://github.com/segmentio/ksuid#how-do-they-work)
For Datamon 2 clients uploading 2 bundles or even the same client uploading the same bundle does not have any need for strong linearization and there is no happens-before relationship.
Thus, in your example there is no relation between k1 and k2 and they would get sorted based on randomness if within the same second. For any semantic tracking of bundles, the application should depend on the labels they assign and not on the sort order. 
The current model of the latest label works if there is only one process to generate a dataset. Since there is no parent-child relationship or a single child policy for a parent between bundles, there is no real concept of HEAD as used in git. 
In the new design, the combination of pipeline run for production being controlled along with semantic versioning and schema versioning will provide the appropriate model on the consumption side for the bundles.
Will include documentation within Datamon 2.0 release for the consumption model.

pkg/wal/wal.go
kerneltime: I should port the design over from the internal documents here.
The main idea is to give a model for the clients of WAL to write their intent to the WAL, get an expiration limit in which they have to finish the work. This is to allow for a listing of WAL that does not need to start from the first entry in the WAL (clients writing to back in time in the WAL due to asynchronous model in play).

pkg/wal/wal.go
```diff
@@ -0,0 +1,345 @@
...
+			return nil, "", err
+		}
+	}
+}
+
+func (w *WAL) issueParallelReads(ctx context.Context, channels *walChannels)  {
+	count := 0
+	for {
+		select {
+		case tokens, ok := <- channels.tokens:
```
ransomw1c: nitpick: i learned about `for tokens := range channels.tokens` from @fredbi 's merge requests.  it is probably a more idiomatic construct to use.

pkg/wal/wal.go
```diff
@@ -0,0 +1,345 @@
...
+			_, updated := entriesMap.Insert([]byte(e.Token), e)
+			if updated {
+				panic(fmt.Sprintf("Received more than one response for token: %s string: %s", e.Token, e.Payload))
+			}
+			count++
+			if count == total {
+				finalize()
+				return
+			}
+		case total = <- channels.count:
```
ransomw1c: haven't fully verified, yet it appears adding the `finalize` conditional to this case as well could ensure correctness?

impression is that it hasn't been necessary to add because the latency thru `walStore` is more than the time from

```
walChannels.tokens <- tokens
```

to

```
walChannels.count <- count
```

in `ListEntries` ... and while this is probably true in all use-cases, i'm in favor of removing such a tacit assumption (if that's even what's going on in the first place 0_o).

pkg/wal/wal.go
fredbi: @ransomw1c the `select { ...}` is actually the inner working of the `range` operator for channels.

pkg/wal/wal.go
ransomw1c: since i have seen [the format of random ksuids with time](https://github.com/segmentio/ksuid/blob/master/ksuid.go#L217), namely `<time><rand>`, i now understand the cohesion<>decoupling tradeoffs involved in the string casts and consider it maintainable.

still some slight vexation about what amounts to explicitly phrasing assumptions about the upstream deps within the datamon source, yet any such are far from blocking.

pkg/wal/wal.go
ransomw1c: to whit, that's the reason i also personally prefer chunking by idiom

pkg/wal/wal.go
```diff
@@ -0,0 +1,345 @@
...
+	if err != nil {
+		channels.oops <- fmt.Errorf("token: %s, err: %s", token, err)
+		return
+	}
+	channels.entry <- entry
+	return
+}
+
+func (w *WAL) collectParallelResponses(ctx context.Context, channels *walChannels) {
+	entriesMap := iradix.New().Txn()
```
ransomw1c: out of partially idle curiosity, what's the radix tree buying here?  space, time, both?  is it perf we need?  the complexity cost is admittedly low-ish.

ransomw1c:
the parallelism setup based on count

count

is considered super

go.mod
```diff
@@ -1,75 +1,35 @@
...
-	k8s.io/apimachinery v0.0.0-20190531161113-d9689afd32c1 // indirect
-	k8s.io/kubernetes v1.14.2
-	k8s.io/utils v0.0.0-20190520173318-324c5df7d3f0 // indirect
-	mvdan.cc/unparam v0.0.0-20190917161559-b83a221c10a2 // indirect
-	sourcegraph.com/sqs/pbtypes v1.0.0 // indirect
+	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550 // indirect
+	golang.org/x/sys v0.0.0-20191010194322-b09406accb47
+	golang.org/x/tools v0.0.0-20191015211201-9c6d90b5a7d0 // indirect
+	google.golang.org/api v0.11.0
+	gopkg.in/yaml.v2 v2.2.2
```
fredbi: why rollbacking yaml to 2.2.2? IIRC, 2.2.4 brought a security fix.

pkg/core/bundle_list_test.go
```diff
@@ -452,7 +452,7 @@ const (
 
 func testListBundles(t *testing.T, concurrency int, i int) {
 	initBatchKeysFixture.Do(buildKeysBatchFixture(t))
-	defer goleak.VerifyNone(t)
+	defer goleak.VerifyNoLeaks(t)
```
fredbi: ok got it, hence the release pin...

pkg/storage/gcs/store_test.go
```diff
@@ -21,87 +24,111 @@ const (
...
+	require.NoError(t, err, "Failed to create bucket:"+bucket)
 
 	gcs, err := New(context.TODO(), bucket, "") // Use GOOGLE_APPLICATION_CREDENTIALS env variable
-	require.NoError(t, err)
+	require.NoError(t, err, "failed to create gcs client")
+	wg := sync.WaitGroup{}
+	create := func(i int, wg *sync.WaitGroup) {
+		err = gcs.Put(ctx, constStringWithIndex(i), bytes.NewBufferString(constStringWithIndex(i)), storage.NoOverWrite)
+		require.NoError(t, err, "Index at: "+fmt.Sprint(i))
+		wg.Done()
```
fredbi: `defer wg.Done()`  more idiomatic and more correct when require exits upon error

pkg/storage/gcs/store_test.go
```diff
@@ -21,87 +24,111 @@ const (
...
+		wg.Add(1)
+		go create(i, &wg)
 	}
+	wg.Wait()
 
 	cleanup := func() {
+		delete := func(key string, wg *sync.WaitGroup) {
+			err = gcs.Delete(ctx, key)
+			require.NoError(t, err, "failed to delete:"+key)
+			wg.Done()
```
fredbi: `defer wg.Done()` more idiomatic and more correct when require exits upon error

pkg/storage/store.go
```diff
@@ -15,10 +16,13 @@ const MaxObjectSizeInMemory = 2 * 1024 * 1024 * 1024 // 2 gigs
...
 
 const (
-	IfNotPresent = true
-	OverWrite    = false
+	// Adding these to make code more readable when looking at Put Call
+	NoOverWrite = true
+	OverWrite   = false
 )
 
+type NewKey bool
```
fredbi: declaring NewKey as a type alias instead of a full-fledged type Would make conversions unnecessary.
`type NewKey = bool`

pkg/wal/model.go
```diff
@@ -0,0 +1,31 @@
...
+const maxConcurrency = 1024
+
+func NewEntry(token string, payload string) *Entry {
+	return &Entry{
+		Token:   token,
+		Payload: payload,
+	}
+}
+
+func Unmarshal(b []byte) (*Entry, error) {
```
fredbi: I think we should rather implement here the YAML unmarshaller interface: https://github.com/go-yaml/yaml/blob/v2.2.4/yaml.go#L32

pkg/wal/wal.go
```diff
@@ -0,0 +1,345 @@
...
+
+	for {
+		select {
+		case entries := <- walChannels.entries:
+			if len(entries) != count {
+				panic(fmt.Sprintf("received count different than the list of tokens entries: %d count: %d", len(entries), count))
+			}
+			return entries, fromToken, nil
+		case err := <- walChannels.oops:
+			return nil, "", err
```
fredbi: ok so maybe add a comment here that we are willingly leaking routines

pkg/wal/wal.go
```diff
@@ -0,0 +1,345 @@
...
+	"github.com/segmentio/ksuid"
+
+	"github.com/oneconcern/datamon/pkg/storage"
+)
+
+type WAL struct {
+	mutableStore       storage.Store  // Location for updating token generator object
+	tokenGeneratorPath string        	// Path to the token generator object in the mutable store
+	walStore           storage.Store  // Append only store where WAL entries are written to
+	maxConcurrency     int           	// Max concurrency when reading
```
fredbi: I am not seeing clearly where this is used. I don't see anything with the go routines pattern that looks like a worker pool. Is that for the future?

go.mod
```diff
@@ -1,75 +1,35 @@
 module github.com/oneconcern/datamon
 
-replace google.golang.org/api => github.com/googleapis/google-api-go-client v0.2.1-0.20190318183801-2dc3ad4d67ba
-
-replace github.com/spf13/cobra => github.com/babysnakes/cobra v0.0.2-0.20180603190830-61ca3af7ef22
-
-replace github.com/spf13/pflag => github.com/spf13/pflag v1.0.3
-
-replace go.uber.org/goleak => go.uber.org/goleak v0.10.1-0.20190823232112-227bd74c3482
```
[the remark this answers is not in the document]
kerneltime: Ok, we need to see how to either update the tests or remove this dependency. Will revert it for now

go.mod
kerneltime: Thank you

pkg/storage/store.go
kerneltime: Thank you!

pkg/wal/model.go
kerneltime: Makes sense. Will do it.

pkg/wal/wal.go
kerneltime: I am not sure we are leading here, there is only one goroutine that reports into these channels and it will report success or a failure. ```collectParallelResponses``` collects parallel responses and at the end reports one of the errors (After logging them all) or reports a success.

pkg/wal/wal.go
kerneltime: Next update has it.

pkg/wal/wal.go
fredbi:
```
// Start routine which collects all parallel responses.
	go w.collectParallelResponses(ctx, walChannels)
	// Start routine that will issue parallel reads.
	go w.issueParallelReads(ctx, walChannels)
```
are not waited for when returning, which is what I am calling leaking routines.
This is fine by me provided you make it explicit that this is on purpose and not a bug by omission.

pkg/wal/wal.go
kerneltime: When we receive responses we need to store them sorted. It was either this or pass an ordered map of the keys to expect and then validate if we received all responses.

pkg/wal/wal.go
kerneltime: Yes that is correct, fixed.

pkg/wal/wal.go
kerneltime: Added a note.

ransomw1c:
approval pre-supposes Fred's comments re. yaml and type alias are to be addressed..

----

.. the `.md` port of the Confluence is a welcome addition.

my partial paraphrase is

> to ensure bundle descriptor upload and WAL entries occur according to the same timer, we rely on urls with expiration and GCS being attached to the same timer

----

here, the wal is landing as a coherent idea before attaching the wal to bundle descriptor uploads

PR #272, opened 2019-10-22
cli error handling consistency

ransomw1c:
addendum to #269 , to specifically address [this](https://github.com/oneconcern/datamon/pull/269#discussion_r335227339) comment about consistency


ransomw1c:
with some changes to the mocks setup in `cli_test.go`, all of `osExit`, `logFatalln`, and `logFatalf` could be removed for a slightly cleaner interface.  i'm okay with leaving these as tech debt in the wake of #269 with the changes here being not only the minimal addition necessary to preserve consistent patterns throughout `cmd/datamon/cmd` but also an improvement to readability since we now have `wrapFatal*(..., <human-readable message>, err)` in many places where there before there was only `logFatalln(err)`.

PR #273, opened 2019-10-25
draft golang sidecar parameterization api

ransomw1c:
this is a draft of the golang api portion of #270 ..

.. yaml deserialization is via public types and the yaml lib.

environment variable deserialization not (yet) added..

### DIFF
--- a/pkg/sidecar/param/pub_api.go
+++ b/pkg/sidecar/param/pub_api.go
@@ -1,0 +1,332 @@
+// Copyright © 2018 One Concern
+package param
+
+import (
+	"bytes"
+	"errors"
+	"io/ioutil"
+
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	CerialFmtYAM = iota
+)
+
+var defaultCerialFmt int = CerialFmtYAM
+
+type Cerializer interface {
+	FirstCutSidecarFmt(destPath string) error
+	CerialString(knownFmt int) (string, error)
+	// @fredbi , declare Stringer implementation?
+}
+
+type fuseParamsBundleParams struct {
+	Name         string `json:"name" yaml:"name"`
+	SrcPath      string `json:"srcPath" yaml:"srcPath"`
+	SrcRepo      string `json:"srcRepo" yaml:"srcRepo"`
+	SrcLabel     string `json:"srcLabel" yaml:"srcLabel"`
+	SrcBundle    string `json:"srcBundle" yaml:"srcBundle"`
+	DestPath     string `json:"destPath" yaml:"destPath"`
+	DestRepo     string `json:"destRepo" yaml:"destRepo"`
+	DestMessage  string `json:"destMessage" yaml:"destMessage"`
+	DestLabel    string `json:"destLabel" yaml:"destLabel"`
+	DestBundleID string `json:"destBundleID" yaml:"destBundleID"`
+	_            struct{}
+}
+
+type FUSEParams struct {
+	Globals struct {
+		SleepInsteadOfExit bool   `json:"sleepInsteadOfExit" yaml:"sleepInsteadOfExit"`
+		CoordPoint         string `json:"coordPoint" yaml:"coordPoint"`
+		Contributor        struct {
+			Name  string `json:"name" yaml:"name"`
+			Email string `json:"email" yaml:"email"`
+			_     struct{}
+		} `json:"contributor" yaml:"contributor"`
+		_ struct{}
+	} `json:"globalOpts" yaml:"globalOpts"`
+	Bundles []fuseParamsBundleParams `json:"bundles" yaml:"bundles"`
+	_       struct{}
+}
+
+func (fuseParams *FUSEParams) CerialString(knownFmt int) (string, error) {
+	if knownFmt != CerialFmtYAM {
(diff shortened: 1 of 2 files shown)


pkg/sidecar/param/pub_api.go
```diff
@@ -0,0 +1,238 @@
...
+	Name         string `json:"name" yaml:"name"`
+	SrcPath      string `json:"srcPath" yaml:"srcPath"`
+	SrcRepo      string `json:"srcRepo" yaml:"srcRepo"`
+	SrcLabel     string `json:"srcLabel" yaml:"srcLabel"`
+	SrcBundle    string `json:"srcBundle" yaml:"srcBundle"`
+	DestPath     string `json:"destPath" yaml:"destPath"`
+	DestRepo     string `json:"destRepo" yaml:"destRepo"`
+	DestMessage  string `json:"destMessage" yaml:"destMessage"`
+	DestLabel    string `json:"destLabel" yaml:"destLabel"`
+	DestBundleID string `json:"destBundleID" yaml:"destBundleID"`
```
jakedsouza: is `DestBundleID` a parameter passed by the client?

ransomw1c: could be better named ... this is a path to a file where the bundle id is written after upload.

the idea of such a specification was an as-needed addition when supporting the seismic pipeline integration with @amarkovits-1C ..
.. i recall, at the time, having some more abstracted plan of implementation that could then be retrofitted to the specific requirements of the seismic workflow.  i think it was something along the lines of having a unified artifact according to some format describing the results of the sidecar, just as we're now developing such an object for the parameterization?  i.e. totally RPC-ify sidecar usage

we implemented the simplest possible thing that fulfilled the seismic workflows requirements with the bundle id file to avoid a separate post-processing step.  and here i've again done the simple thing by preserving the extant interface, re-shaping it according to the parameterization rather than `optargs` as it was originally implemented.

if it's viewed as more useful to have a more general [un]marshallable output format defined in much the same way as the parameterization format is here described, i could implement that as well or instead of this parameter?

jakedsouza: Got it, I was a bit confused by `DestBundleID` because it seemed to be the bundleID and not a file path. It's not really required, but calling it DestBundleIDFile or something similar might make it more clear.

pkg/sidecar/param/pub_api.go
```diff
@@ -0,0 +1,238 @@
...
+		} `json:"contributor" yaml:"contributor"`
+		_ struct{}
+	} `json:"globalOpts" yaml:"globalOpts"`
+	Bundles []fuseParamsBundleParams `json:"bundles" yaml:"bundles"`
+	_       struct{}
+}
+
+type pgParamsDBParams struct {
+	Name        string `json:"name" yaml:"name"`
+	Port        int    `json:"pgPort" yaml:"pgPort"`
```
divya1c: db specs usually have a user too.

ransomw1c:
from discussion with @kerneltime , this merge request needs deserialization routines capable of writing YAML to disk as part of the public api to be considered viable.

additionally, shipping validation as a binary within the datamon sidecar as a helper for scripts that might write their own YAML such that a validation `initContainer` could be `chain`ed or `lift`ed into the _app init_ argo functionality, while not part of the acceptance criteria, will in all likelihood be a useful offering to the official surface.

pkg/sidecar/param/pub_api.go
ransomw1c: there's a default admin user, `postgres`.

"convention over configuration", "dry", and all the usual advantages.  customizability is the marked disadvantage.

see `mock_application.sh` comments.  i could also update the readme?

i see no immediate reason to bake perms into mvp, although it is possible to wind an additional parameter through, if necessary.

pkg/sidecar/param/pub_api.go
```diff
@@ -0,0 +1,238 @@
...
+	DestRepo    string `json:"destRepo" yaml:"destRepo"`
+	DestMessage string `json:"destMessage" yaml:"destMessage"`
+	DestLabel   string `json:"destLabel" yaml:"destLabel"`
+	SrcRepo     string `json:"srcRepo" yaml:"srcRepo"`
+	SrcLabel    string `json:"srcLabel" yaml:"srcLabel"`
+	SrcBundle   string `json:"srcBundle" yaml:"srcBundle"`
```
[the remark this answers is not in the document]
ransomw1c: ✨ 

yes, i didn't include it b/c i've considered the specificity of the bundle id file as useful albeit not useful enough.

i'd prefer doing away with the specificity of recording bundle ids to individual files store a single serialization format rather than lots of individual wires.

the 👍 of this towards consistency.

pkg/sidecar/param/pub_api_test.go
```diff
@@ -0,0 +1,158 @@
...
+	require.NoError(t, err, "serialize string")
+
+	expectedString := `globalOpts:
+  sleepInsteadOfExit: false
+  coordPoint: /tmp/coord
+  contributor:
+    name: contributor name
+    email: contributor@oneconcern.com
+bundles:
+- name: src
```
ransomw1c: absence of quotes indicates there's probably something awray

PR #274, opened 2019-10-28
feat(authentication): first stab at supporting google oauth

fredbi:
google oauth support for CLI

* at this moment, I am just warning CLI user of some deprecated --name/--email args. Please advise if you want to actually desupport those (will this support all use cases? I admit I am a bit lost given the possibility to change build flags etc)
* user name is no guaranteed from google userinf: using my own user for instance, I could see that this is not hydrated in google info. I set up a fall back on email as name, but we might want to make sure that users are better set up in google. Please advise.
* google oauth is an implementation of the new Authable interface, which at this moment, just gains access to the userinfo. We might want to augment that later on and make it a parameter. Now, this primarily serves the purpose of mocking auth for tests.
* I've added some minor fixes with logging output

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,18 @@
 
 # CLI Guide
 
-Make sure your gcloud credentials have been setup.
+Make sure your gcloud credentials have been setup, with proper scopes.
 ```$bash
-gcloud auth application-default login
+gcloud auth application-default login --scopes https://www.googleapis.com/auth/cloud-platform,email,profile
 ```
+
+datamon will use your email and name from your Google ID account.
+
+> **NOTE**: by default `gcloud auth application-default login` will not allow applications to see your full profile
+> In that case, datamon will use your email as your user name.
+>
+> You may control your personal information stored by Google here: https://aboutme.google.com
+
 Download the datamon binary for mac or for linux on the
 [Releases Page](https://github.com/oneconcern/datamon/releases/)
 or use the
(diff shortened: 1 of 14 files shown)


pkg/auth/google/auth.go
```diff
@@ -0,0 +1,67 @@
...
+// On linux, this is located at ~/.config/gcloud/application_default_credentials.json.
+func (g Auth) Principal(credFile string) (model.Contributor, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	svc, err := goauth.NewService(ctx,
+		goption.WithCredentialsFile(credFile),
+		goption.WithScopes(goauth.UserinfoEmailScope, goauth.UserinfoProfileScope),
+	)
+	if err != nil {
+		return model.Contributor{}, fmt.Errorf("invalid credentials: %w", err)
```
kerneltime: ```invalid credential``` => ```failed to use credentials```? It might be a network issue being reported as invalid creds.

pkg/auth/google/auth.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	u, err := svc.Userinfo.Get().Do()
+	if err != nil {
+		return model.Contributor{}, fmt.Errorf("cannot retrieve userinfo: %w", err)
+	}
+
+	return model.Contributor{
+		Email: u.Email,
```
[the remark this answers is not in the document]
fredbi: email is indeed a scope. It is granted by default when performing gcloud login

pkg/auth/google/auth.go
fredbi: ok will change this

fredbi: => forthcoming PR

README.md
```diff
@@ -2,10 +2,18 @@
 
 # CLI Guide
 
-Make sure your gcloud credentials have been setup.
+Make sure your gcloud credentials have been setup, with proper scopes.
 ```$bash
-gcloud auth application-default login
+gcloud auth application-default login --scopes https://www.googleapis.com/auth/cloud-platform,email,profile
```
ransomw1c: plus bien!

cmd/datamon/cmd/root.go
```diff
@@ -49,6 +51,9 @@ var logFatalln = log.Fatalln
 var logFatalf = log.Fatalf
 var osExit = os.Exit
 
+// used to patch over calls to Authable.Principal() during test
+var authorizer auth.Authable = gauth.New()
```
ransomw1c: since `New()` is arbitrary procedure, not patching of static declaration, i suggest declaration here and assignment in `init()`, `initConfig()`, or one of the other existing lifecycle stages such as those in `flags.go`

PR #275, opened 2019-10-28
Wrapping errors

fredbi:
This is an experimental PR attempting to wrap errors in a clean and concise way, 
based on the new go1.13 native error handling.
I am still thinking that the ` fmt.Errorf("%w", err)` hack is untidy...

I applied this to the new `auth` package to illustrate usage. Feedback welcome.

### DIFF
--- a/pkg/auth/google/auth.go
+++ b/pkg/auth/google/auth.go
@@ -2,9 +2,9 @@ package google
 
 import (
 	"context"
-	"fmt"
 	"time"
 
+	"github.com/oneconcern/datamon/pkg/auth/status"
 	"github.com/oneconcern/datamon/pkg/model"
 	goauth "google.golang.org/api/oauth2/v2"
 	goption "google.golang.org/api/option"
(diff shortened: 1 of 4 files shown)


kerneltime:
Go really messed up error handling.. any how this looks fine.

fredbi:
@kerneltime  Now rebased so you can get a clearer view of the actual changes re errors

ransomw1c:
i've gone down the roll-your-own `.cause()` path in python, and it has some advantages and disadvantages as well..

.. if the pattern is something everyone is genuinely on board with, i'm willing to give it a try, except can i suggest mangling the names a little more with respect to the experimental nature of the change?

fredbi:
@ransomw1c could you please be more specific about which names require mangling? I am not sure about what you mean here. Maybe an example would clarify.

fredbi:
@ransomw1c the `error.Cause()` pattern is actually not mine: it forms the basis of the new error handling that has been baked into go1.13 (previously you had to use `github.com/pkg/errors`, the
ideas from this repo have been included in the language).

Unfortunately, go1.13 hasn't provided an easy way to `Wrap` errors without resorting to string messages (which makes sentinel errors difficult to use). Hence this extension with a `Wrap`  method, thus avoiding the use of ` fmt.Sprintf("%w")`.

There are quite a few proposals to improve error handling in go2, but I can't see anything really good coming from there.

ransomw1c:
opinions about [dreaded word] "monadization" via a structure a-la
https://github.com/go-yaml/yaml/blob/v2.2.5/encode.go#L79
?

PR #281, opened 2019-11-05
unreviewed review

ransomw1c:
these are a few items i wasn't able to collaborate on, being away from the kinesis "freestyle 2" :/  keyboard, last week.

low priority in terms of sprint goals.

### DIFF
--- a/cmd/datamon/cmd/root.go
+++ b/cmd/datamon/cmd/root.go
@@ -52,7 +52,7 @@ var logFatalf = log.Fatalf
 var osExit = os.Exit
 
 // used to patch over calls to Authable.Principal() during test
-var authorizer auth.Authable = gauth.New()
+var authorizer auth.Authable
 
 // infoLogger wraps informative messages to os.Stdout without cluttering expected output in tests.
 // To be used instead on fmt.Printf(os.Stdout, ...)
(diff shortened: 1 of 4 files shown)


cmd/datamon/cmd/root.go
```diff
@@ -85,6 +85,7 @@ func Execute() {
 func init() {
 	log.SetFlags(0)
 	cobra.OnInitialize(initConfig)
+	authorizer = gauth.New()
```
fredbi: :+1:

pkg/auth/google/auth.go
```diff
@@ -37,10 +38,28 @@ func (g Auth) Principal(credFile string) (model.Contributor, error) {
 		return model.Contributor{}, fmt.Errorf("cannot create oauth service: %w", err)
 	}
 
-	u, err := svc.Userinfo.Get().Do()
+	var u *goauth.Userinfoplus
+	u, err = svc.Userinfo.Get().Do()
 	if err != nil {
 		return model.Contributor{}, fmt.Errorf("cannot retrieve userinfo: %w", err)
 	}
+	fullName := func() (name string) {
```
fredbi: Matter of taste and preferences, I won't comment on that one. No blocker.

PR #282, opened 2019-11-07
Datamon context

kerneltime:
1. Introduce the Context: Named bucket sets that uniquely identify a
datamon instance.
2. Introduce ContextStores: storage.Store interface implementation set
for a Context.
3. Introduce Configuration for CLI. A remote repository of contexts.
4. Change CLI to only depend on configuration bucket to detect a context
and use it.
5. Remove CSI plugin to reduce code. Can be added back from history when
ephemeral support lands in GKE for CSI.
6. Improve default param handling in CLI.


kerneltime:
Keeping datamon2 branch separate till we are ready to do beta release.

pkg/model/context.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"gopkg.in/yaml.v2"
+)
+
+const ContextVersion = 1.0
+
+// Context defines the details for a Datamon context.
+type Context struct {
+	Name      string `json:"name" yaml:"name"`           // Name for the context
+	WAL       string `json:"wal" yaml:"wal"`             // WAL is the location for the log
```
ransomw1c: upon explicit call for further review, i suppose these could be schematized as interfaces

PR #284, opened 2019-11-09
WIP Introduce runs

kerneltime:

### DIFF
--- a/pkg/model/run.go
+++ b/pkg/model/run.go
@@ -1,0 +1,63 @@
+/*
+ * Copyright © 2019 One Concern
+ *
+ */
+
+package model
+
+import (
+	"time"
+)
+
+type Run struct {
+	Type        string      `json:"type" yaml:"type"` // Type of run
+	Name        string      `json:"name" yaml:"name"` // Name for the run
+	ID          string      `json:"id" yaml:"id"`     // Unique id for an instance of a run. Each time the same Type+Name is run it results in a new ID.
+	Timestamp   time.Time   `json:"timestamp" yaml:"timestamp"`
+	Contributor Contributor `json:"contributor" yaml:"contributor"`
+	Steps       []Steps     `json:"steps" yaml:"steps"`
+}
+
+type FS struct {
+	Context     string `json:"context" yaml:"context"`
+	Repo        string `json:"repo" yaml:"repo"`
+	BundleID    string `json:"bundleid" yaml:"bundleid"`
+	Label       string `json:"label" yaml:"label"`
+	Fuse        bool   `json:"fuse" yaml:"fuse"`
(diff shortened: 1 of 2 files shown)


pkg/model/run.go
```diff
@@ -0,0 +1,63 @@
...
+	Filesystems []FS `json:"filesystems" yaml:"filesystems"`
+	Database    []DB `json:"database" yaml:"database"`
+}
+
+type Steps struct {
+	Containers []RunStepContainer `json:"containers" yaml:"containers"`
+}
+
+type RunStepContainer struct {
+	ContainerID string `json:"containerid" yaml:"containerid"`
```
galvare2: Will these ContainerID's be generated by the pipelines tool, or by Datamon?

pkg/model/run.go
```diff
@@ -0,0 +1,63 @@
...
+type DB struct {
+	Context  string `json:"context" yaml:"context"`
+	Repo     string `json:"repo" yaml:"repo"`
+	BundleID string `json:"bundleid" yaml:"bundleid"`
+	Label    string `json:"label" yaml:"label"`
+	Type     string `json:"type" yaml:"type"`
+	Port     int    `json:"port" yaml:"port"`
+}
+
+type DataSetIn struct {
```
galvare2: How are DataSetIn and DataSetOut going to be used?

pkg/model/run.go
```diff
@@ -0,0 +1,63 @@
...
+	"time"
+)
+
+type Run struct {
+	Type        string      `json:"type" yaml:"type"` // Type of run
+	Name        string      `json:"name" yaml:"name"` // Name for the run
+	ID          string      `json:"id" yaml:"id"`     // Unique id for an instance of a run. Each time the same Type+Name is run it results in a new ID.
+	Timestamp   time.Time   `json:"timestamp" yaml:"timestamp"`
+	Contributor Contributor `json:"contributor" yaml:"contributor"`
+	Steps       []Steps     `json:"steps" yaml:"steps"`
```
galvare2: Steps here are an array, meaning they are ordered linearly (totally ordered). What about runs with complex DAGs that include steps running in parallel that cannot be ordered linearly in this way?

pkg/model/run.go
kerneltime: Pipelines

pkg/model/run.go
kerneltime: This is used to capture the data after it has been processed, so the pipelines tool would need to pass this these attributes to the datamon side car (once).

pkg/model/run.go
kerneltime: Thanks, will include it.

PR #285, opened 2019-11-09
Fix paths to reflect deserialization choice.

kerneltime:
This is an artifact of initial churn in code. The deserialization should
match the file path.

### DIFF
--- a/cmd/metrics/cmd/mock_dest_store.go
+++ b/cmd/metrics/cmd/mock_dest_store.go
@@ -25,7 +25,7 @@ func (mds *mockDestStore) String() string {
 
 func (mds *mockDestStore) Has(ctx context.Context, key string) (bool, error) {
 	// detect and respond to RepoExists() call
-	if strings.HasPrefix(key, "repos/") && strings.HasSuffix(key, "/repo.json") {
+	if strings.HasPrefix(key, "repos/") && strings.HasSuffix(key, "/repo.yaml") {
 		return true, nil
 	}
 	return false, errors.New("mock destination store Has() unimpl (other than for RepoExists calls)")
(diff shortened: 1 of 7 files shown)


cmd/metrics/cmd/mock_dest_store.go
```diff
@@ -25,7 +25,7 @@ func (mds *mockDestStore) String() string {
 
 func (mds *mockDestStore) Has(ctx context.Context, key string) (bool, error) {
 	// detect and respond to RepoExists() call
-	if strings.HasPrefix(key, "repos/") && strings.HasSuffix(key, "/repo.json") {
+	if strings.HasPrefix(key, "repos/") && strings.HasSuffix(key, "/repo.yaml") {
```
fredbi: how about repos created with pre-v2 datamon? should they be migrated or do we want to had some compatibility path here?

cmd/metrics/cmd/mock_dest_store.go
```diff
@@ -50,7 +50,7 @@ func (mds *mockDestStore) Touch(ctx context.Context, objectName string) error {
 var fileListRe *regexp.Regexp
 
 func init() {
-	fileListRe = regexp.MustCompile(`bundle-files-\d+.json$`)
+	fileListRe = regexp.MustCompile(`bundle-files-\d+.yaml$`)
```
fredbi: I think we should be more tolerant here and accept json and support repos created with pre-V2.

cmd/metrics/cmd/mock_dest_store.go
kerneltime: We will move the data over. It will be separate buckets.

cmd/metrics/cmd/mock_dest_store.go
kerneltime: The old repos will be moved but no need to concurrently support both.

cmd/metrics/cmd/mock_dest_store.go
fredbi: ok as per our discussion

cmd/metrics/cmd/mock_dest_store.go
fredbi: ok as per our discussion

PR #286, opened 2019-11-11
Feat parallel repo list

fredbi:


pkg/core/repo_list_test.go
```diff
@@ -0,0 +1,170 @@
...
+			},
+		},
+		{
+			name:     happyWithBatches,
+			expected: expectedRepoBatchFixture,
+		},
+	}
+}
+
+func buildRepoYaml(repo string) string {
```
kerneltime: Why no use the model pkg? I am worried about silent failures as this format is not based on the actual deserialization.

cmd/datamon/cmd/bundle_list.go
```diff
@@ -37,8 +36,8 @@ var BundleListCommand = &cobra.Command{
 			return
 		}
 		err = core.ListBundlesApply(params.repo.RepoName, remoteStores.meta, applyBundleTemplate,
-			core.ConcurrentBundleList(params.core.ConcurrencyFactor),
-			core.BundleBatchSize(params.core.BatchSize))
+			core.ConcurrentList(params.core.ConcurrencyFactor),
```
kerneltime: in ```cmd/repo_list.go``` add ```addConcurrencyFactorFlag(repoList)```

pkg/core/repo_list_test.go
fredbi: Good idea. Have some issues with timestamps as yaml etc, though. Testing that right now

cmd/datamon/cmd/bundle_list.go
fredbi: yep forgot that. True

PR #297, opened 2019-11-19
Add support for capturing the deduplication scheme.

kerneltime:
Currently there is a single deduplication scheme but we should add
support to list the deduplication scheme in preparation for adding
other schemes.

### DIFF
--- a/pkg/cafs/cafs.go
+++ b/pkg/cafs/cafs.go
@@ -17,8 +17,9 @@ import (
 )
 
 const (
-	DefaultLeafSize = 2 * 1024 * 1024
-	MaxLeafSize     = 5 * 1024 * 1024
+	DefaultLeafSize    = 2 * 1024 * 1024
+	MaxLeafSize        = 5 * 1024 * 1024
+	DeduplicationBlake = "blake"
 )
 
 // LeafSize configuration for the blake2b hashes
@@ -98,6 +99,7 @@ type Fs interface {
 	Keys(context.Context) ([]Key, error)
 	RootKeys(context.Context) ([]Key, error)
 	Has(context.Context, Key, ...HasOption) (bool, []Key, error)
+	GetAddressingScheme() string
 }
 
 // New creates a new file system operations instance for a repository
@@ -108,6 +110,7 @@ func New(opts ...Option) (Fs, error) {
 		leafSize:                    uint32(5 * units.MiB),
 		concurrentFlushes:           10,
 		readerConcurrentChunkWrites: 3,
+		deduplicationScheme:         DeduplicationBlake,
 	}
 	f.leafPool = newLeafFreelist()
 	f.lru, _ = lru.NewWithEvict(10, func(lruKey interface{}, lruVal interface{}) {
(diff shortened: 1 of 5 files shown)


pkg/cafs/cafs.go
```diff
@@ -17,8 +17,9 @@ import (
 )
 
 const (
-	DefaultLeafSize = 2 * 1024 * 1024
-	MaxLeafSize     = 5 * 1024 * 1024
+	DefaultLeafSize     = 2 * 1024 * 1024
+	MaxLeafSize         = 5 * 1024 * 1024
+	Deduplication_blake = "blake"
```
fredbi: We should not snake-case go identifiers
```suggestion
	DeduplicationBlake = "blake"
```

pkg/cafs/cafs.go
```diff
@@ -108,6 +110,7 @@ func New(opts ...Option) (Fs, error) {
 		leafSize:                    uint32(5 * units.MiB),
 		concurrentFlushes:           10,
 		readerConcurrentChunkWrites: 3,
+		deduplicationScheme:         Deduplication_blake,
```
fredbi:
```suggestion
		deduplicationScheme:         DeduplicationBlake,
```

pkg/cafs/cafs.go
```diff
@@ -139,6 +142,11 @@ type defaultFs struct {
 	leafPool                    *leafFreelist
 	concurrentFlushes           int
 	readerConcurrentChunkWrites int
+	deduplicationScheme         string
+}
+
+func (d *defaultFs) GetAddressingScheme() string {
+	return Deduplication_blake
```
fredbi:
```suggestion
	return DeduplicationBlake
```

pkg/core/bundle.go
```diff
@@ -96,6 +96,7 @@ func NewBDescriptor(descriptorOps ...BundleDescriptorOption) *model.BundleDescri
 		Contributors:           nil,
 		BundleEntriesFileCount: 0,
 		Version:                model.CurrentBundleVersion,
+		Deduplication:          cafs.Deduplication_blake,
```
fredbi:
```suggestion
		Deduplication:          cafs.DeduplicationBlake,
```

pkg/core/bundle_pack.go
```diff
@@ -313,8 +314,18 @@ func uploadBundle(ctx context.Context, bundle *Bundle, bundleEntriesPerFile uint
 	return nil
 }
 
-func uploadBundleDescriptor(ctx context.Context, bundle *Bundle) error {
+func validateBundle(bundle *Bundle) bool {
+	if bundle.BundleDescriptor.Deduplication == "" {
```
fredbi: this is breaking again against existing stuff, without obvious necessity. Why don't we assume Blake is the default?

pkg/core/bundle.go
fredbi: Also to be consistent we should add the associated builder option

```
func Deduplication(d string) BundleDescriptorOption {
	return func(b *model.BundleDescriptor) {
		b.Deduplication = d
	}
}
```

pkg/core/bundle_pack.go
kerneltime: This would be breaking if the master was to be used against existing stuff. This is to insure that future code paths do not forget to set it.

fredbi:
@kerneltime I think that this gomnd linter has just had a short lifespan in golangci: it looks it is no more in here. Re-running the workflow should be ok.
```
: mnd: Magic number: 5, in <argument> detected (gomnd)
```

PR #298, opened 2019-11-19
Change label path to be similar to repo and bundle.

kerneltime:
Change how labels are written to object stores.

### DIFF
--- a/pkg/core/label_list.go
+++ b/pkg/core/label_list.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	typicalLabelsNum = 1000 // default number of allocated slots for repos
+	typicalLabelsNum = 1000 // default number of allocated slots for labels
 )
 
 // ListLabels returns all labels from a repo
--- a/pkg/model/label.go
+++ b/pkg/model/label.go
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"strings"
 	"time"
+	"unicode"
+
+	"github.com/badoux/checkmail"
 )
 
 // LabelDescriptor describes a label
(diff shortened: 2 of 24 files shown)


pkg/model/label.go
```diff
@@ -39,12 +42,41 @@ func GetArchivePathPrefixToLabels(repo string, prefixes ...string) string {
...
+		}
+	}
+	for _, c := range label.Contributors {
+		if c.Name == "" {
+			return fmt.Errorf("name for contributor cannot be blank")
+		}
+		if c.Email == "" {
+			return fmt.Errorf("email for contributor cannot be blank")
+		}
+		// TODO: Is this sufficient validation?
```
fredbi: You may also use the more standard https://golang.org/pkg/net/mail/#AddressParser.Parse

pkg/core/label_list.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	typicalLabelsNum = 1000 // default number of allocated slots for repos
+	typicalLabelsNum = 1000 // default number of allocated slots for labels
```
fredbi: :1st_place_medal:

pkg/model/label.go
fredbi: Or skip this validation altogether, since email is assumed trusted from auth.

pkg/model/label.go
```diff
@@ -39,12 +42,41 @@ func GetArchivePathPrefixToLabels(repo string, prefixes ...string) string {
 	return fmt.Sprint(getArchivePathToLabels(), repo+"/"+strings.Join(prefixes, "/"))
```
fredbi: this comment for more explanations: how are we supposed now to carry on with prefix-based search on labels?

kerneltime: I am not sure I understand this comment. The prefix should work as expected. Ref: https://github.com/oneconcern/datamon/blob/labels/cmd/datamon/cmd/cli_test.go#L609

fredbi: ok

kerneltime:
@fredbi the bundle_list_test.go needs refactoring (not part of this test) to implement more stringent checks on path validation. (model/bundle.go)

PR #299, opened 2019-11-19
godoc documentation

fredbi:
Beginning of some long standing effort to:
* fully document the datamon API and packages
* reduce the surface of the exposed interface and package level exported variables

* doc: added some godoc-friendly comments (pkg/core, pkg/model, pkg/context)
* refactor(core): refactored core mem profiling to work with options instead of package level constant
* refactor(model): moved unsafe pointer helpers to internal package
* refactor(model): moved regexp compilation to init

### DIFF
--- a/pkg/core/bundle.go
+++ b/pkg/core/bundle.go
@@ -21,9 +21,10 @@ import (
 	"github.com/oneconcern/datamon/pkg/storage"
 )
 
-var MemProfDir string
-
-// Bundle represents the bundle in its archived state
+// Bundle represents a bundle in its archived state.
+//
+// A bundle is a point in time read-only view of a rep:branch and is composed
+// of individual files. Analogous to a commit in git.
 type Bundle struct {
 	RepoID                      string
 	BundleID                    string
@@ -46,7 +47,7 @@ func (b *Bundle) setBundleID(id string) {
 	b.BundleDescriptor.ID = id
 }
 
-// InitializeBundleID create and set a new bundle ID
+// InitializeBundleID creates and sets a new bundle ID
 func (b *Bundle) InitializeBundleID() error {
 	id, err := ksuid.NewRandom()
 	if err != nil {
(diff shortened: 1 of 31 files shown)


pkg/web/web-packr.go
```diff
@@ -0,0 +1,8 @@
+// +build !skippackr
```
kerneltime: Was this expected to be part of the commit?

pkg/web/packrd/packed-packr.go
```diff
@@ -0,0 +1,48 @@
+// +build !skippackr
```
kerneltime: Was this expected to be part of the commit?

pkg/web/web-packr.go
fredbi: nope. I am fixing this.

pkg/web/packrd/packed-packr.go
fredbi: blunder! good catch. Rebasing now

pkg/core/bundle.go
```diff
@@ -293,24 +220,24 @@ func implPublishMetadata(ctx context.Context, bundle *Bundle,
 }
 
 // Upload an bundle to archive
-func Upload(ctx context.Context, bundle *Bundle) error {
-	return implUpload(ctx, bundle, defaultBundleEntriesPerFile, nil)
+func Upload(ctx context.Context, bundle *Bundle, opts ...ListOption) error {
```
ransomw1c: with `ListOption` now documented as options pertaining to list operations (`bundle list`, `repo list`, etc.), using `ListOption`s to `Upload` is not something we'll want to ship as part of a stable api..

..i dunno if it's worth worrying about as part of this merge request or not.  moreover, i don't know if the change should be as shallow as using a different name from`ListOption` to describe the same `func(*Settings)` type or whether splitting hairs at the level of structures containing different types of options appears worthwhile.

the only strong opinion here is to have the documentation agree with actual usage by the time (if ever) we expect developers to use the `core` api independently of the datamon executable.

fredbi: This was added to avoid a global package level export symbol to enable memory profiling, nothing more

pkg/model/bundle.go
```diff
@@ -211,14 +216,21 @@ func GetArchivePathComponents(archivePath string) (ArchivePathComponents, error)
...
 func IsGeneratedFile(file string) bool {
 	// TODO: Need to find a way for AeroFs to convert to abs patch while honoring the fake root
 	//path, err := filepath.Abs(file)
-	match, _ := regexp.MatchString("^.datamon/*|^/.datamon/*|^/.datamon$|^.datamon$|^./.datamon/*|^./.datamon$", file)
-	return match
+	return genFileRe.MatchString(file)
+}
+
+func init() {
+	metaRe = regexp.MustCompile(`^\.datamon/(.*)\.yaml$`)
```
ransomw1c: this is a good catch for performance and catching errors early on.  i'd feel slightly better if all the regular expressions were attached to a structure so that the global names aren't too numerous.. yet another nitpick to be resolved time permitting.

pkg/core/bundle.go
fredbi: I think `ListOption` will be renamed to simply `Option` in the near future

ransomw1c:
keep `internal/prof.go` in mind when considering the overall metrics collection and logging setup used to seek out memory usage errors.

pkg/core/bundle.go
ransomw1c: > I think `ListOption` will be renamed to simply `Option` in the near future

i'd be alright with this.

fredbi:
> 
> 
> overall, everything appears to be in order. i'm all for changes to improve understanding.
> 
> if i were to pick one final change for this diff, it would be the use of `ListOptions` for `Upload` and other operations that aren't about listing, although i'm equally content to get it resolved in later diffs.

ok. Thanks. A few closing remarks:
* The memprofile stuff was already here: I just made it private/driven by options. We have quite a big undertaking ahead improving tracing/logging and overall the ability to monitor things
* `ListOption` was not renamed in that move since it would have incurred to many changes for an easy review. Suggesting to rename it as `Option` for next round of refactoring

PR #300, opened 2019-11-19
refactored Makefile and CI workflows

fredbi:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,11 @@
-version: 2
-jobs:
-  go_lint:
+version: 2.1
+
+executors:
+  docker_minio:
     working_directory: ~/project
+    environment:
+      GOOGLE_PROJECT_ID: "onec-co"
+      GOOGLE_COMPUTE_ZONE: us-west2-c
     docker:
       - image: circleci/golang
       - image: minio/minio
@@ -14,21 +18,29 @@ jobs:
         command:
           - server
           - data
+  docker_builder:
+    working_directory: ~/project
+    docker:
+      - image: circleci/golang
+
+commands:
+  install_base:
     steps:
       - run:
-          name: Install packages
+          name: Install packages with gcloud SDK installed locally
           command: |
-            sudo apt-get update -y
-            sudo apt-get install lsb-release
+            sudo apt-get update -y --quiet
+            sudo apt-get install --quiet lsb-release
             export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)"
             echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
             curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add
-            sudo apt-get update -y
-            sudo apt-get install -y git build-essential
-            sudo apt-get install -y google-cloud-sdk
-            sudo apt-get install -y shellcheck
+            sudo apt-get update -y --quiet
+            sudo apt-get install -y --quiet git build-essential google-cloud-sdk shellcheck zsh
+
+  install_tools:
+    steps:
       - run:
-          name: Install go tools
+          name: Install go tools for testing
           environment:
             GO111MODULE: 'on'
           command: |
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -14,37 +18,64 @@ jobs:
...
+            go get -u github.com/matryer/moq@latest
+            go get -u github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+            go clean -modcache
+
+  login_to_google:
+    steps:
+      - run:
+          name: Initialize gcloud SDK
+          command: |
+            echo $GCLOUD_SERVICE_KEY | gcloud auth activate-service-account --key-file=-
```
ransomw1c: note to self: this environment variable is [set through the circleci web ui](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project)

.circleci/config.yml
```diff
@@ -54,52 +85,30 @@ jobs:
...
             - pkg-cache-{{ checksum "go.sum" }}
             - pkg-cache-
+      - install_tools
+      - login_to_google
       - run:
           name: Run golang tests
           command: |
             mkdir -p ~/extra
+            mkdir -p /tmp/test-results
             echo $GCLOUD_SERVICE_KEY > ~/extra/appcredentials.json
```
ransomw1c: if i'm reading [this](https://github.com/googleapis/google-api-go-client/blob/master/option/option.go#L49) correctly, we'll probably want to use the `.config/gcloud/*` credentials files created by `login_to_google` rather than authenticate with the (currently synonymous) service account credentials in the not-too-distant future?

fredbi: To be tested before confirmation. I believe there is an extra step using `gcloud auth etc...` to actually build the appropriate credential files. At this moment, this just works. Will carry out more testing to gain better insight of this gcloud process when setting up auth for datamon with google workload identity

Makefile
```diff
@@ -48,133 +51,93 @@ help:
...
+	.
+
+.PHONY: push-target
+push-target:
+	docker push $(BUILD_TARGET)
+
+.PHONY: build-datamon-binaries
+build-datamon-binaries: BUILD_TARGET=datamon-binaries
+build-datamon-binaries:
+	@make build-target \
```
ransomw1c: had to look up [the manual](https://www.gnu.org/software/make/manual/html_node/Recursion.html) on this.

i am not a make wizard, and i do not want to be a make wizard (`mk` being a different matter).
anyway, all opinions aside, `make` (in all its flavors, except perhaps automake) is worthwhile to know something about, so i've some readings.

specific gleanings from the manual appear to be

* [`$(MAKE)`](https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable) instead of `@make`
* [export](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html#Variables_002fRecursion) as in `export DOCKERFILE = binaries.Dockerfile` as its own statement preceding the recursive use via `$(MAKE)` rather than [overriding](https://www.gnu.org/software/make/manual/html_node/Overriding.html#Overriding) via shell-like `make` invocation

.circleci/config.yml
ransomw1c: > At this moment, this just works

understood.

sidecar.Dockerfile
```diff
@@ -6,35 +6,25 @@ ADD ./hack/fuse-demo/datamon.yaml /root/.datamon/datamon.yaml
...
+RUN apt-get update --quiet && apt-get install -y --quiet --no-install-recommends fuse sudo vim zsh &&\
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* &&\
+    echo "allow_root" >> /etc/fuse.conf
 
 COPY --from=base /stage /
 ENV ZONEINFO /zoneinfo.zip
 
-ADD ./hack/fuse-demo/datamon.yaml /root/.datamon/datamon.yaml
-
+ADD hack/fuse-demo/datamon.yaml /root/.datamon/datamon.yaml
```
kerneltime: We can delete the credential details in this file.

fredbi: @kerneltime let's make some progress on the kubernetes credentials beforehand.
I gather that there are some unforeseen changes ahead.

Makefile
fredbi: @ransomw1c done for `$(MAKE)`.
Regarding export I more nuanced. I've tested more following your recommendation and this changed a lot the behavior. I've now added some `export` directives for almost all targets, but the one that is precisely cascading, for which I need an override.

PR #301, opened 2019-11-20
Reorg deploy

fredbi:
Cleaning up some ancillary directories and starting to reduce the mass of shell/hacks

This requires a new repo `homebrew-datamon`  to be created, and the recipe to be built there.

- documented the associated simplified installation for mac users (or Linux brew users)
- removed install script for mac users
- move the minio k8 manifest to some better suited place (is that used?)

NOTE: there is a lot of functionality here, since we no more care about automatically checking for updates. If we want to do that, I suggest we do this from the binary itself, e.g. using something like https://github.com/rhysd/go-github-selfupdate


fredbi:
Now onboarded on heavier change with goreleaser etc.

PR #302, opened 2019-11-20
Makefile refactoring: replaced release_tag script by golang (no more …

fredbi:
…zsh dependent)


kerneltime:
I am guess this will get tested as part of the next release?

PR #303, opened 2019-11-20
experiment parallel testing with circleci

fredbi:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,6 +88,7 @@ jobs:
 
   go_test:
     executor: docker_minio
+    parallelism: 4
     environment:
       GOOGLE_PROJECT_ID: "onec-co"
       GOOGLE_COMPUTE_ZONE: us-west2-c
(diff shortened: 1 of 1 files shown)


fredbi:
@kerneltime here is an interesting failure that I often reproduce locally (without --race).
I am assuming there is some kind of race condition in the CLI tests, resulting in a deadlock.
https://circleci.com/gh/oneconcern/datamon/2294

kerneltime:
Yes the CLI tests are dependent. They are closer to integration tests than unit tests, we can add build tag for them and run other tests in parallel?

fredbi:
@kerneltime I'd be glad to parallelize this. Further, the fact that I am often running into issues locally may stem from non-deterministic scheduling of tests. If you could explain the dependency logic, we could secure the test suite, by wrapping all required steps as non-exported test funcs.

fredbi:
@kerneltime I believe this is a lock between some tests running in `pkg` and some other running in `cmd`. Do we refer to the same buckets?

kerneltime:
> @kerneltime I believe this is a lock between some tests running in `pkg` and some other running in `cmd`. Do we refer to the same buckets?

I don't think so, will look into this. There should not be any shared buckets between tests. I know the CLI tests have some sequence of operations it performs but it should be good to make them parallelized.

fredbi:
stopping this line for the moment, trying new ideas following ransom's suggestions.

PR #304, opened 2019-11-21
doc(sidecar): more doc on sidecar

fredbi:


ransomw1c:
🎇 to [documentation created and maintained via similar processes used for source code](https://www.writethedocs.org/) ... could try [asciidoc](http://asciidoc.org/userguide.html) or similar to format the flow charts and other diagrams later on

PR #305, opened 2019-11-25
Feat goreleaser

fredbi:
~~This is based on #300 ~~ Rebased.


ransomw1c:
will review after #300 land and rebase

fredbi:
> 
> 
> will review after #300 land and rebase
> 
> if this branch is rebased onto #302 , would `hack_release_tag.go` be used anywhere other than the Makefile?

@ransomw1c no this is the sole purpose of this program. If you want my say, this could disappear altogether when we have a less complex model for tagging docker images.

ransomw1c:
> this [`hack_release_tag.go`] could disappear altogether when we have a less complex model for tagging docker images.

that'd be great.  as is, i'm not considering #302 particularly helpful

.circleci/config.yml
```diff
@@ -41,6 +44,8 @@ commands:
     steps:
       - run:
           name: Install go tools for testing
+          # update tools to latest release but clear the module cache after install
+          # to avoid conflicts with the main go dependencies.
```
ransomw1c: `main go dependencies` are those in `go.mod`?

is it possible to change `go.mod` to install `gotestsum`, `moq`, and so on?

i'm viewing this as a desirable change based on the preference to maintain the linter rc instead of params passed to the linter in the circle-ci yaml.  the linter rc is conceptually equivalent to `go.mod` in this comparison wherein we're attempting to shrink `.circleci/config.yaml` in favor of project-global (i.e. not just ci) setup.

.circleci/config.yml
```diff
@@ -188,13 +193,29 @@ jobs:
...
+            chmod u+x goreleaser
+            release_notes="./notes/NOTES.${CIRCLE_TAG}.md"
+            if [[ -n ${CIRCLE_TAG} && -f ${RELEASE_NOTES} ]] ; then
+              # use custom release notes
+              echo "Adding release notes ${release_notes}"
+              opts="--release-notes ${release_notes}"
+              # otherwise, leaves standard changelog from goreleaser
+              # (picks up all commits)
+            fi
+            ./goreleaser ${opts-""}
```
ransomw1c: although the `get()` with default metaphor is a useful placeholder to mention vis. shell scripting, `$opts` or `${opts}` is equivalent here since the default is `""` i.e. empty.

.circleci/config.yml
```diff
@@ -188,13 +193,29 @@ jobs:
...
+    # > goreleaser needs access to github API, not only the git repo.
+    # > Hence ssh key is not sufficient and GITHUB_TOKEN is required.
+    steps:
+      - run:
+          name: Prepare github release
+          command: |
+            # self-installing goreleaser
+            curl -sL https://git.io/goreleaser > goreleaser
+            chmod u+x goreleaser
+            release_notes="./notes/NOTES.${CIRCLE_TAG}.md"
```
ransomw1c: add a placeholder `notes/NOTES.${CIRCLE_TAG}.md` where `CIRCLE_TAG=master` to test?

ransomw1c: note to self: `CIRCLE_TAG` is the pattern matched by the filter in the workflow filter

Makefile
```diff
@@ -290,3 +289,15 @@ pg-demo-coord-build-app: export TAGS=latest
 pg-demo-coord-build-app:
 	$(MAKE) build-target
 	$(MAKE) push-target
+
+.PHONY: release
+## Cut a new release. Example: make release TAG=v1.0.6
+release:
```
ransomw1c: it would make sense to add a 

```if [ -f ./notes/NOTES.$(TAG).md ]; then 1>&2 echo 'release notes not found' && exit 1; fi```

or at least

```if [ -f ./notes/NOTES.$(TAG).md ]; then 1>&2 echo "${YELLOW}release notes not found${RESET}" && echo 'enter to continue' && read; fi```

to this target to warn early?

fredbi: generating release notes / changelog remains a todo at this moment. I wouldn't make it a requirement. I've just prepared the stage in CI to push custom release notes.

.circleci/config.yml
fredbi: don't see how making it a branch would help. I tested with a fork, but without CI.

.circleci/config.yml
fredbi: if managing dependencies is an issue, let's solve it separetely

fredbi:
>    this [hack_release_tag.go] could disappear altogether when we have a less complex model for > > tagging docker images.

> that'd be great. as is, i'm not considering #302 particularly helpful

i am not considering the smarter-than-life tagging scheme particularly helpful either. That's why, to even things, I find it best to just tag images with git tag + latest full stop.

Makefile
ransomw1c: whether it's a requirement or not, i'm okay with the decision.

my take is that it's a nice feature that'd be easy to forget about, so i'd prefer to get a difficult-to-miss warning in case release notes aren't found.

.circleci/config.yml
ransomw1c: it would help to test the ci.

PR #308, opened 2019-11-26
fix(usability): do not require config to be set for commands that don…

fredbi:
…'t need it

ex: `datamon version` should not require config env or file.


ransomw1c:
this merge request says that datamon needn't have a config file to init.

it says that it needs a config file only if it's going to connect to remote stores.

i suppose that there could be commands that need a config file and don't connect to remote stores.

fredbi:
> this merge request says that datamon needn't have a config file to init.

> it says that it needs a config file only if it's going to connect to remote stores.

> i suppose that there could be commands that need a config file and don't connect to remote stores.

What I am saying is, in fact, that config is not needed for _every_ command. I do not assume anything about connecting to remote store or doing something else with the config.
What this PR says is that, any command which needs the config runs through a `PreRun` step to populate the config. This is not a new thing, it has always been there. The new thing is that if you are calling `populateConfig...`  then CLI settings for config are required.

ransomw1c:
> any command which needs the config runs through a `PreRun` step to populate the config. This is not a new thing, it has always been there

yes, and i'm supplying the additional information that `context.Descriptor`, the config that's populated, is only used in `paramsToDatamonContext`... this could be more clear since the `context.Descriptor.Name` is set from the viper config and then overwritten by the unmarshalled context data from gcs.

ransomw1c:
https://github.com/oneconcern/datamon/tree/d1912011010-configparsecleanup--wip

contains some ideas for maintainability of the source as well as enabling the desired functionality from the cli.

i'll either open up a separate request based on those changes or suppose we iterate on this diff with some of these corresponding changes to take some care about the existing `PreRun` setup:  it is not altogether clear and moreover some commands are supposed to use `populateConfig...` under the existing setup that do not, in fact, need to call the function.

ransomw1c:
please lmk whether any of the changes in #313 are to be included here of if i should approve as-is

fredbi:
This one is very short and focused on having simple commands like `datamon version`, `datamon upgrade` to run without any special environment. That's it. #313 seems to have a different focus.
The current status is just a rebase of the original PR, before you issued #313. For me these are different matters

PR #313, opened 2019-12-02
context cmd cleanup

ransomw1c:
motivated by #308


cmd/datamon/cmd/root.go
```diff
@@ -110,7 +110,8 @@ func initConfig() {
 	// `viper.ConfigFileUsed()` returns path to config file if error is nil
 	var err error
 	// 2. Initialize config
-	config, err = newConfig()
+	config := new(CLIConfig)
```
fredbi: I believe we should now be able to remove the global `config` var, line 45

cmd/datamon/cmd/config.go
```diff
@@ -19,31 +21,23 @@ type CLIConfig struct {
...
-	return &config, nil
-}
-
-func (c *CLIConfig) setDatamonParams(flags *flagsT) {
-	if flags.context.Descriptor.Name == "" {
-		flags.context.Descriptor.Name = c.Context
-	}
+func populateRemoteConfig() {
+	var flags *flagsT
+	flags = &datamonFlags
```
fredbi:
```
flags := &datamonFlags
```

cmd/datamon/cmd/context_create.go
```diff
@@ -27,9 +27,11 @@ func createContext() {
 	if err != nil {
```
fredbi: I am assuming we should populate config here too (cf. panic in unit testing)

cmd/datamon/cmd/config.go
ransomw1c: this is an ongoing stylistic discussion:  whether to hindley-milner or not to hindley-milner.

from previous conversations with ritesh, i've variously considered type inference super and harmful.

i do not have a thorughgoing, apply-everywhere stance, except i do recall that i wanted explicit declaration by type separate from assignment here for some reason.  i think simple b/c i found it difficult to remember, in whatever context of a mental model i have for the `cmd/` package environment what with the rapid proliferation of global non-function variables, the type assumptions on `datamonFlags` at this particular part of the codebase, and i actually wanted the reminder.

ransomw1c:
this is a draft.  it's open for review yet not in the same state of readiness for review as, say, #314 .

cmd/datamon/cmd/config.go
fredbi: ok. Use // nolint hint then

ransomw1c:
closing this down for now.. please do not delete branch

PR #310, opened 2019-11-27
context get

ransomw1c:
implement `context get` similar to that of bundles and repos in order to implement fail-fast on non-exist context when plumbing parameter thru sidecar in #306

PR #314, opened 2019-12-02
fuse sidecar context parameterization

ransomw1c:
#305 and #310 together with

* plumbing to connect context bucket and context name to yaml
* demo/test script ported to ci


fredbi:
@ransomw1c could we merge #305 then, so we can more clearly see changes here?
I am assuming that producing things based on #305 is a kind of implicit approval for it, isn't it?

ransomw1c:
no, #305 is not a prerequisite for this review.

PR #315, opened 2019-12-04
At rest schema for runs.

kerneltime:
Take 2 for runs.

This adds the at rest schema for runs and bundles.


pkg/model/bundle.go
```diff
@@ -25,6 +30,7 @@ type BundleDescriptor struct {
 	BundleEntriesFileCount uint64        `json:"count" yaml:"count"`                                     // Number of files which have BundleDescriptor Entries
 	Version                uint64        `json:"version,omitempty" yaml:"version,omitempty"`             // Version for the bundle
 	Deduplication          string        `json:"deduplication,omitempty" yaml:"deduplication,omitempty"` // Type of deduplication used
+	RunStage               string        `json:"runstage,omitempty" yaml:"runstage,omitempty"`           // Path to the run stage
```
ransomw1c: what is a `RunStage`?

i see that there's also a run stage id as part of an input dataset, yet the run itself might contain data about the semantics of stages only from the `RunFile`, as far as i can tell?

overall, i'm still confused about the meaning of "stages."  it is clear that "run" is a novel, unitary concept.  i'm wanting stages to connect to other internal concepts about runs, and i don't see a clear connection encoded anywhere.

like why is a run made of "stages" rather than "levels?"  what does a stage consist of other than a place within a run that data is used?  wouldn't it be more standard to have a dedicated `RunStage` struct entity type that links here and within `model/run.go` via a unique id?

pkg/model/run.go
```diff
@@ -0,0 +1,106 @@
...
+type Category struct {
+	Name        string      `json:"name" yaml:"name"`               // Unique name for the category
+	Message     string      `json:"message" yaml:"message"`         // Message describing the category.
+	Contributor Contributor `json:"contributor" yaml:"contributor"` // Credentials for who created the category.
+	Version     int         `json:"version" yaml:"version"`         // Version of the run Schema.
+}
+
+type Run struct {
+	Name           string      `json:"name" yaml:"name"`                     // Name for the run
+	InstanceID     string      `json:"instanceid" yaml:"instanceid"`         // External ID used for a run. Not guaranteed to be unique.
```
galvare2: This is set by the pipeline?

pkg/model/run.go
```diff
@@ -0,0 +1,106 @@
...
+	RunID          string      `json:"id" yaml:"id"`                         // Internally generated Unique id for an instance of a run. Each time the same Type+Name is run it results in a new ID.
+	Timestamp      time.Time   `json:"timestamp" yaml:"timestamp"`           // Timestamp for when the run was created.
+	Contributor    Contributor `json:"contributor" yaml:"contributor"`       // Contributor who created the run.
+	RunFile        string      `json:"runfile" yaml:"runfile"`               // Raw file describing the run. Example: A yaml file.
```
galvare2: Would this be the argo workflow?

pkg/model/run.go
```diff
@@ -0,0 +1,106 @@
...
+	Version        int         `json:"version" yaml:"version"`               // Version of the run Schema.
+	PipelineEngine string      `json:"pipelineengine" yaml:"pipelineengine"` // The entity that executes the run.
```
galvare2: Could you give an example of what a PipelineEngine might be?

pkg/model/run.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+type Container struct {
+	Image   string `json:"image" yaml:"image"`     // Image name
+	ImageID string `json:"imageid" yaml:"imageid"` // Image ID
+}
+
+type DataSetIn struct {
+	Timestamp   time.Time `json:"timestamp" yaml:"timestamp"`     // Timestamp for the dataset in
+	Stage       string    `json:"stage" yaml:"stage"`             // Stage in which the dataset is used
```
galvare2: Would the stage be the name of the individual step in the pipeline that the data is being used for?

galvare2: Such as "DelftModelCleanup" or something like that?

pkg/model/run.go
kerneltime: Pipeline may have a notion of a unique ID that they would like to preserve.

pkg/model/run.go
kerneltime: For now yes.

pkg/model/run.go
kerneltime: This would be argo in the initial implementation but could be a Kubernetes job as well.

pkg/model/run.go
kerneltime: Yes. I will add more documentation.

pkg/model/bundle.go
kerneltime: It is important to define the concepts beyond the commends in the structs. Will do so.

pkg/model/run.go
galvare2: So it's a long string whose contents is the entire contents of the yaml?

ransomw1c: my understanding of `RunFile` is that it, together with `PipelineEngine` form a sort of [tagged union](https://en.wikipedia.org/wiki/Tagged_union) within the `Run` struct.

i kinda recommend making an interface type

```
interface RunEngine (
  RunFile() string
  PipelineEngine() string
)
```

and including a concrete type implementing only this interface in `Run` -- i.e.
```
type Run struct {
...
  Engine RunEngine
...
}
```
this'd allow inquiring `Run.Engine.(PipelinesTool)` or `Run.Engine.(Argo)`

kerneltime: @galvare2 yes it would be the string content generated by the pipeline tool

kerneltime: @ransomw1c for now splitting the runengine out as a type. The interface can be added later

PR #318, opened 2019-12-05
fuse sidecar yaml params

ransomw1c:


ransomw1c:
resurrecting #270 for fuse-only with context debug and ci tests in-place

ransomw1c:
going to add building yaml from golang as part of the demo and rm the wip tag..

hack/k8s/example-coord.template.yaml
```diff
@@ -69,8 +72,14 @@ spec:
         - mountPath: /tmp/gac
           name: google-application-credentials
         env:
+
         - name: GOOGLE_APPLICATION_CREDENTIALS
           value: /tmp/gac/google-application-credentials.json
+
+        - name: dm_fuse_params
+          value: /fuse-params.yaml
```
galvare2: Where does this file come from in this example? Is it built into the datamon sidecar container?

hack/fuse-demo/wrap_datamon.sh
```diff
@@ -20,113 +31,236 @@ POLL_INTERVAL=1 # sec
...
+    env_vars_file=/tmp/env_vars_from_params.sh
+    if [[ -f $dm_fuse_params ]]; then
+        dbg_print "parsing parameters YAML file $dm_fuse_params"
+        dm_fuse_params_val=$(cat $dm_fuse_params)
+    else
+        dbg_print "parsing parameters YAML from env var"
+        dm_fuse_params_val=$dm_fuse_params
+    fi
+    print -- $dm_fuse_params_val | \
+        datamon_sidecar_param parse fuse > $env_vars_file
```
galvare2: If I understand correctly, you've implemented a new datamon command-line endpoint, invoked here, that deserializes the yaml into a set of env vars that are saved in the env vars file. Is that correct?

hack/k8s/example-coord.template.yaml
ransomw1c: it's built via `hack/fuse-demo/write_fuse_params.go` ...

... external to docker then there's an `ADD` in the dockerfile.

let's sync up re. pipelines tool to find out how to tighten up integration.

hack/fuse-demo/wrap_datamon.sh
ransomw1c: yes, although this is an implementation detail that's intended to be opaque from your review

PR #319, opened 2019-12-06
fix(cli): make config create use the same file as config root

fredbi:
fix(cli): make config create use the same file as config root
 (i.e. ~/.datamon2/datamon.yaml)

* renamed config create source after actual command name
* error handling
* home dir retrieval using golang os builtin


cmd/datamon/cmd/config.go
```diff
@@ -15,14 +15,13 @@ import (
 type CLIConfig struct {
 	// bug in viper? Need to keep names of fields the same as the serialized names..
 	Credential string `json:"credential" yaml:"credential"` // Credentials to use for GCS
-	Config     string `json:"config" yaml:"config"`         // Config for datamon
+	Config     string `json:"config" yaml:"config"`         // Config bucket for datamon
```
ransomw1c: still makes little sense to me to call the bucket where contexts are stored the "config bucket" .. would think "context bucket" easier to remember and can only suppose that we're allowing for the possibility store config info other than context in this bucket.

cmd/datamon/cmd/config.go
```diff
@@ -15,14 +15,13 @@ import (
...
 	Context    string `json:"context" yaml:"context"`       // Context for datamon
 }
 
 func newConfig() (*CLIConfig, error) {
 	var config CLIConfig
-	err := viper.Unmarshal(&config)
-	if err != nil {
+	if err := viper.Unmarshal(&config); err != nil {
```
ransomw1c: more nitpick: whether to create non-`if ...`-scoped errors or `func`-scoped errors is another ongoing stylistic decision.  i still waffle on whether to use `if <statement>; <cond> { <block> }` or strictly stick with `if <cond> { <block> }` forms.  i usually err with the former as a function of how clever i'm feeling, and since "don't be clever" is the general advice in the golang ecosystem, it feels like eschewing the setup statement everywhere would be the more sensible rule ... i still feel clever sometimes, however.

cmd/datamon/cmd/config.go
fredbi: yeah I just wanted to strike the difference with the "local config file". If you feel strong about some clever formulation for it, feel free to change it.

cmd/datamon/cmd/config_create.go
```diff
@@ -2,56 +2,74 @@ package cmd
...
+
 		config := CLIConfig{
 			Config:     datamonFlags.core.Config,
 			Context:    datamonFlags.context.Descriptor.Name,
 			Credential: datamonFlags.root.credFile,
 		}
-		o, e := yaml.Marshal(config)
-		if e != nil {
-			wrapFatalln("serialize config to yaml", e)
+		o, err := yaml.Marshal(config)
```
ransomw1c: could there be a
```
switch configExt {
case ".yaml":
  o, err := <serialize-yaml>
default:
  wrapFatalln("implementation error", errors.New("unknown config format"))
```

here?  i mean, what's the point of keeping the extension as a constant if that constant remains separate and apart from the serialization?

cmd/datamon/cmd/config_create.go
```diff
@@ -2,56 +2,74 @@ package cmd
...
 	Run: func(cmd *cobra.Command, args []string) {
 		_, err := paramsToContributor(datamonFlags)
 		if err != nil {
 			wrapFatalln("contributor datamonFlags present", err)
 			return
 		}
-		user, err := user.Current()
-		if user == nil || err != nil {
-			wrapFatalln("Could not get home directory for user", nil)
+		home, err := os.UserHomeDir()
```
ransomw1c: good api catch 🎇

cmd/datamon/cmd/config_create.go
fredbi: I think i ll bring serialization / deserialization together for improved readability

cmd/datamon/cmd/config_create.go
```diff
@@ -0,0 +1,78 @@
...
+			wrapFatalln("contributor datamonFlags present", err)
+			return
+		}
+
+		config := CLIConfig{
+			Config:     datamonFlags.core.Config,
+			Context:    datamonFlags.context.Descriptor.Name,
+			Credential: datamonFlags.root.credFile,
+		}
+		o, err := config.MarshalConfig()
```
fredbi: @ransomw1c hopes this now addresses your concerns about whether to yaml or not to yaml

cmd/datamon/cmd/root.go
```diff
@@ -103,38 +79,48 @@ func init() {
...
+	// NOTE:
+	//   Since there is no viper env prefix set, we use raw viper key names from env vars, i.e. $CONTEXT will set
+	//   the context config key, something that is hard-wired below as DATAMON_CONTEXT...
+	//
+	//   Setting a prefix like "DATAMON" would be impractical, since DATAMON_CONFIG (the location of the local config file)
+	//   would conflict with the Config confif key (the bucket of the remote config).
+	//
+	// TODO: IMHO we should rename the 'config' key as 'remoteConfig', remove the DATAMON_CONTEXT code below and use
+	// viper.SetEnvPrefix("DATAMON").
+	//
```
fredbi: @ransomw1c here is a more complete explanation about what is wrong with env-driven config here, why I did not change it at this time (because this requires some extra changes and some agreement) and a proposal to go. Please advise.

cmd/datamon/cmd/root.go
```diff
@@ -150,15 +136,23 @@ func initConfig() {
...
-	for _, flag := range flags {
-		err := cmd.MarkFlagRequired(flag)
-		if err != nil {
-			err = cmd.MarkPersistentFlagRequired(flag)
-		}
-		if err != nil {
-			wrapFatalln(fmt.Sprintf("error attempting to mark the required flag %q", flag), err)
-			return
-		}
+func handleConfigErrors(err error) {
```
fredbi: @ransomw1c  hope this partially addresses your concerns ("qualms"...) expressed when reviewing the PR on error handling/wrapping.

cmd/datamon/cmd/config.go
```diff
@@ -39,31 +65,24 @@ func (c *CLIConfig) setDatamonParams(flags *flagsT) {
...
-	switch err {
-	case storagestatus.ErrInvalidResource:
-		return fmt.Sprintf("%s: please check that the config bucket %q is a valid gcs bucket", msg, about)
-	case storagestatus.ErrNotExists:
-		return fmt.Sprintf("%s: please check that the context has been created in your config", msg)
-	}
-	return msg
+// MarshalConfig produces a CLI config as a YAML document
+func (c *CLIConfig) MarshalConfig() ([]byte, error) {
+	return yaml.Marshal(c)
```
ransomw1c: same as when this was inlined:

```
switch configExt {
case ".yaml":
  o, err := <serialize-yaml>
default:
  wrapFatalln("implementation error", errors.New("unknown config format"))
```

to encode

> the extension constant agrees with serialization routine

cmd/datamon/cmd/config_create.go
ransomw1c: have added a comment in `MarshallConfig` to continue this nitpick

cmd/datamon/cmd/root.go
ransomw1c: have considered the overall golang arg parse situation at more length over the weekend.

as far as environment variables are concerned, being able to encode a direct env var to config correspondence rather than this string munging mapping (via prefix or anything else) would be more in accord with my prior expectations from, e.g., [Click](https://click.palletsprojects.com/en/7.x/arguments/#environment-variables).

a more drastic opinion is that all golang programs might as well be parameterized by either
* a simple `ARGBEGIN`/`ARGEND` or `optarg` sort of `for`-`switch` construction 

or

* a config file, either standard unix `.conf` style or `.yaml`, plus validation routines.

ransomw1c: for now, i think it's probably best to either spin off another issue based on these comments as well as the inline comments (removing the inline comments from this patch), or expect this to be deferred from the `2.0.0-<label>` release

cmd/datamon/cmd/root.go
ransomw1c: yes, this uses the declarative [type info] data to make substantive decision about program behavior (as opposed to string formatting), so it's considered the sort of win that adding types beyond `error` could afford that `fmt.Errorf("<new-msg>: %v", cause)` doesn't.

so cf. here from #323 in case the intent (still trying to overload the "concern" buzzword) is not clear from the writeup there.

cmd/datamon/cmd/config.go
fredbi: ok done with a if and with a dynamic extension check, since the target file may be overridden with DATAMON_CONFIG var (see example)

cmd/datamon/cmd/config_create.go
fredbi: wanted to be consistent with the MarshalXXX way to produce YAML docs adopted in pkg/models

cmd/datamon/cmd/root.go
fredbi: Agree that this is drifting away from the original intent of this PR. Let's have this one merged and move on forward.

My suggestion was merely about exploiting the capabilities of spf13/viper regarding env vars.

Just a word to say that viper is quite popular in goland, notwithstanding the qualities of your python references. My comment (now removed and traced as a new github issue) as intended to provide you with some understanding of the viper-way of doing it.

Not wanting to play the cargo cult developer here, not wanting either to waste to much time on this.

My point with this comment, was that (i) atm, it just doesn't work as a regular user would expect (principle of no surprise broken) and (ii) Since we've adopted the cobra/viper/pflags triptych, I'd favor ways in line with what the lib is suggesting us

cmd/datamon/cmd/config.go
ransomw1c: ok, i see that the `configExt` constant has been replaced by a `configFileLocation` function

cmd/datamon/cmd/root.go
ransomw1c: i see that #328 was spun off of this discussion. huzzah!

wrt. your point, total agreement vis (i).

re. (ii), the supposition that we're fully invested in cobra/viper/pflags, from conversation with Ritesh, if i recall correctly (and that's a distinct "if"), what i'm relaying here is that removing viper (if not cobra) is a possible horizon.

PR #320, opened 2019-12-06
feat: leverages error wrapping

fredbi:
* pkg (storage, core): wrap underlying API errors and process well-known error statuses
* CLI: add more detailed guidance on well-known errors (e.g. invalid bucket, unknown context)

**NOTE**: if we all agree on the idea behind this PR (which starts levering my previous error wrapper), I'll start generalizing error wrapping in packages. Step-by-step CLI testing shall also drive toward more "extra guidance"-like handling.

### DIFF
--- a/cmd/datamon/cmd/self_upgrade.go
+++ b/cmd/datamon/cmd/self_upgrade.go
@@ -107,7 +107,7 @@ func doCheckVersion() error {
 		return errors.New(fmt.Sprintf("could not fetch release from github repo (%s)", githubRepo)).Wrap(err)
 	}
 	if !found {
-		return errors.New(fmt.Sprintf("no matching release from github repo (%s)", githubRepo)).Wrap(err)
+		return errors.New(fmt.Sprintf("no matching release from github repo (%s)", githubRepo))
 	}
 
 	if isRelease && latest.Version.Equals(v) {
--- a/pkg/storage/gcs/errors.go
+++ b/pkg/storage/gcs/errors.go
@@ -1,0 +1,42 @@
+package gcs
+
+import (
+	"strings"
+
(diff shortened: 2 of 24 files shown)


cmd/datamon/cmd/self_upgrade.go
```diff
@@ -107,7 +107,7 @@ func doCheckVersion() error {
 		return errors.New(fmt.Sprintf("could not fetch release from github repo (%s)", githubRepo)).Wrap(err)
 	}
 	if !found {
-		return errors.New(fmt.Sprintf("no matching release from github repo (%s)", githubRepo)).Wrap(err)
+		return errors.New(fmt.Sprintf("no matching release from github repo (%s)", githubRepo))
```
ransomw1c: wait, now we're _not_ wrapping cause errors in the errors they cause?

i thought that entire discussion about the error programming pattern change in 1.13 was to use implement and use a `getCause()`-like api (in java land .. many languages implement the error chaining motif)

pkg/storage/gcs/errors.go
```diff
@@ -0,0 +1,42 @@
...
+		return status.ErrForbidden.Wrap(err)
+	case 404:
+		return status.ErrNotFound.Wrap(err)
+	default:
+		return status.ErrStorageAPI.Wrap(err)
+	}
+}
+
+func toSentinelErrors(err error) error {
+	// return sentinel errors defined by the status package
```
ransomw1c: what is the status package?

cmd/datamon/cmd/self_upgrade.go
fredbi: This very line has actually not err to wrap henve the remival

PR #322, opened 2019-12-06
fix(context): commands on repo, bundle, labels should be able to set …

fredbi:
…a context

* added inherited context flag to repo, bundle, label command sets
* removed context flags from individual subcommands
* factorized the way we specify required flags and support inherited flags locally required


ransomw1c:
incidentally, now that we don't specify flags only at the leaves of the command tree, i no longer understand how `add*()` functions correspond to flags available to commands...

.. the `add*()` functions don't indicate whether the flag is added with `PersistentFlags()` or `Flags()`, and i find it difficult to keep track of whether a flag is persistent or not without referring back to `flags.go`.

this is not something i want to juggle, although it might also be out of scope for this merge request.  file this comment under generalized qualms with the `cmd/` implementation that i'd like to see addressed at some point.

fredbi:
Yes introducing persistent flags now raises some inconsistencies in the approach. I did not want to change all of this in one go

PR #329, opened 2019-12-10
chore(deps): updated dependencies

fredbi:
* updated google storage and google api dependencies
* updated jacobsa/fuse and implemented extra fs interface
* udpdated zap dep

* switched to mainstream spf13/cobra: I don't know why we got stick to a fork, but now this fork is unmaintained
* commented the replace clause for test dependency on goleak (all replace should be somehow explained as they feel mostly like temp hacks)
* go mod tidy (cleaned up undue dep on lstags - artifact of a local experimental build of mine...)


go.mod
```diff
@@ -44,13 +38,20 @@ require (
 	github.com/spf13/viper v1.4.0
 	github.com/stretchr/objx v0.2.0 // indirect
 	github.com/stretchr/testify v1.4.0
+	go.opencensus.io v0.22.2 // indirect
 	go.uber.org/goleak v0.0.0-00010101000000-000000000000
```
ransomw1c: i'm confused about why goleak still appears here in the `require` directive ..

.. in fact, attempting to consume the datamon api from another module, i'm noticing that i get a build error unless i duplicate the `replace` directive for goleak..

.. not something that needs to be resolved as part of this merge request, although it does intersect the overall dep mgmt. story, and i could use some help getting that sorted in a less hack way.

ransomw1c:
`jacobsa/fuse` has [`NotImplementedFileSystem`](https://godoc.org/github.com/jacobsa/fuse/fuseutil#NotImplementedFileSystem) to use as an [embedded field](https://golang.org/ref/spec#Struct_types) such that additional methods to the fuse filesystem are automatically added.

idk if this is viewed as a helpful or harmful design decision:  it might be better to have to explicitly maintain stubs for all fuse callbacks?  i did want to mention this and get a definite decision (or explicitly get results of that decision in case it was already considered) on whether to use `NotImplementedFileSystem` or not.

go.mod
fredbi: Goleak is a test dependency I introduced a couple weeks ago. Pb with this repo is that the latest release is niw quite old with sime breaking api changes in master. I agree that we could pin the commit directly rather than using replace which as you rightly point out not honored when consumed by another module. I am fixing this part today

fredbi:
Good point. I ve seen seen too (i also implemented similar things in some other projects).

go.mod
fredbi: done

fredbi:
@ransomw1c openend follow-up #332. Let's have a discussion on which is stub and which is not and some clarification on why we wanted specifically to return ENOSYS on some many methods.

pkg/core/fs_ro_ops.go
```diff
@@ -404,6 +404,15 @@ func (fs *readOnlyFsInternal) Destroy() {
 	)
 }
 
+func (fs *readOnlyFsInternal) Fallocate(
```
kerneltime: WIP slipped in?

PR #334, opened 2019-12-13
feat: context list CLI

fredbi:
* adds datamon context list command
* revisits remote config metadata structure for contexts, i.e. adds path `contexts/{context}`
* makes context rendering more explicit (template)
* fixes [RES-1438]

## ATTENTION
The CI ~~does not pass because this PR~~ now passes -- This PR brings a breaking change to the structure of the remote config. If we agree on the principle of this change, I can make it to work by adapting the remote config (simply copying the */context.yaml into contexts/*/context.yaml in the config bucket).

~~For the moment, I am not changing the test remote config that Ransom did create.~~

@ransomw1c I've now adapted the remote config bucket used by the fuse test, but did not remove anything yet: just copied the context.yaml definitions into the new `/contexts/{xxxx}/context.yaml` 

[RES-1438]: https://oneconcern.atlassian.net/browse/RES-1438

### DIFF
--- a/cmd/datamon/cmd/context.go
+++ b/cmd/datamon/cmd/context.go
@@ -6,8 +6,11 @@
 package cmd
 
 import (
+	"context"
 	"text/template"
 
+	"github.com/oneconcern/datamon/pkg/storage"
+	"github.com/oneconcern/datamon/pkg/storage/gcs"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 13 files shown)


cmd/datamon/cmd/context.go
```diff
@@ -28,10 +31,24 @@ var contextTemplate *template.Template
 
 func init() {
 	rootCmd.AddCommand(ContextCmd)
-	addContextFlag(ContextCmd)
 
 	contextTemplate = func() *template.Template {
-		const listLineTemplateString = `{{.Version}} , {{.Name}} , {{.WAL}} , {{.ReadLog}} , {{.Blob}} , {{.Metadata}} , {{.VMetadata}}`
+		const listLineTemplateString = `Model Version: {{.Version}}
```
ransomw1c: everywhere else datamon needs to print metadata to stdout, it uses the ` , `-delimited format.

i suggest we keep a consistent user interface at the cli:  all `get` and `list` commands should print the same kind of output.

now, we have had requests to change the output format and/or allow different formats for this kind of output (e.g. Gabe has mentioned wanting to use json).

my only request here is for consistency:  let's have all the `get` and `list` commands do about the same thing if possible.

fredbi: ok will do that. I just found out too that the output was difficult to read...

fredbi: ok done @ransomw1c

PR #345, opened 2019-12-27
Log bundle file list to std out

kerneltime:


ransomw1c:
i have verified that `fmt.Print*` do always print to standard out.

not overly opinionated about how to implement, although it does occur to me that since error-handling is somewhat "DSL-ified" via the `wrapFatal*` functions, there could be something similar here wherein all the `get` and `list` commands all rely on the same package-private routine(s) to write their output.

PR #346, opened 2019-12-30
Rename create config to config set.

kerneltime:
Alias create to set to avoid breakage.

### DIFF
--- a/docs/usage/datamon_config_set.md
+++ b/docs/usage/datamon_config_set.md
@@ -1,12 +1,12 @@
 **Version: dev**
 
-## datamon config create
+## datamon config set
 
 Create a local config file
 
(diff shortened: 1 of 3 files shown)


docs/usage/datamon_config_set.md
```diff
@@ -0,0 +1,59 @@
+**Version: dev**
```
fredbi: the md file `datamon_config_create.md`  should be deleted. The git hook does not take care of that automatically yet. Maybe it is the right time to add this (or I'll do it later).

kerneltime: thanks!

kerneltime: done

PR #352, opened 2020-01-02
feat(web): added auto-start with browser on datamon web CLI

fredbi:
* changed default port to random, --port can still be used to specify a port
* added no-browser flag, to prevent launching of browser


fredbi:
@kerneltime  this has been tested on linux (with xdg-open feature). Could you please run some minimal sanity check on macos?

PR #354, opened 2020-01-02
[wip] dbg mutable mount

ransomw1c:
_prioritized in today's scrum_

### DIFF
--- a/pkg/core/fs.go
+++ b/pkg/core/fs.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -82,8 +83,16 @@ func NewReadOnlyFS(bundle *Bundle, l *zap.Logger) (*ReadOnlyFS, error) {
 }
 
 // NewMutableFS creates a new instance of the datamon filesystem.
-func NewMutableFS(bundle *Bundle, pathToStaging string) (*MutableFS, error) {
+func NewMutableFS(bundle *Bundle) (*MutableFS, error) {
 	logger, _ := zap.NewProduction()
+	consumableStoreString := bundle.ConsumableStore.String()
+	consumableStoreStringSplit := strings.Split(consumableStoreString, "@")
+	if len(consumableStoreStringSplit) != 2 ||
+		consumableStoreStringSplit[0] != "localfs" {
+		return nil, errors.New("bundle doesn't have localfs consumable store " +
+			"to provide local cache for muable fs")
+	}
+	pathToStaging := consumableStoreStringSplit[1]
 	fs := &fsMutable{
 		bundle:       bundle,
 		readDirMap:   make(map[fuseops.InodeID]map[fuseops.InodeID]*fuseutil.Dirent),
--- a/pkg/core/fs_rw_ops.go
+++ b/pkg/core/fs_rw_ops.go
@@ -493,8 +493,12 @@ func (fs *fsMutable) ReadFile(
 func (fs *fsMutable) WriteFile(
 	ctx context.Context,
 	op *fuseops.WriteFileOp) (err error) {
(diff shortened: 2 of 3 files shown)


ransomw1c:
##### note to self

see about getting the fuse tests activated in ci as part of this effort.

pkg/core/fs.go
```diff
@@ -82,8 +83,16 @@ func NewReadOnlyFS(bundle *Bundle, l *zap.Logger) (*ReadOnlyFS, error) {
 }
 
 // NewMutableFS creates a new instance of the datamon filesystem.
-func NewMutableFS(bundle *Bundle, pathToStaging string) (*MutableFS, error) {
+func NewMutableFS(bundle *Bundle) (*MutableFS, error) {
 	logger, _ := zap.NewProduction()
+	consumableStoreString := bundle.ConsumableStore.String()
```
ransomw1c: this string typing and parsing is, of course, a mess.

what's going on here is an attempt to impedance match two extant implementations:

* cli parsing

and

* mutable mount internal data structures

without changing the assumptions made by either such that the overall advertised functionality (omitting `--destination` on the cli) is in place

fredbi: :+1: for removing redundancy here

pkg/core/fs_rw_ops.go
```diff
@@ -536,10 +540,12 @@ func (fs *fsMutable) FlushFile(
 	f := fs.backingFiles[op.Inode]
```
fredbi: DISCLAIMER: I am discovering this piece, so this might not be a comment for this PR, but rather a marker for future investigation.

This map access just makes our fs not concurrency-safe.

pkg/core/fs_rw_ops.go
```diff
@@ -536,10 +540,12 @@ func (fs *fsMutable) FlushFile(
...
 	if f != nil {
 		file := *f
+if file != nil {
```
fredbi: I don't quite understand when we insert a nil in the map.
I believe there is still a possible panicking cause in `SyncFile()` with `file := *backingFiles[op.Inode]` 
but `FlushFile` should be safe, calling `file.Sync()` without necessarily passing by the `file := *fs`  staging.
Like:
```go
	file := fs.backingFiles[op.Inode]
	if file != nil {
		err := file.Sync()
		if err != nil {
			return fuse.EIO
		}
	}
	return
```

pkg/core/fs.go
ransomw1c: > if we want to keep this customizable

a reason to keep this customizable is to allow fixed-size ramdisk k8s volumes to be specified in the yaml, leaning on the kubernetes system to provide storage, rather than the virtualized os within the k8s container.

> I am thinking that a `--staging` flag would be clearer for the datamon bundle mount new command

i agree that the meaning of `--destination` within `bundle mount new` is different from other commands, `bundle download`, say.  namely, because everywhere else (and i should probably double-check this) `--destination` is used, it corresponds to a directory that's modified via a discrete, one-time download operation.

pkg/core/fs_rw_ops.go
ransomw1c: i threw this conditional in while debugging .. yes, it needs to be removed before merge.

pkg/core/fs_rw_ops.go
fredbi: Re concurrency-safety on ro FS: reading further, I realize that the populateFS procedure is never parallelized atm. Hence false alarm here.

Re concurrency-safety on rw FS: similarly, I inspected further how the global single lock prevents  concurrent map access to the readDir map, but not to the backingFiles map.
I am proposing a fix for that in #358

ransomw1c:
please leave branch on remote to re-open later

ransomw1c:
rebase 6473ea64605c5384f indicates this was effectively debugged already

PR #355, opened 2020-01-02
Make flags for context config persistent at root.

kerneltime:
Context and config are global flags and need to be persistent.

### DIFF
--- a/cmd/datamon/cmd/flags.go
+++ b/cmd/datamon/cmd/flags.go
@@ -170,7 +170,7 @@ func addContextFlag(cmd *cobra.Command) string {
 
 func addConfigFlag(cmd *cobra.Command) string {
 	config := "config"
-	cmd.Flags().StringVar(&datamonFlags.core.Config, config, "", "Set the config backend store to use (bucket name: do not set the scheme, e.g. 'gs://')")
+	cmd.PersistentFlags().StringVar(&datamonFlags.core.Config, config, "", "Set the config backend store to use (bucket name: do not set the scheme, e.g. 'gs://')")
 	return config
 }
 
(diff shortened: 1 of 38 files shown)


cmd/datamon/cmd/repo.go
```diff
@@ -28,7 +28,6 @@ var repoDescriptorTemplate *template.Template
 
 func init() {
 	rootCmd.AddCommand(repoCmd)
-	addContextFlag(repoCmd)
```
fredbi: The `addContextFlag`  is also present in `bundle.go`  and `config_set.go`. It should be removed there.
For `context_*.go` we have a special case because the flag is required.

cmd/datamon/cmd/flags.go
```diff
@@ -169,7 +169,7 @@ func addContextFlag(cmd *cobra.Command) string {
 
 func addConfigFlag(cmd *cobra.Command) string {
 	config := "config"
-	cmd.Flags().StringVar(&datamonFlags.core.Config, config, "", "Set the config backend store to use (bucket name: do not set the scheme, e.g. 'gs://')")
+	cmd.PersistentFlags().StringVar(&datamonFlags.core.Config, config, "", "Set the config backend store to use (bucket name: do not set the scheme, e.g. 'gs://')")
```
fredbi: The `addConfigFlag` call should be removed in `config_set.go` and `context_*.go`

cmd/datamon/cmd/repo.go
kerneltime: Will do.

fredbi:
@kerneltime looks like your initial commit is not PGP sgned and this is a requirement on this repo.
I believe that squashing your commits into a PGP signed single commit would allow us to merge

PR #358, opened 2020-01-06
coreFS minor enhancements

### LINKED ISSUE #332: stubbing fuse fs
## Follow up on #329 (updated fuse dependency)

@ransomw1c 
> jacobsa/fuse has NotImplementedFileSystem to use as an embedded field such that additional 
methods to the fuse filesystem are automatically added.

> idk if this is viewed as a helpful or harmful design decision: it might be better to have to explicitly maintain stubs for all fuse callbacks? i did want to mention this and get a definite decision (or explicitly get results of that decision in case it was already considered) on whether to use NotImplementedFileSystem or not.

@fredbi 
> Good point. I ve seen seen too (i also implemented similar things in some other projects).

fredbi:
fix(core): removed explicit stubbed FS methods and replaced them by imported fuse lib NotImplemented versions

This makes it easier to update the dependency to the fuse package (but we lose the logging of non implemented methods)

* fixes #332

fix(core): factorized type declaration and logging on core FS implementations

Same logging mechanism and interface drives both ro and rw implementations.

### DIFF
--- a/cmd/datamon/cmd/bundle_mutable_mount.go
+++ b/cmd/datamon/cmd/bundle_mutable_mount.go
@@ -9,6 +9,7 @@ import (
 	daemonizer "github.com/jacobsa/daemonize"
 
 	"github.com/oneconcern/datamon/pkg/core"
+	"github.com/oneconcern/datamon/pkg/dlogger"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/datamon/cmd/bundle_mutable_mount.go
```diff
@@ -52,7 +53,12 @@ The destination path is a temporary staging area for write operations.`,
 		bundle := core.NewBundle(bd,
 			bundleOpts...,
 		)
-		fs, err := core.NewMutableFS(bundle, datamonFlags.bundle.DataPath)
+		logger, err := dlogger.GetLogger(datamonFlags.root.logLevel)
```
ransomw1c: 🎇

ransomw1c:
golang note to self:  methods are promoted as well as struct fields.

PR #338, opened 2019-12-16
WIP fix(ci): experiments on how to make CI run faster

fredbi:
1. Make linting startup with minimal overhead

TODO:
2. Parallel unit test [hit bug here]
3. build images earlier in proces?

Signed-off-by: Frederic BIDON <frederic@oneconcern.com>

PR #359, opened 2020-01-07
Add pacrk support for go releaser

kerneltime:

This is not tested, I will have to cut a release and try..


fredbi:
Good catch. I've been investigating (see #338) another method which is to pre-build - and update, say, weekly - a builder image (and save a little time rebuilding this during CI)

fredbi:
Well so far, my investigation with #338 show only marginal gains, with 1 min less spent building... :(

.goreleaser.yml
```diff
@@ -1,6 +1,7 @@
 project_name: datamon
 before:
   hooks:
+    - go get -u github.com/gobuffalo/packr/v2/packr2
```
fredbi: No this won't work:
- goreleaser will detect that the git tree is not clean and will refuse to release

The only way, AFAIK, is to commit the generated code.

kerneltime: Good to know. I was debating the same choice, it seems to be the cleanest.

ransomw1c: in case we're looking for additional options,

using the `Makefile` to drive packr, including and `git` setup and teardown necessary to satisfy goreleaser, both in CI and on the local, is also possible.  the advantage here is single source of truth.

PR #361, opened 2020-01-09
chore(core): more unit testing of the fuse mount core capabilities

### LINKED ISSUE #332: stubbing fuse fs

fredbi:
* refactored bundle unit tests
* refactored fuse mount unit tests, with more test scenarios
* fixes #332
* fixes map concurrency issues in fuse RW mount
* configured a parallel job on circleCI VM to run integration test with fuse (introduced some throttling to avoid hangs, which I could not reproduce otherwise)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,16 @@ executors:
     docker:
       - image: circleci/golang
 
+  fuse_tester:
+    working_directory: ~/project
+    environment:
+      GOOGLE_PROJECT_ID: "onec-co"
+      GOOGLE_COMPUTE_ZONE: us-west2-c
+      GOROOT: /usr/local/go
+      GOPATH: /go
+    machine:
+      image: ubuntu-1604:201903-01
+
 commands:
   install_base:
     steps:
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -43,6 +53,28 @@ commands:
...
+          environment:
+            GOROOT: /usr/local/go
+            GOPATH: /go
+          command: |
+            GO_VERSION=1.13.5
+            CI_USER=$(id -u)
+            sudo apt-get install --quiet fuse
+
+            cd /tmp
+            wget --no-verbose https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz
```
ransomw1c: does the FUSE CI setup play without this golang environment install?

not saying we need to change the method of installing golang.  wondering if the golang and fuse installs are conceptually grouped, or if this patch is more about going with what works in order to prop up fuse on ci in the first place.

ransomw1c:
> fixes map concurrency issues in fuse RW mount

specifically, added lock acquisition around a couple of critical sections of the FUSE IoC mechanism where it had been omitted.

abstractly, the fuse implementations each consist of two parts

* populating an DM-internal data structure from data in GCS
* connecting the data structure to FUSE via the `go-fuse` IoC api

i when mutable mount development trailed off, we were primarily considering changes to the `populateFS` function.  the internal data structure currently burns far more memory as a function bundle size than we expected.

.circleci/config.yml
fredbi: @ransomw1c yes I've added to golang to run the unit tests only (since I spin up a VM, I did not go so far so as to use that VM _only_ to build docker containers to run golang. That could have been a way to go, albeit a lengthy one.

We also might want to use the VM as  a more general purpose docker server specifically to run privileged containers (which can't be achieved with the standard circleCI docker), as per required by the fuse driver, but this is out of the scope of this PR: going further toward CI optimization, one could pre-built the VM image and make it available globally to our circleCI account etc - as for now, I don't have the grants to do something like that.

PR #364, opened 2020-01-13
minor sidecar stdio fixes

ransomw1c:
there are a few error messages or parts of error messages that the sidecar was printing to the wrong output stream that this patch corrects


ransomw1c:
are test failures due to known issues?

failing tests pertain to golang, and this patch doesn't change any of the golang, yet golang tests are failing in ci.

should i look into debugging the ci setup?

fredbi:
test fails because of the change in binary name in release artifacts. Pushing a fix for that today

PR #365, opened 2020-01-14
adapted self-upgrade test to new binary datamon2

fredbi:


fredbi:
@ransomw1c that should get you going for  #364

PR #368, opened 2020-01-23
Review of the fuse mount chain of command

fredbi:
**Tracing**
* storage/gcs: added logging and debug traces

* cmd(CLI): refactored logger injection, added more commands subject to log level setting

* core: enabled fuse internal debug logger (experimental, since it is quite verbose)

* cafs: injected logger

**Tuning/user experimentation**
* cmd(CLI): added new options to tune fuse mount cache size and prefetch behavior (e.g. --cache-size 100MB, --prefetch 3)

* core: bundle conveys new options down to cafs factory

**Refactoring & readability**
* core: refactored NewXXXFS to suppress redundant staging path and logging (following initial idea from PR#354)
* core: refactored RO mount for readability

* cafs: more godoc, resolved question marks left in code
* cafs: refactoring of struct factories, chunkReader and writer, more options to control LRU cache and prefetch behavior
* cafs: refactored hasher ("helpers") for readability

**Testing**
* cmd(CLI): re-enacted fuse CLI e2e testing (runs on CI)

* cafs: added more unit tests on freelists, hasher and ReadAt with concurrency testing

**Fixes**
* core: fixed spurious warnings on fs (extended attributes, flush on RO mount)

* cafs: enabled evicted LRU buffers to be relinquished to freelist
* cafs: fixed concurrency issues with corrupted relinquished buffers
* cafs: fixed leaf hash verification for ReadAt

**Performance improvements**
* cafs: set up cache for resolved leaf keys from root
* cafs: more efficient prefetching, with much less waste of work and prefetching blobs with tunable look-ahead

**Memory footprint reduction**
* core: relinquished unused RadixTree after ro fs construction

* cafs: adapted freelist buffers to leaf size, not systematically max size

**CI changes**
* cafs: cafs tests are memory hungry: run them in CI separately with different settings (less parallelism, no race)
* hack/fuse_demo: adapted fuse mount detection, since we slightly changed the mount signature (see below)

**Other enhancements**
* core: set fuse moumt subtype to "datamon" and "datamon-mutable", so it appears as such in the output of the "mount" UNIX command


fredbi:
Still working on the prefetch stuff. Also experimenting with some read chunking of a single leaf

pkg/cafs/cafs.go
```diff
@@ -4,89 +4,51 @@ import (
...
+	// DefaultLeafSize sets the the default size of a blob leaf (2 MB). It cannot exceed MaxLeafSize.
+	// The actual leaf size used is usually specified by each bundle.
+	DefaultLeafSize uint32 = 2 * units.MiB
 
-type hasOpts struct {
-	OnlyRoots, GatherIncomplete bool
-	_                           struct{} // disallow unkeyed usage
-}
+	// MaxLeafSize is the maximum size of a buffer in the memory pool
+	MaxLeafSize = 5 * units.MiB
```
kerneltime: We can set this to be the same as leaf size?

pkg/storage/multi.go
```diff
@@ -27,11 +28,18 @@ func ReadTee(ctx context.Context, sStore Store, source string, dStore Store, des
...
+	Store Store
+
+	// TolerateFailure to false breaks multi-store operations whenever an error is encountered.
 	TolerateFailure bool
 }
 
+// MultiPut duplicates write operations to an array of stores, under the same name
+//
+// TODO(fred): the aliased type NewKey is just confusing: fallback to bool
 func MultiPut(ctx context.Context, stores []MultiStoreUnit, name string, buffer []byte, doesNotExist NewKey) error {
```
kerneltime: FYI: This was added to allow for a local cache to which the same data is written to as the remote CAFS.. This was when we were still chasing k8s CSI integration.

pkg/storage/gcs/store.go
```diff
@@ -99,18 +119,23 @@ func (r *gcsReader) ReadAt(p []byte, offset int64) (n int, err error) {
 }
 
 func (g *gcs) Get(ctx context.Context, objectName string) (io.ReadCloser, error) {
+	g.l.Debug("Start Get", zap.String("objectName", objectName))
 	objectReader, err := g.readOnlyClient.Bucket(g.bucket).Object(objectName).NewReader(ctx)
+	g.l.Debug("End Get", zap.String("objectName", objectName))
```
kerneltime: Capture the error, same comment for other lines here.

pkg/storage/gcs/store.go
```diff
@@ -125,26 +150,35 @@ func (g *gcs) GetAt(ctx context.Context, objectName string) (io.ReaderAt, error)
...
 	reader io.Reader
 }
 
-func (rc readCloser) Read(p []byte) (n int, err error) {
+func (rc readCloser) Read(p []byte) (int, error) {
 	return rc.reader.Read(p)
 }
+
 func (rc readCloser) Close() error {
+	// TODO(fred): here is what the google storage api states on Write:
```
kerneltime: This is for a read closer.. if the read completed then close can fail. This is not used for writes.

pkg/cafs/cafs.go
fredbi: @kerneltime this is used only by the FreeList. It corresponds to the maximum supported size to allocate leaf buffers. When the freelist kicks in, it dynamically adopts a buffer size not too much larger than the actual leaf size of the bundle.

pkg/storage/gcs/store.go
fredbi: ok good catch

pkg/storage/gcs/store.go
fredbi: ok

pkg/storage/multi.go
fredbi: @kerneltime ok got it.
My remark was specifically about using a type alias here, which is basically pointless.
Type aliases are not the way to construct enums in go, if this was the initial intent here. Moreover, there is no real point in constructing an enum for a boolean value.

My point was just to replace this by: 
`func MultiPut(ctx context.Context, stores []MultiStoreUnit, name string, buffer []byte, doesNotExist bool) error`

pkg/cafs/reader.go
```diff
@@ -201,115 +273,332 @@ func (r *chunkReader) WriteTo(writer io.Writer) (n int64, err error) {
...
+}
+
+func calculateKeyAndOffset(off int64, leafSize uint32) (index int, offset int64) {
+	index = int(off / int64(leafSize))
 	offset = off % int64(leafSize)
 	return
 }
 
-func (r *chunkReader) ReadAt(p []byte, off int64) (totread int, err error) {
+func readLeafFunc(r *chunkReader) func(Key, int, int, <-chan struct{}) (LeafBuffer, bool, error) {
```
kerneltime: This is a busy function, please add a comment to the signature on what this function will do. It also triggers the prefetch mechanism in addition to reading the leaf.

pkg/cafs/cafs.go
fredbi: won't fix

pkg/storage/multi.go
fredbi: will do later

pkg/cafs/reader.go
fredbi: yeah as it used to... Ok to add comments.

pkg/cafs/reader.go
```diff
@@ -201,115 +273,332 @@ func (r *chunkReader) WriteTo(writer io.Writer) (n int64, err error) {
...
+	// * we request leaves ahead only once, unless they have been wiped from buffer cache (reduce wasted work)
+	// * we cannot totally rule out wasted work because the buffer cache may not be able to retain all buffers
+	//
+	// A way to assert that there is no more wasted work than allowed by memory constraint:
+	// $ DEBUG_TEST=1 go test -v -run Reader_All|grep -E '(wasted work)|(has read leaf)'|jq -c '{key: .key,msg: .msg}'|sort
+
+	r.fetchingLatch.Lock()
+	defer r.fetchingLatch.Unlock()
+
+	if _, beingFetched := r.fetching[index+1]; !beingFetched {
```
kerneltime: Why not just have a hash map of the keys in flight, start a goroutine that waits for the key to be removed from the map and be present in the cache. If we implement a different prefetch scheme the logic will still hold true.

fredbi: I thought about this, but came to recursive decision making about prefetching, reassessing the current situation every time the process requests a new leaf.

Indeed, this is a moving target: new keys to fetch ahead are discovered as the reading goes: every time a leaf is cached, I jump next, and assess the new next fetch-ahead list.

Moreover, the cache cannot guarantee to hold buffers fetched ahead: sometimes (a few as I could, and only when we have much more readers than available cache entries) the prefetch work is wasted (must more uncommon event than in the previous implementation, though) and the leaft must be read again.

fredbi: so basically, the doPrefetch is here to:
i) decide whether to prefetch or not
ii) manage the current list of leafs being prefetched
iii) use the buffer cache as the referee on whether to consider work done or declare it wasted

.circleci/config.yml
```diff
@@ -125,7 +125,7 @@ jobs:
           command: |
             hack/go-generate.sh
             go mod download
-            golangci-lint run
+            golangci-lint run --build-tags fuse_cli,fsintegration
```
ransomw1c: didn't we start the lint rc file, `.golangci.yml` because we think separation of concerns vis
* linter config
and
* ci config
 is desirable?

ransomw1c:
meta-review:  where, as a project, do we stand on the ["one idea per commit"](https://secure.phabricator.com/book/phabflavor/article/recommendations_on_revision_control/) practice advocated by some?

my first impression upon reading the description and noticing the size of the diff is that this could probably be separated out into a few merge requests?

i know this is extra effort on the person preparing the requests, and [we've seen](https://oneconcern.slack.com/archives/CCMQ25SP5/p1579196570047700) that needing to rebase several commits for merge all at once can lead to errors.  yet i still suppose that the benefits to the project that would result from getting this landed as a sequence of commits in the history of `master` could result in both a clearer review now as well as a clearer record of the changes for later.

.circleci/config.yml
```diff
@@ -158,7 +158,11 @@ jobs:
             export GOOGLE_APPLICATION_CREDENTIALS=$HOME/extra/appcredentials.json
             hack/go-generate.sh
             go mod download
-            gotestsum --junitfile /tmp/test-results/go-test-report.xml --format short-with-failures -- -race ./...
+            # cafs tests are memory intensive: best to run them separately
+            gotestsum --junitfile /tmp/test-results/go-test-report.xml --format short-with-failures -- \
+              -race $(go list ./...|grep -v cafs)
```
ransomw1c: nit:  number of source files now limited by shell's maximum number of parameters

cmd/datamon/cmd/config.go
```diff
@@ -100,6 +105,17 @@ func (*CLIConfig) populateRemoteConfig(flags *flagsT) {
 	flags.context.Descriptor = contextDescriptor
 }
 
+func (c *CLIConfig) mustGetLogger(flags flagsT) *zap.Logger {
```
ransomw1c: cool 🎇 this sort of thing gives hope that there's a refactoring path to keep `cmd/` modular

* attach more dsl-ified funcs to types' namespaces of methods
* refactor types defining such dsl-ish things to a separate package

pkg/core/fs.go
```diff
@@ -71,27 +80,50 @@ func NewReadOnlyFS(bundle *Bundle, l *zap.Logger) (*ReadOnlyFS, error) {
...
+	// NOTE: this fetches the bundle metadata only and builds
+	// an in-memory view of the file system structure.
+	// This may lead to a large memory footprint for bundles
+	// with many files (e.g. thousands)
+	//
+	// TODO: reduce memory footprint
 	return fs.populateFS(bundle)
 }
 
+func localPath(consumable fmt.Stringer) (string, error) {
```
ransomw1c: can we land this incremental improvement with a `TODO` or `NOTE` or `???` s'th to indicate that the methodology of extracting this info via 
* string operations on the return values from the `storage` api
rather than
* functions or methods defined in `storage`
remains a questionable design decision?

i think i suggested doing this at one point in order to make a one-idea-per-commit change, and i suppose the current implementation is better than the previous, so i'm not looking to see `storage` change here, just to get the `grep`-able tag in the comments so we recall that this could be further refined.

cmd/datamon/cmd/bundle.go
```diff
@@ -69,3 +69,12 @@ func setLatestOrLabelledBundle(ctx context.Context, remote context2.Stores) erro
 	log.Printf("Using bundle: %s", datamonFlags.bundle.ID)
 	return nil
 }
+
+func getConcurrencyFactor(batchSize int) int {
```
ransomw1c: ✨ yes, totally forgot to add this coercion to prevent show-stoppers, which this could've been otherwise

cmd/datamon/cmd/cli_fuse_test.go
```diff
@@ -3,166 +3,331 @@
...
+	require.NoError(t, err)
+	pipeErr, err := cmd.StderrPipe()
+	require.NoError(t, err)
+
+	// combine stdout & stderr, tee this output to os.Stdout and return pipe reader for output scanning
+	pipeR, pipeW := io.Pipe()
+	c := io.MultiWriter(os.Stdout, pipeW)
+
+	go func() {
+		_, _ = io.Copy(c, pipeOut)
```
ransomw1c: i'm trying to understand golang stdlib api here:

why use `MultiWriter` to fan out but two `Copy` calls instead of a `MultiReader` and one `Copy`?  are the two equivalent?

cmd/datamon/cmd/cli_fuse_test.go
```diff
@@ -3,166 +3,331 @@
...
+func testLogger(t testing.TB) *zap.Logger {
+	// a logger for the test sequence itself
+	return dlogger.MustGetLogger("debug").With(zap.String("test", t.Name()))
+}
+func testCommand(t testing.TB, target string, args ...string) (*exec.Cmd, io.ReadCloser) {
+	// NOTE: this starts a process, not a goroutine like for other tests
+	// That is primarily because we want to send a signal to the process
+	// in order to unmount.
+	cmd := exec.Command(target, args...)
+	cmd.Env = append(cmd.Env, "DATAMON_CONTEXT="+os.Getenv("DATAMON_CONTEXT")) // TODO(fred): not sure this is used
```
ransomw1c: is this environment variable forwarding what was preventing the tests from running in ci?

i fully appreciate the difficulty of the changes here to block based on stdio rather than a dumb timeout.

it's that i'm having difficulty separating the changes necessary to replace the dumb timeout with the blocking function created by `testMountReady` for those necessary to `Exec` the fuse command in ci.

cmd/datamon/cmd/config.go
```diff
@@ -52,9 +55,11 @@ func configFileLocation(expandEnv bool) string {
...
 type CLIConfig struct {
 	// bug in viper? Need to keep names of fields the same as the serialized names..
-	Credential string `json:"credential" yaml:"credential"` // Credentials to use for GCS
-	Config     string `json:"config" yaml:"config"`         // Config bucket for datamon contexts and metadata
-	Context    string `json:"context" yaml:"context"`       // Current context for datamon
+	Credential  string `json:"credential" yaml:"credential"` // Credentials to use for GCS
+	Config      string `json:"config" yaml:"config"`         // Config bucket for datamon contexts and metadata
+	Context     string `json:"context" yaml:"context"`       // Current context for datamon
+	logger      *zap.Logger
+	oneceLogger sync.Once
```
ransomw1c: typo?  `oneceLogger` -> `onceLogger`?

kerneltime:
I have to go over tests but the changes overall are in the right direction. Will continue to review.

cmd/datamon/cmd/context_create.go
```diff
@@ -24,7 +24,10 @@ var ContextCreateCommand = &cobra.Command{
 }
 
 func createContext() {
-	configStore, err := gcs.New(context2.Background(), datamonFlags.core.Config, config.Credential)
+	logger := config.mustGetLogger(datamonFlags)
+	configStore, err := gcs.New(context2.Background(), datamonFlags.core.Config, config.Credential,
+		gcs.Logger(logger),
```
ransomw1c: nit:  inline `mustGetLogger`

ransomw1c:
could we update the doc strings to explicitly include a few common available unit strings (e.g. KB, MB, GB)?

opinion here is that `flagext` makes flags more usable only to the extent that users know how to user `flagext`.

cmd/datamon/cmd/flags.go
```diff
@@ -290,11 +295,30 @@ func addTargetFlag(cmd *cobra.Command) string {
 	return c
 }
 
+func addCacheSizeFlag(cmd *cobra.Command) string {
+	c := "cache-size"
+	datamonFlags.bundle.CacheSize = flagext.ByteSize(50 * units.MB)
+	cmd.Flags().Var(&datamonFlags.bundle.CacheSize, c, "The desired size of the memory cache used when streaming is enabled")
```
ransomw1c: could we update the doc strings to explicitly include a few common available unit strings (e.g. KB, MB, GB)?

opinion here is that `flagext` makes flags more usable only to the extent that users know how to user `flagext`.

cmd/datamon/cmd/mocks_test.go
```diff
@@ -78,10 +78,14 @@ func (a AuthMock) Principal(_ string) (model.Contributor, error) {
 	}, nil
 }
 
+func testContext() string {
```
ransomw1c: nit: cruft?

hack/fuse-demo/wrap_datamon.sh
```diff
@@ -412,7 +412,7 @@ while [[ -z $MOUNT_COORD_DONE ]]; do
             # mount point already found
             continue
         fi
-        if echo "$mount_data" | grep -q "^$mount_point fuse$"; then
+        if echo "$mount_data" | grep -q "^$mount_point fuse.datamon$"; then
```
ransomw1c: what happened here?

pkg/cafs/freelists.go
```diff
@@ -1,63 +1,362 @@
...
- */
+// BytesToBuffers returns the number of buffers for a given
+// size target in bytes, given a leaf size. Buffers are rounded up,
+// so we may exceed the target by at most the size of a leaf.
+func BytesToBuffers(target int, leafSize uint32) int {
+	return int(math.Ceil(float64(target) / float64(leafSize)))
+}
+
+// FreeList knows how to allocate buffers and release them
+type FreeList interface {
```
ransomw1c: vis one idea per commit:  i wouldn't mind seeing this freelist re-write as a separate commit.

it's not clear to me if we actually need all the complexity encoded here, yet i do want some of the improvements, and i definitely want to land the majority of this patch as a whole.

pkg/cafs/cafs.go
```diff
@@ -102,28 +64,56 @@ type Fs interface {
...
+
+	var err error
+	f.lru, err = lru.NewWithEvict(cacheBuffers, func(_ interface{}, lruVal interface{}) {
+		f.leafPool.Release(lruVal.(LeafBuffer)) // relinquish buffers to the freelist
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	f.keysCache, err = lru.New(f.keysCacheSize)
```
ransomw1c: let's either attach a freelist here or make a note to profile memory usage:  it could be that the `lru` causes GC to lose track of `alloc`s in all cases, or it could be that what's going on with the leaf buffers is unique to that cache.

pkg/cafs/cafs.go
```diff
@@ -132,54 +122,85 @@ type cafsStore struct {
...
-	prefix                      string
-	zl                          zap.Logger //nolint:structcheck,unused
-	l                           log.Logger //nolint:structcheck,unused
+	store    cafsStore
+	leafSize uint32
+	l        *zap.Logger
+
+	// prefix determines a namespace for keys
+	prefix string
+	pather func(Key) string // pathing logic
```
ransomw1c: i agree that specifying a procedure is the most generalized abstraction ... if it's at all possible to parameterize the procedural code in this package, i'd prefer to do that for now rather than pre-emptively use the most abstract form.

slight deja-vu here from #266 #267 and #268

pkg/cafs/cafs.go
```diff
@@ -132,54 +122,85 @@ type cafsStore struct {
...
-	defer w.Close()
+	d.l.Debug("Start cafs Put")
+	defer d.l.Debug("End cafs Put")
+
+	w := d.writer()
+
+	// write leaf blobs
 	written, err := io.Copy(w, src)
 	if err != nil {
+		w.Close()
```
ransomw1c: nit: prefer `defer w.Close()` + function scope ... so

```
	w := d.writer()

	// write leaf blobs
	written, err := io.Copy(w, src)
	if err != nil {
		w.Close()
		return PutRes{}, err
	}
	root, keys, err := w.Flush()
	if err != nil {
		w.Close()
		return PutRes{}, err
	}

	if err = w.Close(); err != nil {
		return PutRes{}, err
	}
```

becomes

```
        var (
root Key
keys []Key
)

	// write leaf blobs
err := func() {
	w := d.writer()
defer w.Close()
	written, err := io.Copy(w, src)
	if err != nil {
		return err
	}
	root, keys, err = w.Flush()
	if err != nil {
		return err
	}
return nil
}
if err != nil { return PutRes{}, err }
```

these are not exactly equivalent, yet i think the second is more idiomatic?

like, don't we usually cleanup in golang with `defer <cleanup>`?

or is it that the writer here has a non-idiomatic `Close()` method?

pkg/cafs/cafs.go
```diff
@@ -132,54 +122,85 @@ type cafsStore struct {
...
 	}
+
 	if err = w.Close(); err != nil {
 		return PutRes{}, err
 	}
-	destinations := make([]storage.MultiStoreUnit, 0)
 
-	found, _ := d.store.backend.Has(ctx, d.prefix+key.String())
+	// write the root key as a blob containing all leaf keys, if it does not exist already
+	found, _ := d.store.backend.Has(ctx, d.pather(root))
```
ransomw1c: in [my comment](https://github.com/oneconcern/datamon/pull/368/files#r370896993) on `defaultFs`, what i'm saying (i think – haven't reviewed entire patch yet) is why not replace `d.pather(root)` here with `lks.StringWithPrefix(root)`?

pkg/cafs/cafs.go
```diff
@@ -132,54 +122,85 @@ type cafsStore struct {
...
 	if !found {
```
ransomw1c: note to self:  changes in this block for readability only

pkg/cafs/hasher.go
```diff
@@ -0,0 +1,206 @@
+package cafs
```
ransomw1c: this is a rename of `helpers.go` + some re-factor that git is recording as a delete and new file.

would prefer review + commit history that records re-factoring as a diff on the file ... or a single commit that contains the changes to the hashing implementation only?

ransomw1c:
of the physical layout conventions so far, the `name.go` and `name_options.go` is my favorite. usually disinterested in physical layout.

pkg/core/bundle_read.go
```diff
@@ -14,56 +15,52 @@ func errNotEOF(err error) bool {
 	return err != nil && err.Error() != "EOF"
 }
 
-// ReadAt reads some bundle data with an optional offset. Streamed bundles ignore the offset.
-func (b *Bundle) ReadAt(file *fsEntry, destination []byte, offset int64) (int, error) {
+// ReadAt reads some bundle data with an optional offset
+func (b *Bundle) ReadAt(file *FsEntry, destination []byte, offset int64) (int, error) {
+	logger := b.l.With(
```
ransomw1c: this is s'th to start doing everywhere, i guess?

use the zap `With()` api to add locals to logger and log a start message before something that could possibly return an error that we need to log, then returning the wrapped error and not logging an error message before the return?

cc. @kerneltime @joaoh821C

pkg/core/fs_ro_ops.go
```diff
@@ -184,34 +186,59 @@ func (fs *readOnlyFsInternal) ReleaseFileHandle(
...
+	return
+}
+
+type fsNodeToAdd struct {
+	parentINode fuseops.InodeID
+	FsEntry     FsEntry
+}
+
+// FsEntry is a node in the filesystem
+type FsEntry struct {
```
ransomw1c: keep this type private?

pkg/core/fs_rw_ops.go
```diff
@@ -548,6 +547,170 @@ func (fs *fsMutable) ReleaseFileHandle(
...
+	}
+	fs.readDirMap[id][dirEnt.Inode] = dirEnt
+}
+
+func (fs *fsMutable) insertLookupEntry(id fuseops.InodeID, child string, entry lookupEntry) {
+	fs.lookupTree, _, _ = fs.lookupTree.Insert(formLookupKey(id, child), entry)
+}
+
+// Create a node. Need to hold the locks before calling.
+func (fs *fsMutable) createNode(lk []byte, parentINode fuseops.InodeID, childName string,
```
ransomw1c: another case where we're shuffling physical layout (move `createNode` here from `fs_common.go`) ... and possibly changing logic as well?  i cannot easily tell what changed b/c git is recording it as a rewrite.

pkg/core/fs_ro_ops.go
```diff
@@ -233,28 +260,47 @@ func (txns *populateFSTxns) commitToFS(fs *readOnlyFsInternal) {
...
-/* unwound recursion to build a list of ents terminating at the first extant parent */
+func newFSTxns(fs *readOnlyFsInternal) *populateFSTxns {
+	return &populateFSTxns{
+		dirStore:     fs.fsDirStore.Txn(),
+		lookupTree:   fs.lookupTree.Txn(),
+		fsEntryStore: fs.fsEntryStore.Txn(),
+	}
+}
+
+type populate struct {
```
ransomw1c: `populateFSAddBundleEntryParams`?

pkg/cafs/reader.go
```diff
@@ -201,115 +273,351 @@ func (r *chunkReader) WriteTo(writer io.Writer) (n int64, err error) {
...
 		nextIndex := index + 1
-		if nextIndex >= int64(len(r.keys)) {
-			return
+		if nextIndex >= len(r.keys) || nextIndex-initiator > r.maxFetchAhead {
+			// don't attempt to fetch farther ahead than permitted
+			return false
 		}
-		if r.lru.Contains(r.keys[nextIndex].StringWithPrefix(r.prefix)) {
-			return
+		if r.lru.Contains(r.pather(r.keys[nextIndex])) {
```
ransomw1c: again, suggest using `r.prefix` only, at least for now, until requirements necessitate configurable paths.

if deduplication (because the key here _must_ be the same as in the eviction callback) is the goal, i suggest naming a `func` at the package level and and using it in both places.

ransomw1c:
i'm going to have to split this review up over a couple of days..

.. have given a reasonably diligent first read to everything except cafs and fuse changes, and i have read the cafs changes in part, leaving a few comments.

the fuse changes are just logging and readability?

pkg/cafs/freelists.go
ransomw1c: afterthought

note on terminology:  freelist, alloc/free, and buffer pool are all words that we use to refer to approximately the same kind of memory management setup
* alloc/free - system-level memory management, either via standard C library or Go runtime internals.
  while here, note that it's also not altogther uncommon to write custom `alloc`/`free` routines on systems (e.g. plan9) that weren't written expressly to optimize memory management and that haven't been productionalized to the extent that linux, etc. have
* freelist -  `github.com/jacobsa/fuse/internal/freelist` is my original source for hearing about this.  having read some more about them, i take freelist to connote pretty much strictly a slice of some fixed type that's used to recycle buffers in the manner of the original implementation from `jacobsa/fuse`
* buffer pool - Ritesh's mention of this when i was looking into the memory usage was the first time i heard about it.  this is the usual term for golang libraries responsible for allocating and releasing buffers of a more arbitrary nature of a more arbitrary means

so i guess all i'd suggest doing here for now is renaming to buffer pool, and spinning this comment off into a separate issue to look into using a drop-in replacement later.

in the original implementation, i recall making the decision between using freelists and buffer pools with the intent of evolving the program by choosing a golang buffer pool library.  like, essentially i threw the freelist in place b/c i knew it solved the immediate error and didn't want to introduce an upstream dep without some investigation.

in the long run, as we discussed on the most recent call, i suspect (or perhaps hope is a better term) we'd be better off leveraging the [golang/linux] ecosystem than getting into the nitty-gritty of this kind of performance tuning as part of the datamon project, at least for this version.  like, there is a semi-serious thread of conversation about rust re-writes and such.

----

again, quite enthused about this patch.

.circleci/config.yml
fredbi: good catch. Moving this to config

.circleci/config.yml
fredbi: counter-nit: `go list` yields the number of packages

cmd/datamon/cmd/config.go
fredbi: > is mustGetLogger now used everywhere that a logger is initialized?
yes

> refactor types defining such dsl-ish things to a separate package
Yes. Indeed this is the goal I am pursuing. I'd like many little things like this (including the whole flagsT, with default settings and GCS strategy) to move to some `pkg/api/helpers` or `pkg/sdk/helpers` package to be available as an higher level API for instance, a python wrapper. Please suggest a name for that.

pkg/core/fs.go
fredbi: ok agree. Added the TODO (the `???`  is more your distinctive mark :) ) Defered.

cmd/datamon/cmd/cli_fuse_test.go
fredbi: this part has been a pain, really.
The way output is piped from the forked process looks like a bit dodgy because of the way the `cmd` implementation is copying the output... 

I've tried many combinations along the lines of what you are describing (TeeReader, MultiWriter, etc...). I admit there is still room for improvement here, and I don't have the answers to all the _whys_ in this section.

The main issue here was to be able to catch the _normally unbuffered_ stdout & stderr _before_ the `cmd.Wait()`. It turned out that an explicit copy to a pipe was needed to accomplish that. Any more elegant combination I've tried did actually fail to capture the output progress.

cmd/datamon/cmd/cli_fuse_test.go
fredbi: I am not sure about DATAMON_CONTEXT (since this is normally overridden with a flag in the tests to run next). I've to tidy up that a bit later on (we already have had a talk about propagating env variables).

 The main takeaway here is that we need to propagate the env to the forked process, esp. GOOGLE_APPLICATION_CREDENTIALS

cmd/datamon/cmd/config.go
fredbi: yep. typo. fixing that

cmd/datamon/cmd/context_create.go
fredbi: ok. Doing that

cmd/datamon/cmd/flags.go
fredbi: sure. Doing this.

hack/fuse-demo/wrap_datamon.sh
fredbi: >   core: set fuse moumt subtype to "datamon" and "datamon-mutable", so it appears as such in the output of the "mount" UNIX command

Experimenting things on my own box (which has a lot of other fuse mounts) gave me the idea of giving a proper signature to _our_ mounts. I then realized you had a strict check there, so updated this.

pkg/cafs/freelists.go
fredbi: > "one idea by commit" 

I generally agree with this principle and you know I have been trying to stick to this as much as I could.

Regarding this particular PR, however,  we are in a situation in which the change is driven by a user report such as "reading a big file on fuse RO mount is slow".

Besides, previous discussions regarding performances of the fuse mount led to trying to work on the memory footprint.

Addressing this issue required:
(i) to understand the whole chain of communication across several packages, for this specific use case -- that makes a lot of code, but this is really only covering the read of file on a RO fuse mount, -- improving readability was really required
(ii) to figure out many smaller issues, which were all contributing somehow to the degradation of performances, or to a degraded understanding of how this works - essentially this has driven toward injecting more debug logging everywhere (i'd like to revert on that later on, favouring  exported metrics  over debug traces),
(iii) to unroll a nasty yarn of fixes, as applying smaller fixes did in turn uncover hidden or latent bugs,
(iv) introducing/re-enacting ways to test things

fredbi: > in the original implementation, i recall making the decision between using freelists and buffer pools with the intent of evolving the program by choosing a golang buffer pool library. like, essentially i threw the freelist in place b/c i knew it solved the immediate error and didn't want to introduce an upstream dep without some investigation

I've tried to fix and improve the simplistic approach to freelists and cache buffers, without challenging the very existence of this feature.
Managing memory and caching FS is essentially a complex thing, which **cannot** be addressed by simple approaches.
I don't want to keep this in this state, quite the contrary.

Rather than finding a "pool library" I am more in favor of selecting a full-fledged embedded KV store to which we could delegate _all_ the internal caching and relieve pressure on memory (and memory management code...). My current pick is badgerdb. You may share your thoughts about this idea. That would be a drop-in replacement of the caching/pooling logics, but not the prefetching one.

Of course, the fuse cache tuning is also a possibility, that I would think as complementary.

pkg/cafs/cafs.go
fredbi: `defer` is indeed idiomatic. It is also slower (every go version comes with some improvements here, though).
But we also want to check the error returned by `w.Close` (which is important here, but only in the main path, not on other error handling).

Hence I specifically removed the `defer` to avoid closing twice, which could lead to unspecified results.

pkg/cafs/cafs.go
fredbi: I chose just to do the opposite, with `StringWithPrefix` method bound to disappear soon (with planned refactoring of deduplication scheme in a separate subpackage).

pkg/core/bundle_read.go
fredbi: yes I thought it was a good pattern for readability, trading the extra (minor) cost of storing this logger for less clutter in tracing.

pkg/core/fs_ro_ops.go
fredbi: no.
Since we expose `Bundle.ReadAt()` we need to expose the types of its parameters as well. Hence `FsEntry` and not `fsEntry`.
Aside: I'd really like to move all this to some `pkg/core/fuse` or `pkg/fuse` package. This would probably need to create a common testing utility pkg since testing fuse bends on a lot of the bundle testing code.

pkg/core/fs_ro_ops.go
fredbi: No I specifically don't want to adopt the `myLargerThanLifeTypeNaming` convention.

Similarly, I renamed many functions with long names.
You are welcome to suggest a more appropriate name, it only has to be short and meaningful 

Basically, this is the same trend as for package naming.
You might find this useful: https://blog.golang.org/package-names
https://rakyll.org/style-packages/

fredbi:
> the fuse changes are just logging and readability?
Mostly yes. There are some small changes, though:
- relinquish to GC the RadixTree used for fs construction after fs initialization
- explicitly set the ReadOnly: true as a fuse option
- sign fuse mount as `fuse.datamon` or `fuse.datamon-mutable` so this appear in `mount` commands
- stubbed `Xattr` methods to avoid spurious warnings when this feature is not going to be ever supported
- stubbed `Flush` method on Read-Only mount to avoid (unblocking) meaningless errors reported
- fixed wrong error reporting logic on on mutable mount on "...open file will retry this" warning
- fixed logic of the `isDir()` func, which was in fact doing `isNotDir`

cmd/datamon/cmd/bundle.go
fredbi: since concurrency is mostly driven by network consideration (bandwidth, latency), and not really driven by the capabilities of the local computer the program is running on, i don't think we can think of ever producing some automated factor, unless the execution is preceded by a brief benchmarking of gcs from the local machine.

pkg/cafs/freelists.go
fredbi: My next move on this is to conduct some research on supporting an alternate way of caching things, using badgerdb as KV store on local FS. My intention is to let the current mechanism as a viable alternative, then eventually drop it if the KV approach proves better in all use cases.

pkg/cafs/cafs.go
fredbi: The default reader takes  a nil instance of the freelist, which means it is just a wrapper around the gc, without any specific eviction mechanism. I am using this for testing (chunkReader testing has been augmented to cover several combinations of the options at hand). The default for the fs is to use a freelist with a high watermark tuned to the LRU size plus some read concurrency setting (which is grossly determined as for now)

pkg/cafs/cafs.go
fredbi: the original cafs code introduced the `pather` concept, which has been later scattered under different forms. What matters is that we have one single approach. I chose the func.

I believe that adopting prefixes to form namespaces for keys is something we will have to deploy rather sooner than later, as adoption will rapidly make our current implementation of key searches quite impractical

pkg/cafs/cafs.go
fredbi: follow-up note: disagree - won't fix

pkg/cafs/cafs.go
fredbi: follow-up note: disagree - won't fix

pkg/cafs/cafs.go
fredbi: that is correct

pkg/core/bundle_read.go
fredbi: follow-up note: comment only. No change requested

pkg/core/fs_ro_ops.go
fredbi: follow-up note: disagree - won't fix

pkg/core/fs_rw_ops.go
fredbi: yes that is unfortunate, but that's the way it goes when a repo has accumulated too much debt

pkg/core/fs_ro_ops.go
fredbi: follow-up note: disagree - won't fix

pkg/cafs/reader.go
fredbi: the key is not used atm in the eviction callback, just the address of the buffer

fredbi: follow-up note: disagree - won't fix

pkg/cafs/cafs.go
fredbi: follow-up note: disagree - won't fix

pkg/cafs/hasher.go
fredbi: yes that is unfortunate, but that's the way it goes when a repo has accumulated too much debt

fredbi: follow-up note: open discussion thread

pkg/core/fs_rw_ops.go
fredbi: follow-up note: open discussion thread

cmd/datamon/cmd/cli_fuse_test.go
fredbi: follow-up note: open discussion thread

cmd/datamon/cmd/cli_fuse_test.go
fredbi: follow-up note: open discussion thread

.circleci/config.yml
fredbi: follow-up note: comment only. No change requested

cmd/datamon/cmd/config.go
fredbi: follow-up note: comment only. No change requested

cmd/datamon/cmd/bundle.go
fredbi: follow-up note: comment only. No change requested

pkg/cafs/reader.go
fredbi: follow-up note: open discussion thread

pkg/cafs/freelists.go
fredbi: follow-up note: open discussion thread

pkg/cafs/cafs.go
fredbi: follow-up note: open discussion thread

cmd/datamon/cmd/cli_fuse_test.go
```diff
@@ -3,166 +3,331 @@
...
+	cmd.Env = append(cmd.Env, "DATAMON_GLOBAL_CONFIG="+os.Getenv("DATAMON_GLOBAL_CONFIG"))
+	cmd.Env = append(cmd.Env, "HOME="+os.Getenv("HOME"))
+	cmd.Env = append(cmd.Env, "PATH="+os.Getenv("PATH"))
+	cmd.Env = append(cmd.Env, "GOOGLE_APPLICATION_CREDENTIALS="+os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"))
+	pipeOut, err := cmd.StdoutPipe()
+	require.NoError(t, err)
+	pipeErr, err := cmd.StderrPipe()
+	require.NoError(t, err)
+
+	// combine stdout & stderr, tee this output to os.Stdout and return pipe reader for output scanning
```
kerneltime: Since we want to test if the output of a cli command is correctly being routed to STDOUT, we might want to keep the separation of what gets logged to which channel. Since operations such a list needs to be piped in shell to other unix tools, we have to add a test to make sure the output is to STD OUT.

cmd/datamon/cmd/cli_fuse_test.go
```diff
@@ -3,166 +3,331 @@
...
+	go func() {
+		_, _ = io.Copy(c, pipeOut)
+	}()
+	go func() {
+		_, _ = io.Copy(c, pipeErr)
+	}()
+	return cmd, pipeR
+}
+
+func TestFred(t *testing.T) {
```
kerneltime: Cleanup?

fredbi: yes sure. I forgot that one

cmd/datamon/cmd/cli_fuse_test.go
fredbi: ok will remove the combined output

docs/usage/datamon_bundle_mount_new.md
```diff
@@ -20,6 +20,7 @@ datamon bundle mount new [flags]
       --destination string   The path to the download dir. Defaults to some random dir /tmp/datamon-mount-destination{xxxxx}
   -h, --help                 help for new
       --label string         The human-readable name of a label
+      --loglevel string      The logging level (default "info")
```
kerneltime: Maybe the log level is best placed as a persistent flag under root?

cmd/datamon/cmd/cli_fuse_test.go
ransomw1c: ok, so what i'm taking away from this is that there's a buffer on stdout and stderr somewhere, and we need to flush it before `.Wait()`, and this ensures that the buffer gets flushed, whereas a `Copy` on `MultiReader` doesn't.

cmd/datamon/cmd/cli_fuse_test.go
ransomw1c: > The main takeaway here is that we need to propagate the env to the forked process

understood.

i think i realized this during the daemonization but forgot that it was also the issue with the fuse tests.. at time of writing i'd forgotten that the default wasn't to copy the current environment variables when calling `Exec()`.

hack/fuse-demo/wrap_datamon.sh
ransomw1c: i've since found the change to `core/fs.go`.

and i'm pretty sure the ci tests cover the pipeline tool's expectations.

in any case, i'll know to keep an eye out for this change in case it affects the Argo-ization anywhere else

pkg/cafs/freelists.go
ransomw1c: > selecting a full-fledged embedded KV store to which we could delegate all the internal caching and relieve pressure on memory (and memory management code...). My current pick is badgerdb.

> You may share your thoughts about this idea.

so much the better.  i mean, i know we definitely don't want to write our own caches and memory allocators.

the stronger opinion here to recycle badgerdb to replace both the cache _and_ memory allocator is all the better by me.

i'll leave this thread of discussion for now without spinning off additional issues.  thanks for the additional note on intent.

pkg/cafs/cafs.go
ransomw1c: np

pkg/core/fs_ro_ops.go
ransomw1c: > I'd really like to move all this to some `pkg/fuse` package.

creating a package for the filesystem implementation stuff has been a todo for a while, although idk if we have an explicit issue for it.

separating out the tests, specifically, would be much of the time cost.

pkg/cafs/cafs.go
ransomw1c: making a mental note to profile memory usage here in case it is an issue in order to resolve the thread

pkg/core/fs_rw_ops.go
kerneltime: I am open to dropping this functionality for now.

docs/usage/datamon_bundle_mount_new.md
fredbi: Yes agree: that is coming next. Deferred

pkg/core/fs_rw_ops.go
fredbi: Agree

pkg/core/fs_ro_ops.go
fredbi: Yes agree. Moving this to plg/fuse is going to be the next move.
This way we remove a lot of options from bundle, remove the bumdle.GetAt and move back FsEntry to unexported

cmd/datamon/cmd/cli_fuse_test.go
kerneltime: @fredbi I guess you need to push an update to clean this up.

cmd/datamon/cmd/cli_fuse_test.go
fredbi: @kerneltime after some more inspection, it so happens that I am asserting on both output... 
I am deferring the more accurate separate assertions on stderr and stdout as a TODO (see #374).
Deferred.

kerneltime:
Will merge this in once tests pass.

PR #370, opened 2020-01-27
Proposal for diamond workflow

kerneltime:


docs/porposals/diamond.md
```diff
@@ -0,0 +1,27 @@
+# Diamond workflow support.
+
+Diamond workflow allows for a pipeline to scale out the generation of a single bundle.
+
+The diamond workflow consists of the following steps.
+
+1. Before a diamond starts, Datamon SDK allows for the creation of an unique ID for a diamond.
+   1. This can be a KSUID.
+   2. The client is expected to use the KSUID at each step.
+   3. command: ```datamon bundle split initialize --repo <repo> --message```
```
fredbi: Since this is mostly about pre-uploading objects before actually committing, why don't we use the current CLI paradigm like: `bundle upload --no-commit`, which would upload, prepare some metadata (e.g. `bundles/{bundleID}/precommited/{diamon-id}/file-list.yaml`) without updating the bundle yet, the uncommitted uploads. Each participant to the diamond could thus upload things pre-emptively, at any time, possibly even several times without committing.
Eventually, the standard `bundle upload` 

Benefit:
* Client wouldn't have to remember the KSUID: pre-emptive upload implicitly all relate to the next commit
* We don't have to break up the process in an initialize and a split stage: clients are expected to `bundle upload --no-commit`, then we only expect some minimal coordination to issue a final `bundle upload`.
* Since the command is reused, messages and labels can be specified and merged by the eventual commit
* This way of doing thing may be used to "diamond" but also as a way to resume things whenever a failure occurs during an upload. We would have to modify a bit the current upload so it always starts by creating this "staging" and would resume automatically when restarted.
* Requirements on client coordination are less stringent: we can even let clients push intermediary commits (they would be very fast), provided datamon guarantees same the shuffling of metadata is atomic (e.g. a lock file is needed to move and merge the content of `bundles/{bundleID}/precommited` to `bundles/{bundleID}`

docs/porposals/diamond.md
```diff
@@ -0,0 +1,27 @@
...
+   
+2. Each node in a diamond generates a filelist that is stored in the versioned metadata folder.
+   1. Each filelist is given a client assigned ID. Example node1,node2,node3 or 1,2,3 or a UUID.
+   2. On retries the ID should be the same for a given node.
+   3. command: ```datamon bundle split upload --repo <repo> --split-id <id> --source <path>```
```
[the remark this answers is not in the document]
fredbi: NOTE: all kind of distributed commit management involves a "pre-commit" stage. The main takeaway from the most popular alg (2-phase-commit) is that the final commit requires all the pre-committed parts to be complete. In my opinion, this may be achieve with some kind of flag file in the staging metadata (e.g. `bundles/{bundleID}/precommit/{diamonID}/.ready`)

docs/porposals/diamond.md
```diff
@@ -0,0 +1,27 @@
+# Diamond workflow support.
```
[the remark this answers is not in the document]
kerneltime: thanks

docs/porposals/diamond.md
```diff
@@ -0,0 +1,27 @@
...
+
+Diamond workflow allows for a pipeline to scale out the generation of a single bundle.
```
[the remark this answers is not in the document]
kerneltime: Not exactly, the nodes occur in parallel not serial. Will clarify.

docs/porposals/diamond.md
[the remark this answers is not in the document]
kerneltime: I will add the backend layout I have in mind. Will add more details for failure handling.

docs/porposals/diamond.md
```diff
@@ -0,0 +1,27 @@
...
+   2. The client is expected to use the KSUID at each step.
+   3. command: ```datamon bundle split initialize --repo <repo> --message```
+   
+2. Each node in a diamond generates a filelist that is stored in the versioned metadata folder.
+   1. Each filelist is given a client assigned ID. Example node1,node2,node3 or 1,2,3 or a UUID.
+   2. On retries the ID should be the same for a given node.
+   3. command: ```datamon bundle split upload --repo <repo> --split-id <id> --source <path>```
+
+3. At the end the file lists are combined into a bundle with the commit message. 
+   1. The eventual bundle has a single namespace and if 2 nodes in the diamond have duplicate files, one will win.
```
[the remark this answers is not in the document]
kerneltime: For the second case, we want that behavior that the data is not detected and the file is written.
This is the case were 2 separate splits write the same file in the filesystem namespace of the bundle. Example: /some-folder-path/data.csv and during the commit we are generating the filelist. We do not want duplicate entries in the file list of the bundle. The proposal here is to create a unified list of all files in all splits and combining them. The alternative is that during download or read only file system mount we just pick a winner. The benefit of picking a winner at the download time of bundle is that the commit logic will be simple.

docs/porposals/diamond.md
```diff
@@ -0,0 +1,27 @@
...
+   3. command: ```datamon bundle split initialize --repo <repo> --message```
+   
+2. Each node in a diamond generates a filelist that is stored in the versioned metadata folder.
+   1. Each filelist is given a client assigned ID. Example node1,node2,node3 or 1,2,3 or a UUID.
+   2. On retries the ID should be the same for a given node.
+   3. command: ```datamon bundle split upload --repo <repo> --split-id <id> --source <path>```
+
+3. At the end the file lists are combined into a bundle with the commit message. 
+   1. The eventual bundle has a single namespace and if 2 nodes in the diamond have duplicate files, one will win. 
+    The other files including the winner will be stored in a special path in the bundle. (Proposal ./.<ID>/conflicts/<orig path>/file-suffix of client ID)
```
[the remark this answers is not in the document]
kerneltime: That is fine.. we are exposing the commit merge issues as an artifact in the merge itself and leaving it to the client to deal with it. This is due to the automated nature of Datamon's usage vs. the human centric model of git.

docs/porposals/diamond.md
kerneltime: how would we deal with concurrency? There can be multiple split uploads on going for the same repo. We do not have a parent child relationship between bundles. I see this feature more analogous to the multi part upload of S3. Also, the bundle bucket is expected to be write once and never update to guarantee no change in data.  We can discuss this more in person.

docs/porposals/diamond.md
```diff
@@ -0,0 +1,40 @@
+# Diamond workflow proposal.
+
+Diamond workflow allows for a pipeline to scale out the generation of a single bundle. There 
+is little or no co-ordination between the nodes. Each node uploads part of the bundle in the
+ filesystem namespace of the bundle. 
+
+The diamond workflow consists of the following steps.
+
+1. Before a diamond starts, Datamon SDK allows for the creation of an unique ID for a diamond.
```
galvare2: My assumption is that the unique ID is created by Datamon and returned (i.e. printed to stdout) by the CLI?

docs/porposals/diamond.md
```diff
@@ -0,0 +1,40 @@
...
+   3. command: ```datamon bundle split initialize --repo <repo> --message```
+   4. backend model: In the vmetadata bucket the following path will be created.
+    ```/splits/repo/<id>/split.yaml```
+   
+2. Each node in a diamond generates a filelist that is stored in the versioned metadata folder.
+   1. Each filelist is given a client assigned name. Example node1,node2,node3 or 1,2,3 or a UUID.
+   2. On retries the name should be the same for a given node.
+      1. A node on failure can rerun the same command, as long as the split-name is the same
+      the previous filelist contents will be superseded by the ones uploaded later.
+   3. command: ```datamon bundle split add --repo <repo> --split-id <id> --source <path>```
```
galvare2: How is the node name being specified by the client in this command? Shouldn't there be a `--node-name` argument?

kerneltime:
@snizamoc @galvare2 @fredbi @ransomw1c rendered view: https://github.com/oneconcern/datamon/blob/diamond/docs/porposals/diamond.md

docs/porposals/diamond.md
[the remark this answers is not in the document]
galvare2: I assume that it's okay to have files with the same name in different directories, right? I.e.:
`domain1/result.csv`
`domain2/result.csv`

docs/porposals/diamond.md
fredbi: Concurrency would be achieved by allowing each participant to write pre-commited resources to a unique location in the config (e.g. `bundles/{bundleID}/precommited/{diamon-id}/file-list.yaml`) then putting a `.ready` signal there when it is done with its upload operation.

Since the "bundle" is not yet committed maybe the path should be more like: 
bundles/{base bundle ID|"first"}/precommit/{diamond-id}/`
"first" here means that there is no bundle created in this repo yet.

However I admit that initiating several branches concurrently is not supported by my model. I have to think again.

docs/porposals/diamond.md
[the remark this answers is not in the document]
kerneltime: @galvare2 yes different file names in the same same directory is fine.

docs/porposals/diamond.md
kerneltime: @galvare2 yes it will be an ID to STDOUT at the end.

docs/porposals/diamond.md
kerneltime: ```split-id``` is the same as ```node-name``` will clean up nomenclature in the document.

docs/porposals/diamond.md
```diff
@@ -0,0 +1,40 @@
...
+   4. backend model: In the vmetadata bucket the following path will be created or rewritten.
+   ```/splits/repo/<id>/filelist-<split-id>-<ksuid>.yaml```. KSUID suffix generated at start of the
```
ransomw1c: `<id>` and `<split-id>` is always the same value?

if so, would you describe the tradeoffs between an always-consistent schema and duplicating the id in the schema?  (in github comments, i mean)

ransomw1c:
having read some of the description about the solution, it appears to me that this is a proposal moreover for

* datamon functionality a _performant_ diamond workflow
* incremental uploads

like, it's possible to implement diamond workflows independent of changing datamon at all, except they'd result in a lot of re-hashing, etc.

additionally, although the proposal doesn't immediately mention incremental uploads in the problems its trying to solve, 

> This command can also be reused to have incremental uploads where there is only one node in the split.

is mentioned midway through the description of the solution.

i suggest hoisting incremental uploads from the implementation details to the description of what this proposal allows.

docs/porposals/diamond.md
ransomw1c: whoops, see that this was described in `bundle split add`.  resolving this in favor of [gabe's comment](https://github.com/oneconcern/datamon/pull/370/files#r371436203O)

docs/porposals/diamond.md
```diff
@@ -0,0 +1,40 @@
...
+2. Each node in a diamond generates a filelist that is stored in the versioned metadata folder.
+   1. Each filelist is given a client assigned name. Example node1,node2,node3 or 1,2,3 or a UUID.
+   2. On retries the name should be the same for a given node.
+      1. A node on failure can rerun the same command, as long as the split-name is the same
+      the previous filelist contents will be superseded by the ones uploaded later.
+   3. command: ```datamon bundle split add --repo <repo> --split-id <id> --source <path>```
+   4. backend model: In the vmetadata bucket the following path will be created or rewritten.
+   ```/splits/repo/<id>/filelist-<split-id>-<ksuid>.yaml```. KSUID suffix generated at start of the 
+   add command allows for ordering of uploads from the same split.
+   5. This command can also be reused to have incremental uploads where there is only one node in the split.
```
ransomw1c: having read some of the description about the solution, it appears to me that this is a proposal moreover for

* datamon functionality to support a _performant_ diamond workflow with _conflict resolution_
* incremental uploads

like, it's possible to implement diamond workflows independent of changing datamon at all, except they'd result in a lot of re-hashing, etc.

additionally, although the proposal doesn't immediately mention incremental uploads in the problems its trying to solve, 

> This command can also be reused to have incremental uploads where there is only one node in the split.

is mentioned midway through the description of the solution.

i suggest hoisting incremental uploads from the implementation details to the description of what this proposal allows.

ransomw1c:
about inter- as well as intra-node resolution:

what happens upon `split add` getting two files at the same path with the same `--split-id`?

i'm ok with punting on exactly details until the implementation phase, provided this intra-node resolution situation isn't a conceptual barrier to the overall setup.

docs/porposals/diamond.md
kerneltime: Will defer this discussion.

docs/proposals/diamond.md
```diff
@@ -0,0 +1,39 @@
...
+   3. command: ```datamon bundle split initialize --repo <repo> --message```
+      1. Command will output to STDOUT the id for the diamond.
+   4. backend model: In the vmetadata bucket the following path will be created.
+      ```/splits/repo/<id>/split.yaml```   
+1. Each split in a diamond generates a filelist that is stored in the versioned metadata folder.
+   1. Each split is given a client assigned ID. Example split1,split2,split3.. or 1,2,3.. or a set of UUIDs.
+   2. On retries the split ID should be the same for a given split.
+      1. A node on failure can rerun the same command, as long as the split id is the same
+      the previous filelist contents will be superseded by the ones uploaded later.
+   3. command: ```datamon bundle split add --repo <repo> --split-id <client generated id> --source <path>```
```
fredbi: So i am taking it that `datamon bundle split add` is _actually_ uploading blobs, which will eventually result in a bundle. Correct?

Do we plan to provide the companion command `datamon bundle split remove` to drop/cancel  things? (that would prove quite tricky with our deduplicated blobs?

kerneltime:
@galvare2 @jakedsouza @snizamoc An alternate to the proposal was added which implies the usage a bit. Take a look.
@fredbi Let me know if I captured it correctly.
I also made come corrections and simplifications to the existing proposal.

docs/proposals/diamond.md
```diff
@@ -0,0 +1,39 @@
+# Diamond workflow support.
+
+Diamond workflow allows for a pipeline to scale out the generation of a single bundle. The individual nodes writes to 
+splits of the eventual bundle without any co-ordination between them. At the time of commit any conflicts will be
+dealt with. Conflicts can only occur if the same file in the namespace of the bundle is written to by more than one split.
+
+The diamond workflow consists of the following steps.
+1. Before a diamond starts, Datamon SDK allows for the creation of an unique ID for a diamond.
+   1. This can be a KSUID.
+   2. The client is expected to use the KSUID at each step.
```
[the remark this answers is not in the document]
kerneltime: Yes

docs/proposals/diamond.md
```diff
@@ -0,0 +1,73 @@
...
+    1. The eventual bundle has a single namespace and if 2 nodes in the diamond have duplicate files, one will win. 
+    The other files including the winner will be stored in a special path in the bundle. (Proposal ./.{ID}/conflicts/{orig path}/file-suffix of client ID)
+    2. command: ```datamon bundle diamond commit --repo {repo} --id {id} --message {message} --label {label}```
+    3. Optionally a commit can be set to fail on conflict.
+        1. command: ```datamon bundle diamond commit --repo {repo} --id {id} --message {message} --label {label} --no-confict``` 
+    4. A file is written to mark the completion of a diamond and the corresponding bundle ID is written. On failure 2 bundles can be created in the repo.
+    for the same diamond but only one will win closing the bundle. ```/diamonds/repo/{id}/bundle.yaml```
+## Other commands
+ 
+```datamon bundle diamond list --repo {repo}```
```
snizamoc: Will the diamond bundles also show up in `datamon bundle list --repo {repo}` since diamond bundles are a subset of generic bundles?

docs/proposals/diamond.md
```diff
@@ -0,0 +1,73 @@
...
+conflicts. 
+
+Pros:
+
+1. Client does not need to create or manage ```split-id```
+2. Easy of use (similar to 1)
+
+Cons:
+1. Datamon will loose the mapping of client side logic for splits to actual additions to a diamond. ```split-id``` allows
+the grouping of which files originated from which split. The client can log the output of the command to capture the details.
```
snizamoc: I believe skipping the client split-id generation is a good idea - this is a cleaner, more intuitive interface because the client will not need to be concerned with the inner workings of Datamon (`split-id`s to me seem like a tangential detail from the client's perspective - it is just concerned with adding (a series of) uploads and knowing that other nodes may be uploading to this bundle as well, without caring about what specific id it is within the group of nodes uploading to the bundle). This also makes sense because even in the original proposal, Datamon does not have a view of which uploads from a given node are retries and which are additions (since both cases generate a new KSUID for each upload), so Datamon would have to figure this out anyway by determining conflicts and resolving them for each node (correct me if I'm wrong). This alternate proposal just pushes that task fully out to Datamon so that it has to do this across all nodes. Now there may be a performance enhancement in Datamon having this mapping of which filelists came from which node, but that improvement would have to be quantified in order to determine that having this mapping really is advantageous.

docs/proposals/diamond.md
```diff
@@ -0,0 +1,73 @@
...
+
+### User generated diamond id
+
+Diamond ID is used to 
+
+1. Track splits that are grouped together
+2. Separate the namespace of the metadata to allow for concurrent diamonds to the same repo.
+
+Diamond ID can also be client generated which would fail if the same diamond ID is used more than once in the initialize step.
```
snizamoc: What is the advantage of having a user-generated diamond id?

kerneltime: As an application developer one can choose the scheme you want to use, as an example one can use pipeline name + run id, thus there would not be any need to figure out how to plumb the KSUID through the pipeline. each stage can calculate the same diamond ID. It is not fool proof

docs/proposals/diamond.md
kerneltime: Once a diamond is committed it is a normal bundle. A diamond bundle is an transient state, the final state is the a normal bundle

docs/proposals/diamond.md
```diff
@@ -0,0 +1,73 @@
...
+   3. command: ```datamon bundle diamon add --repo {repo} --id {ID} --split-id {client generated ID for each split in a diamond} --source {path}```
+   4. backend model: In the vmetadata bucket the following path will be created or rewritten.
+      ```/diamonds/{repo}/{id}/filelist-{split-id}-{ksuid}.yaml```. KSUID suffix generated at start of the 
+      add command allows for ordering of uploads from the same split.
+   5. This command can also be reused to have incremental uploads before a bundle is created.
+   6. The ordering of ```add``` from the same ```--split-id``` can be strongly ordered based on a design similar to WAL.
+      This is a nice to have addition and can be added later in the implementation.
+1. At the end the file lists are combined into a bundle with the commit message. 
+    1. The eventual bundle has a single namespace and if 2 nodes in the diamond have duplicate files, one will win. 
+    The other files including the winner will be stored in a special path in the bundle. (Proposal ./.{ID}/conflicts/{orig path}/file-suffix of client ID)
```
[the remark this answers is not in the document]
kerneltime: Yes, the definition of newest will be based on some scheme but might not agree with what a client thinks as newest.

docs/proposals/diamond.md
```diff
@@ -0,0 +1,73 @@
...
+      ```/diamonds/{repo}/{id}/diamond.yaml```   
+1. Each split in a diamond generates a filelist that is stored in the versioned metadata bucket.
+   1. Each split is given a client assigned ID. Example split1,split2,split3.. or 1,2,3.. or a set of UUIDs.
+   2. On retries the split ID should be the same for a given split.
+      1. A node on failure can rerun the same command, as long as the split id is the same
+      the previous filelist contents will be superseded by the ones uploaded later.
+   3. command: ```datamon bundle diamon add --repo {repo} --id {ID} --split-id {client generated ID for each split in a diamond} --source {path}```
+   4. backend model: In the vmetadata bucket the following path will be created or rewritten.
+      ```/diamonds/{repo}/{id}/filelist-{split-id}-{ksuid}.yaml```. KSUID suffix generated at start of the 
+      add command allows for ordering of uploads from the same split.
```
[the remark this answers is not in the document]
kerneltime: it would be unto the pipeline tool to generate the split id. It is not part of this specification.

docs/proposals/diamond.md
```diff
@@ -0,0 +1,73 @@
...
+   3. command: ```datamon bundle diamond initialize --repo {repo} --message```
+      1. Command will output to STDOUT the id for the diamond.
+   4. backend model: In the vmetadata bucket the following path will be created.
+      ```/diamonds/{repo}/{id}/diamond.yaml```   
+1. Each split in a diamond generates a filelist that is stored in the versioned metadata bucket.
+   1. Each split is given a client assigned ID. Example split1,split2,split3.. or 1,2,3.. or a set of UUIDs.
+   2. On retries the split ID should be the same for a given split.
+      1. A node on failure can rerun the same command, as long as the split id is the same
+      the previous filelist contents will be superseded by the ones uploaded later.
+   3. command: ```datamon bundle diamon add --repo {repo} --id {ID} --split-id {client generated ID for each split in a diamond} --source {path}```
```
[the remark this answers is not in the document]
kerneltime: thank you

docs/proposals/diamond.md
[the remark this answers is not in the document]
snizamoc: Got it

docs/proposals/diamond.md
snizamoc: Makes sense, thanks

docs/proposals/diamond.md
snizamoc: Got it. For the flood pipeline, we plumb the run id through the pipeline anyway after generating it once so having non-user generated KSUIDs would not be much of a disadvantage

docs/proposals/diamond.md
fredbi: @kerneltime this indeed reflects well our conversation and the ideas I pushed forward.
I am also in favor of removing the need for client to manage the split-id at the (reasonable) cost of adding some merge logic to handle incomplete `split add` commands.

docs/proposals/diamond.md
fredbi: @kerneltime @snizamoc once the runID fully becomes part of the datamon specification and metadata (for lineage tracking), we may plan to move further ahead and auto-select this. So adding files to the split could start without any preliminary stage and no explicit client coordination. After discussing that topic with Ritesh, we believe that this could be a nice addition, but we'd like to prove the other concepts before we move into that eventually.

docs/proposals/diamond.md
```diff
@@ -0,0 +1,39 @@
...
+
+Diamond workflow allows for a pipeline to scale out the generation of a single bundle. The individual nodes writes to 
+splits of the eventual bundle without any co-ordination between them. At the time of commit any conflicts will be
+dealt with. Conflicts can only occur if the same file in the namespace of the bundle is written to by more than one split.
+
+The diamond workflow consists of the following steps.
+1. Before a diamond starts, Datamon SDK allows for the creation of an unique ID for a diamond.
+   1. This can be a KSUID.
+   2. The client is expected to use the KSUID at each step.
+   3. command: ```datamon bundle split initialize --repo <repo> --message```
```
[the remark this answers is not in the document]
kerneltime: the message is to allow a client to attach the context, this can be useful to reverse engineer the source of a split (which pipeline)

docs/proposals/diamond.md
kerneltime: The metadata for split can be removed but the only safe way to remove blobs is via background garbage collection. Not making it part of this PR.

docs/proposals/diamond.md
kerneltime: Updated

docs/proposals/diamond.md
kerneltime: Updated, take a look.

fredbi:
@kerneltime let's merge that one and iterate starting from here. Before doing that, we could leave a trailing note to recap deferred/unaddressed comments by reviewers.

PR #371, opened 2020-01-27
test upload and download subsetting flags

ransomw1c:
add tests for `--files` flag on `bundle upload` and the `--name-filter` flag on `bundle download`

### DIFF
--- a/cmd/datamon/cmd/cli_test.go
+++ b/cmd/datamon/cmd/cli_test.go
@@ -307,6 +307,104 @@ func TestUploadBundle_filePath(t *testing.T) {
 	}, "observe error on bundle upload path as file rather than directory", true)
 }
 
+func getTempFile(t *testing.T, filename string) (*os.File, func()) {
+	if filename == "" {
+		filename = "datamon_test_temp_file"
+	}
+	tempDirPath, err := ioutil.TempDir("", "datamon-TestUploadBundleFilelist")
+	require.NoError(t, err, "create temp dir")
+	tempPath := filepath.Join(tempDirPath, filename)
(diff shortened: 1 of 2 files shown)


pkg/core/bundle_unpack.go
```diff
@@ -466,6 +465,7 @@ func downloadBundleEntries(ctx context.Context, bundle *Bundle,
 		bundle.l.Info("downloading bundle entries",
 			zap.Int("num", len(bundle.BundleEntries)))
 		for _, b := range bundle.BundleEntries {
+			var selectionPredicateOk bool
```
ransomw1c: at first tests were failing b/c i didn't notice that `z` matches `leafsize` 0_o .. made this change before realizing that i was debugging a "derp."

so this is for readability only, then.

cmd/datamon/cmd/cli_test.go
```diff
@@ -307,6 +307,104 @@ func TestUploadBundle_filePath(t *testing.T) {
 	}, "observe error on bundle upload path as file rather than directory", true)
 }
 
+func getTempFile(t *testing.T, filename string) (*os.File, func()) {
```
fredbi: nitpick/information sharing: on internal test functions, we may use `t testing.TB` interface instead. This allows test utility funcs to be reused by benchmarks.

pkg/core/bundle_unpack.go
fredbi: :+1:

cmd/datamon/cmd/cli_test.go
ransomw1c: noted

PR #373, opened 2020-01-28
Datamon migration job v1 -> v2

fredbi:
This PR provides a kubernetes job to migrate a v1 repo to v2.

This job is parameterized and deployed via helm. Each job applies to one repo.
A script is provided to deploy several such jobs, then delete the k8 resources after completion.

Completed jobs must be deleted explicitly after completion, using helm. This is on purpose, so logs can be checked on completed/error pods.

The job parameters cover the following use cases:

1. migrate a repo entirely
2. migrate a repo with a truncated history, from bundle id or label spec
3. migrate a single bundle or label in a repo
4. restart a failed migration job

Repo description, bundle ID, label and messages are preserved.

An extra tag may be added to mark the migrated repo.

Tested manually on a local minikube cluster and on gke_onec-co_us-west2-c_onec-dev.

See included README.md for usage.


k8s/migratev2/values.default.yaml
```diff
@@ -0,0 +1,68 @@
...
+  credential: /home/project/.config/gcloud/application_default_credentials.json
+
+# some default config to map as ${HOME}/.datamon2/datamon.yaml
+config2:
+  config: workshop-config
+  context: dev
+  credential: /home/project/.config/gcloud/application_default_credentials.json
+
+# the size of the volume claim to attach as download staging
+stagingSize: 2Gi
```
kerneltime: might be too small

fredbi: this is just a default value. Will take something higher like 10GB.  What do you suggest. Have to document this parameter too.

k8s/migratev2/scripts/migrate.sh
```diff
@@ -0,0 +1,102 @@
...
+fi
+echo "==================================="
+
+res=$(datamon1 repo get --repo ${REPO})
+if [[ $? -ne 0 ]] ; then
+  echo "Failed to get repo on V1 config"
+  exit 1
+fi
+# retrieve the original description of the repo
+description=$(echo $res) # trim space
```
ransomw1c: this is a neat bash-ism ... if i start using this, however, since the same syntax produces different results in different shells (contrast with `${(z)my_var}` or s'th that exists in zsh only and is an error elsewhere) i'm likely to forget it only works on bash.

this is not a blocker on review, just a note to myself to hopefully remember about this better.

k8s/migratev2/values.default.yaml
fredbi: set default to 10Gi

fredbi:
@kerneltime @ransomw1c finalized version with full history support and idempotent restart.
Tested manually, docker image on gcr built and pushed manually: I don't intend to make a CI for that.

kerneltime:
Taking look.

k8s/migratev2/migrator.Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ENV GO111MODULE on
+ENV GOBIN ${GOPATH}/bin
+RUN mkdir -p ${GOBIN} && mkdir -p /stage/usr/bin
+
+WORKDIR ${GOPATH}/src/github.com/oneconcern
+RUN git clone https://github.com/oneconcern/datamon datamon1 && cd datamon1 && git checkout v1.0.0
+RUN cd ${GOPATH}/src/github.com/oneconcern/datamon1/cmd/datamon && go build && upx datamon && mv datamon /stage/usr/bin/datamon1
+
+WORKDIR ${GOPATH}/src/datamon2
+RUN go get -tags bundle_preserve github.com/oneconcern/datamon/cmd/datamon@master
```
kerneltime: Why the tag?

k8s/migratev2/migrator.Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN upx $GOBIN/datamon && mv $GOBIN/datamon /stage/usr/bin/datamon2
+
+FROM base
+COPY --from=builder /stage/usr/bin/datamon1 /usr/bin/datamon1
+COPY --from=builder /stage/usr/bin/datamon2 /usr/bin/datamon2
+WORKDIR /home/project
+RUN mkdir -p /data
+ENV HOME /home/project
+ENV GOOGLE_APPLICATION_CREDENTIALS /home/project/.config/gcloud/application_default_credentials.json
```
kerneltime: Both V1 and V2 Should be able to use the service account for the pod, why mount this?

k8s/migratev2/migrator.Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ENV VERSION ${version}
+ENV GIT_COMMIT ${commit}
+ENV GIT_DIRTY ${dirty}
+
+ENV GO111MODULE on
+ENV GOBIN ${GOPATH}/bin
+RUN mkdir -p ${GOBIN} && mkdir -p /stage/usr/bin
+
+WORKDIR ${GOPATH}/src/github.com/oneconcern
+RUN git clone https://github.com/oneconcern/datamon datamon1 && cd datamon1 && git checkout v1.0.0
```
kerneltime: Download the release?

k8s/migratev2/migrator.Dockerfile
fredbi: I've constructed this specifcally to be enabled for this binary

k8s/migratev2/migrator.Dockerfile
fredbi: yes could do that. I was not so sure about the artifact... In doubt I've built from source.
Anyway, this is done once: I pushed the image on gcr and it probably won't be updated any time soon

k8s/migratev2/migrator.Dockerfile
fredbi: ok will experiment with the service account and push a fix

PR #375, opened 2020-01-29
CLI output for list, get should go to STDOUT.

kerneltime:

### DIFF
--- a/cmd/datamon/cmd/bundle.go
+++ b/cmd/datamon/cmd/bundle.go
@@ -36,7 +36,7 @@ func init() {
 	rootCmd.AddCommand(bundleCmd)
 
 	bundleDescriptorTemplate = func() *template.Template {
-		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}} {{printf "\n"}}`
+		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}}`
 		return template.Must(template.New("list line").Parse(listLineTemplateString))
 	}()
 }
--- a/cmd/datamon/cmd/bundle.go
+++ b/cmd/datamon/cmd/bundle.go
@@ -36,7 +36,7 @@ func init() {
 	rootCmd.AddCommand(bundleCmd)
 
 	bundleDescriptorTemplate = func() *template.Template {
-		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}}`
+		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}} {{printf "\n"}}`
 		return template.Must(template.New("list line").Parse(listLineTemplateString))
 	}()
 }
--- a/cmd/datamon/cmd/bundle_list.go
+++ b/cmd/datamon/cmd/bundle_list.go
@@ -4,8 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"log"
-
 	"github.com/oneconcern/datamon/pkg/core"
 	"github.com/oneconcern/datamon/pkg/model"
 
@@ -19,7 +17,7 @@ func applyBundleTemplate(bundle model.BundleDescriptor) error {
 		// by stopping upon errors while it was previously non-blocking
 		return fmt.Errorf("executing template: %w", err)
 	}
-	log.Println(buf.String())
+	logStdOut(buf.String())
 	return nil
 }
 
--- a/cmd/datamon/cmd/cli_test.go
+++ b/cmd/datamon/cmd/cli_test.go
@@ -156,13 +156,14 @@ func listRepos(t *testing.T) ([]repoListEntry, error) {
 	if err != nil {
 		panic(err)
 	}
-	log.SetOutput(w)
+	logStdOut = func(format string, a ...interface{}) (int, error) {
+		return fmt.Fprintf(w, format, a...)
+	}
 	//
 	runCmd(t, []string{"repo",
 		"list",
 	}, "create second test repo", false)
 	//
-	log.SetOutput(os.Stdout)
 	w.Close()
 	//
 	lb, err := ioutil.ReadAll(r)
@@ -375,12 +376,13 @@ func listBundles(t *testing.T, repoName string) (bundleListEntries, error) {
 	if err != nil {
 		panic(err)
 	}
-	log.SetOutput(w)
+	logStdOut = func(format string, a ...interface{}) (int, error) {
+		return fmt.Fprintf(w, format, a...)
+	}
 	runCmd(t, []string{"bundle",
 		"list",
 		"--repo", repoName,
 	}, "list bundles", false)
-	log.SetOutput(os.Stdout)
 	w.Close()
 	//
 	lb, err := ioutil.ReadAll(r)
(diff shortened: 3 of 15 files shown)


cmd/datamon/cmd/context.go
```diff
@@ -33,7 +33,7 @@ func init() {
 	rootCmd.AddCommand(ContextCmd)
 
 	contextTemplate = func() *template.Template {
-		const listLineTemplateString = `Model Version: {{.Version}}, Name: {{.Name}}, WAL: {{.WAL}}, ReadLog: {{.ReadLog}}, Blob: {{.Blob}}, Metadata: {{.Metadata}}, Version Metadata: {{.VMetadata}}`
+		const listLineTemplateString = `Model Version: {{.Version}}, Name: {{.Name}}, WAL: {{.WAL}}, ReadLog: {{.ReadLog}}, Blob: {{.Blob}}, Metadata: {{.Metadata}}, Version Metadata: {{.VMetadata}} {{printf "\n"}}`
```
fredbi: I think all this carriage return handling is not good: let's keep assuming we are using a standard logger which does that.
In a near future, I'd like template format to be a user-input (like say in docker CLI), so this will definitely surprise users.

cmd/datamon/cmd/bundle_list.go
```diff
@@ -19,7 +17,7 @@ func applyBundleTemplate(bundle model.BundleDescriptor) error {
 		// by stopping upon errors while it was previously non-blocking
 		return fmt.Errorf("executing template: %w", err)
 	}
-	log.Println(buf.String())
+	logStdOut(buf.String())
```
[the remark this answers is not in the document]
ransomw1c: > Shouldn't zap output go to stderr?

imo, what we're running into here is a limitation, great as it is, of the standard *nix facilities.  i would reframe the question:  shouldn't we have a a stdlog file handle?

and the fact is that we _don't_ have stdlog, so semantics of whether to log to error or out are up to us.

third option:  log to file.

----

also, we could log to file or stdio according to which command is being run.  `get` and `list` commands, for instance,  have their output written to stdio and logged to disk, while commands like `upload`, `download`, and `mount` might more sensibly log to stdio and record output to disk.

i think changing the output channels according to the meaning of the command is more rather that less surprising, although i wouldn't be surprised if this is not universal.

cmd/datamon/cmd/cli_test.go
```diff
@@ -156,13 +156,14 @@ func listRepos(t *testing.T) ([]repoListEntry, error) {
 	if err != nil {
 		panic(err)
 	}
-	log.SetOutput(w)
+	logStdOut = func(format string, a ...interface{}) (int, error) {
+		return fmt.Fprintf(w, format, a...)
```
ransomw1c:
```suggestion
		return fmt.Fprintf(w, format + "\n", a...)
```

cmd/datamon/cmd/cli_test.go
```diff
@@ -375,12 +376,13 @@ func listBundles(t *testing.T, repoName string) (bundleListEntries, error) {
...
```
ransomw1c:
```suggestion
		return fmt.Fprintf(w, format+"\n", a...)
```

cmd/datamon/cmd/context.go
ransomw1c:
```suggestion
		const listLineTemplateString = `Model Version: {{.Version}}, Name: {{.Name}}, WAL: {{.WAL}}, ReadLog: {{.ReadLog}}, Blob: {{.Blob}}, Metadata: {{.Metadata}}, Version Metadata: {{.VMetadata}}`
```

cmd/datamon/cmd/bundle.go
```diff
@@ -36,7 +36,7 @@ func init() {
 	rootCmd.AddCommand(bundleCmd)
 
 	bundleDescriptorTemplate = func() *template.Template {
-		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}}`
+		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}} {{printf "\n"}}`
```
ransomw1c:
```suggestion
		const listLineTemplateString = `{{.ID}} , {{.Timestamp}} , {{.Message}}`
```

ransomw1c:
to motivate suggestions regarding where to encode newlines/carriage returns:

these chars are used to separate list items.  they're encoded where the list is defined rather than where the items are defined.

cmd/datamon/cmd/bundle.go
kerneltime: We need the newline as part of the template.

cmd/datamon/cmd/cli_test.go
kerneltime: Why not just include it as a the template for CLI?

cmd/datamon/cmd/bundle_list.go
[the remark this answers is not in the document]
kerneltime: We need to refactor the CLI logging and output. 
We have 3 modes of operating.
1. STD OUT for piping to unix tools. This PR is trying to make sure our commands that need output piping from STDOUT work as per intuition.
2. Logging that integrates with automated log collection in workflows
3. Logging for a human that is using the CLI. Arguably we can improve the most here. the Upload and other commands that print to STDERR are mostly ignored by the developer and are hard to parse. 
I am open to suggestions, one proposal
1. STDOUT is a specification that is based on easy of scripting and cannot break
2. When used by a human we log to a file. Certain information such as deduplication rate, size of bundle pre and post deduplication should be part of the metadata of the bundle. Individual blobs uploaded can be a verbose output.
3. In workflows we log to STDERR (the pipeline tool will setup the CLI accordingly).

cmd/datamon/cmd/context.go
kerneltime: Hmm.. I will need to look into the templating by docker CLI, I guess there is a need for a graphQL equivalent for a CLI :-D

fredbi: docker commands have options like `--format '{{ .Image }}` and it's great to build scripts.

cmd/datamon/cmd/cli_test.go
fredbi: @kerneltime it just looks like you are reinventing the stdlib "log" facility.
The way to go is IMHO:
- create our own instances of `log.Logger`, let's say `outlog` and `errlog`, with `New(os.Stdout, ...)` and `New(os.Stderr, ...)`. These are the one that may be mocked.
- decide how to be consistent with zap loggers (a similar output setting is possible)
- use the loggers with the usual `Fprintf`, `Fatalf` etc,
- do not use `fmt.Fprintf` anywhere
- do not clutter every output with `+"\n"`

cmd/datamon/cmd/bundle.go
ransomw1c: i think [fred is also saying](https://github.com/oneconcern/datamon/pull/375/#discussion_r372628175) we need the newline as part of the output api, just within zap settings instead of a custom `fmt.Printf` call?

cmd/datamon/cmd/context.go
ransomw1c: > a graphQL equivalent for a CLI

what we're necessarily moving toward given that CLI is the user interface and we're implementing query capabilities.

i don't see this as a complexity explosion as of yet, and the motif of passing DSL as a string is the sort of escape hatch i'd suggest as well should the CLI wind up overly noisy/busy as a result.

cmd/datamon/cmd/bundle.go
kerneltime: Ok

cmd/datamon/cmd/cli_test.go
kerneltime: OK

PR #376, opened 2020-01-29
feat(upload): added option to preserve bundleID on upload

fredbi:
This option is interesting to migrate bundles from one store to another while preserving the initial ID.
Preservation allows for migration jobs to be idempotent and to restart on errors.

TODO(fred): tested manually so far, needs some U.T

### DIFF
--- a/cmd/datamon/cmd/bundle_preserve_disable.go
+++ b/cmd/datamon/cmd/bundle_preserve_disable.go
@@ -1,0 +1,6 @@
+// +build !bundle_preserve
+
+package cmd
+
+// enableBundlePreserve enables to have the --bundle in datamon upload and be able to rewrite history with ad'hoc bundle IDs
+const enableBundlePreserve = false
--- a/cmd/datamon/cmd/bundle_preserve_enable.go
+++ b/cmd/datamon/cmd/bundle_preserve_enable.go
@@ -1,0 +1,6 @@
+// +build bundle_preserve
+
+package cmd
+
+// enableBundlePreserve enables to have the --bundle in datamon upload and be able to rewrite history with ad'hoc bundle IDs
+const enableBundlePreserve = true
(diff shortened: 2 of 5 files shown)


ransomw1c:
i agree that this is the correct implementation of the described functionality.

moreover, i understand that the changes here are to be used to implement #373 .

an alternate way to get the desired functionality for #373 is to build a separate binary, or to use a build-time feature flag on this,
because my essential worry is that exposing this in the cli this might break some guarantees on integrity that're tacitly encoded elsewhere in the source: 

specifically, do we ever assume that 

> the sort order of ksuids is the same as the sort order of bundle creation by time?

so i'm at least delaying approval here to get confirmation that the above assumption doesn't exist anywhere.

fredbi:
@ransomw1c ksuid is built on top of a timestamp. The assumption holds down to the second afaik

fredbi:
> an alternate way to get the desired functionality for #373 is to build a separate binary, or to use a build-time feature flag on this,

ok I can do that, if this help resolve your qualms and allow to merge that more rapidly: the migration job is indeed using a dedicated docker image.

ransomw1c:
> ksuid is built on top of a timestamp

exactly.  so now that we're allowing to, in effect, pass that timestamp as a parameter, the order of ksuids might now be the same as the order of bundle creation times.

i'm trying to ask if any other parts of the codebase assume the sort order of ksuids is the same as the sort order of bundle creation by time?  because after this diff, it is not anymore.

i _suspect_ the answer to my question is "no, this patch is alright," and this can be approved by @kerneltime .

i'd prefer to approve the separate docker image, so it's easier to verify that i'm not approving a breaking change.

fredbi:
> i'm trying to ask if any other parts of the codebase assume the sort order of ksuids is the same as the sort order of bundle creation by time? because after this diff, it is not anymore.
Well sort of. Uploading this rewrites the history of the repo... Indeed you are inserting a bundle in the past... That's all about migration. I am adding a guard with a build-defined tag, so we won't let users do that.

fredbi:
@ransomw1c ok i've added a feature guard managed by a build flag.
This way, the behavior remains unchanged for users of the regular binary and I can build a migration-focused binary that knows how to rewrite history.

fredbi:
@ransomw1c @kerneltime  thanks

PR #372, opened 2020-01-28
init web tests

ransomw1c:
start smoke tests on `pkg/web`.

specifically, the goal here is to land a demo of how to test the `chi` router used to implement the web front.

notice in particular that `pkg/web`'s import of `pkg/core` is mocked out in test in order to test the web server implementation.

PR #378, opened 2020-01-29
mod tidy commit hook

ransomw1c:
go.mod [comment](https://github.com/oneconcern/datamon/pull/372#discussion_r372286671) from #372 

Signed-off-by: Ransom Williams <rwilliams@oneconcern.com>


ransomw1c:
i somehow had a `vendor.bkup/` in the git repo root directory that was causing `mod tidy` to error in #372 .. still unclear what was going on there.

anyway, with this change, `go mod tidy` is part of the pre-commit hooks.

unclear if it should be there or in ci (some discussion with @fredbi @joaoh821C here over the last couple weeks).

convinced that 
* the modules should be kept tidy
* without programmatic tidying, the modules will not be tidy

fredbi:
It belongs to the realm of git hooks. If you want to do that on CI (i.e. enforce this server side), you need to setup hooks on github to drive CI. This is _much_ more complicated than the easy going githook, which relies on developer's good will not to bypass the hooks (... most of the time :) )

PR #381, opened 2020-02-01
test(cli): added CLI test to cover workshop sessions items

fredbi:
This unrolls the workshop entirely, including `brew install datamon` (with `linuxbrew`).

Not covered:
- web browser launch (covered by other tests)
- self upgrade (covered by other tests)
- context creation (covered by other tests)
- RW fuse mount (covered by other tests)

**NOTE on this for the future**: struggled with two pain points I wouldn't have expected at first:
1. non-interactive install of brew on CI has been a pain (trying  many hacks and stuff, eventually it works...)
2. some tests run os.Setenv(), which in turn pollute some other ones depending on the order of execution => need to rethink a bit our common launch pad mockery

### DIFF
--- a/cmd/datamon/cmd/cli_workshop_test.go
+++ b/cmd/datamon/cmd/cli_workshop_test.go
@@ -1,0 +1,494 @@
+// +build fuse_cli
+
+package cmd
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	gcsStorage "cloud.google.com/go/storage"
+	"github.com/oneconcern/datamon/internal"
+	"github.com/oneconcern/datamon/pkg/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+)
+
+type testFile struct {
+	data                        []byte
+	size                        int
+	listed, downloaded, mounted bool
+}
+
+// TestWorkshop goes through all the steps in our workshop.
+//
+// Test is run with spawned process on freshly built datamon binary, not test goroutine: no mocks, no tricks
+func TestWorkshop(t *testing.T) {
+	logger := testLogger(t)
+	logger.Info("test initialization")
+	repo := generateRepoName("workshop")
+	targetBinary := buildTestBinary(t)
+
+	// create config
+	cmd, _ := testCommand(t, false, false, targetBinary,
+		"config",
+		"set",
+		"--config", "workshop-config",
+		"--credential", os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"),
+	)
+	output, err := cmd.CombinedOutput()
+	require.NoErrorf(t, err, "could not create config: %s", string(output))
+	// TODO(fred): should separate assertion on stdout and stderr here
+	assert.Contains(t, string(output), "config file created in")
+
+	// obtain available contexts
+	logger.Info("listing contexts")
+	cmd, _ = testCommand(t, false, false, targetBinary,
+		"context",
+		"list",
+	)
+
+	output, err = cmd.CombinedOutput()
+	require.NoErrorf(t, err, "could not list contexts, got: %s", string(output))
+	// TODO(fred): should separate assertion on stdout and stderr here
+	assert.Contains(t, string(output), "[dev]")
+
+	// create repo
+	cmd, _ = testCommand(t, false, false, targetBinary,
+		"repo",
+		"create",
+		"--repo", repo,
+		"--description", "workshop CI test repo",
+	)
+	output, err = cmd.CombinedOutput()
+	require.NoErrorf(t, err, "got: %s", string(output))
+
+	defer deleteRepoMeta(t, repo)
+
+	// list repos
+	cmd, _ = testCommand(t, false, false, targetBinary,
+		"repo",
+		"list",
+	)
+	output, err = cmd.Output()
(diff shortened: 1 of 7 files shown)


cmd/datamon/cmd/cli_workshop_test.go
```diff
@@ -0,0 +1,494 @@
...
+type testFile struct {
+	data                        []byte
+	size                        int
+	listed, downloaded, mounted bool
+}
+
+// TestWorkshop goes through all the steps in our workshop.
+//
+// Test is run with spawned process on freshly built datamon binary, not test goroutine: no mocks, no tricks
+func TestWorkshop(t *testing.T) {
```
kerneltime: Test Setup and Clean up can be simplified and include config and context creation.
I understand this is a script vs individual tests.
Might make sense to move the workshop to docs at some point.

kerneltime:
Will wait for Ransom to review as well.

cmd/datamon/cmd/cli_workshop_test.go
fredbi: Created issues for that in github

cmd/datamon/cmd/cli_workshop_test.go
```diff
@@ -0,0 +1,494 @@
...
+	name := internal.RandStringBytesMaskImprSrc(10)
+	data := []byte{} // empty file
+	err := ioutil.WriteFile(filepath.Join(pathToData, name), data, 0644)
+	require.NoError(t, err)
+	files[name] = &testFile{data: data, size: 0}
+	return files
+}
+
+func TestWorkshopBrewInstall(t *testing.T) {
+	if runtime.GOOS == "linux" { // TODO: don't know where brew installs things on OSX
```
ransomw1c: > where brew installs things on OSX

```
~% whence datamon2
/usr/local/bin/datamon2
~% ls -l /usr/local/bin/datamon2
lrwxr-xr-x  1 ransomwilliams  admin  42 Jan 14 10:56 /usr/local/bin/datamon2 -> ../Cellar/datamon/2.0.0-beta6/bin/datamon2
```

of course, so `/usr/local/Cellar/`?

except i suspect the question is more nuanced that i'm representing it.

cmd/datamon/cmd/cli_workshop_test.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+	name := internal.RandStringBytesMaskImprSrc(10)
+	data := []byte{} // empty file
+	err := ioutil.WriteFile(filepath.Join(pathToData, name), data, 0644)
+	require.NoError(t, err)
+	files[name] = &testFile{data: data, size: 0}
+	return files
+}
+
+func TestWorkshopBrewInstall(t *testing.T) {
```
ransomw1c: this test seems kind of loopy:

let's say the test fails because the release mechanism isn't working.  now, to fix the release mechanism, we need to make a change to the source.  we can't make the change to the source while ci is failing.

so if this test fails, i guess the fix always involve at least two merges:
* one to skip this test and get the release mechanism fixed
* one to re-enable the test

cmd/datamon/cmd/cli_workshop_test.go
fredbi: @ransomw1c well I don't know really how brew runs on MacOS. I've installed it on linux (specifically for datamon, since I am not a regular user of brew) and I designed this test to be executed by CI, which runs brewlinux. I think this is _reasonably_ close to the targeted audience. However, I cannot tell how this works on Mac (e.g. if you run go -test locally) hence I disabled the special path resolution we have on linux distros (where basically brew installs its wares on some home directory, and not in /usr). If this the command is by default in the path on mac, then my code should work also on mac.

cmd/datamon/cmd/cli_workshop_test.go
```diff
@@ -0,0 +1,494 @@
...
+		t.Logf("brew not installed. Skipping test")
+		t.SkipNow()
+	} else {
+		t.Logf("using brew: %s", string(output))
+	}
+
+	output, err := exec.Command("brew", "tap", "oneconcern/datamon").CombinedOutput()
+	require.NoErrorf(t, err, "could not run brew tap: %v:\n%s", err, string(output))
+
+	output, err = exec.Command("brew", "install", "datamon").CombinedOutput()
```
fredbi: @kerneltime @ransomw1c you may notice here that the install here works as initially advertised and not using brew install datamon/datamon. Perhaps this is a specificity of linuxbrew.

cmd/datamon/cmd/cli_workshop_test.go
fredbi: Well this is mitigated in part because it attempts to install the _previous_ release, the one that is already in place and supposed to be working. If the current  dev breaks the next release we won't be able to see it with this test.

Indeed, if we get to break things this will start failing (similar issue with self_update_test.go btw) we need to disable the test then to reenable the test.

This is the half empty bottle.

The half full bottle indicates that we have automated CI test that validate our releases can be readily installed by users...

cmd/datamon/cmd/flags.go
```diff
@@ -432,8 +435,21 @@ func paramsToDestStore(params flagsT,
 	return destStore, nil
 }
 
-func paramsToContributor(_ flagsT) (model.Contributor, error) {
-	return authorizer.Principal(config.Credential)
+func paramsToContributor(flags flagsT) (model.Contributor, error) {
```
ransomw1c: this appears to be the change in functionality that need land for v2 ..

..is it a workaround for
> some tests run os.Setenv(), which in turn pollute some other ones depending on the order of execution
or a change in the functional design of the program or both?

fredbi: work around. Also reflected on one of your (now famous) "???" in your latest PR regarding the env setup here. 
I don't think I really _changed_ the functionality... I made the influence of the env var more explicit and the priorities clearer.

PR #382, opened 2020-02-01
feat(cli): custom format templating for output

fredbi:
This is inspired by the set of docker commands (e.g. docker ps, etc), which
allow the user to specify the desired output.

This makes the tool much more flexible and easier to use when building scripts,
e.g. making the below shell constructs possible:

`bundleID=$(datamon bundle upload ... --format '{{.BundleID}}'`

`datamon bundle list --repo test-fred --format '{{.ID}}'|while read bundleID; do process(${bundleID});done`

`stagingSize=$(datamon bundle list files --bundle 1X1z123PEygTZ92ajbmQwPb9EHz --repo test-fred  --format '{{.Size}}'|jq -s 'add')`


fredbi:
> Need to fix the path for the generated docs.

Ah crap! I messed up when doing this manually. Fixed.

docs/datamon.md
```diff
@@ -0,0 +1,40 @@
+**Version: dev**
```
kerneltime: Why this new file?

kerneltime: This file needs deleting

fredbi: Yes. Done. Good catch.

ransomw1c:
recording (non-blocking) opinions here for the record:

i think that for the several use-cases here (`list` and `get`), what we're actually interested in is having access to a mapping type (name to value) rather than non-homogenous lists (current csv output includes several types of values in each line).  being able to access the struct with templates as implemented here is one way to get at name->value.  another way to is to serialize output as a mapping type (yaml, json, etc.) rather than csv.

my initial qualm here was that we're exposing something (the field names of the `pkg/model` structs) that was previously an internal implementation detail.  the answer here is that changes to `pkg/model` structs are expected to precipitate a major version bump in all cases, anyway, so we needn't worry about facading them away from the ui.

fredbi:
@ransomw1c  Point  taken. My opinion here is that the model _is_ the facade.

python SDK is going to go json instead to convey data internally down to python objects.

Having worked on the migration problem, I sure don't want this to occur again. Forthcoming changes in the metadata model will have to come sided with a migration procedure to update metadata (yes even when reading...).

Issue #374, opened 2020-01-28
forthcoming refactoring actions, following review of #368


fredbi:
Reviewers comments from #368, left as TODOs, in a nutshell:

* [ ] Confusing/pointless use of type alias for bool
> My point was just to replace this by:
`func MultiPut(ctx context.Context, stores []MultiStoreUnit, name string, buffer []byte, doesNotExist bool) error`
* [ ] separate porcelain funcs from cmd
>   refactor types defining such dsl-ish things to a separate package
> Yes. Indeed this is the goal I am pursuing. I'd like many little things like this (including the whole flagsT, with default settings and GCS strategy) to move to some pkg/api/helpers or pkg/sdk/helpers package to be available as an higher level API for instance, a python wrapper. Please suggest a name for that.
* [ ] `func localPath(consumable fmt.Stringer) (string, error)`
> functions or methods defined in storage
remains a questionable design decision
* [ ] cli_fuse_test.go: `func testContext() string`
 cruft?
* [ ]     I'd really like to move all this to some pkg/fuse package.
> creating a package for the filesystem implementation stuff has been a todo for a while, although idk if we have an explicit issue for it.
* [ ] Maybe the log level is best placed as a persistent flag under root?

PR #383, opened 2020-02-01
chore(storage): smallish refactoring series, part 1

fredbi:
Removed aliased type to bool.

Go aliased types (i.e. type A = B) are a syntactic sugar only.
This is different from type redefinitions (i.e. type A B), which create a new type for the compiler.

The original intent was to construct some kind of enumerated type, for a... boolean.

I chose to remove all this and just go back to plain bool.

* contributes #374

### DIFF
--- a/pkg/storage/gcs/store.go
+++ b/pkg/storage/gcs/store.go
@@ -173,7 +173,7 @@ func (rc readCloser) Close() error {
 	return nil
 }
 
-func (g *gcs) Put(ctx context.Context, objectName string, reader io.Reader, newObject storage.NewKey) (err error) {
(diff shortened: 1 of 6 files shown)


kerneltime:
The goal was to force use of the named values vs. true and false which require jumping around in code to know what it means.. I am ok with a builder pattern to executes similar to the google sdk

fredbi:
I understood the goal, and enforcement is not achieved with aliased types.
If you want an enum (well, for true, false, ...) you need  to redefine it this way `type KeyX bool`.

I feel that the choice of wording in the constants was at best misleading, and ended up with something _less_ readable than the plain `overwrite bool`.

PR #385, opened 2020-02-01
refact(fuse): move fuse to standalone package

fredbi:
# refact(fuse): move to standalone package

* no functional change or fix - caring about my favorite reviewers ... :smiling_imp: 
* core tests and fuse tests refactored to delegate part of the fixture building to "mocks" packages
* trim down core from options that are FS specific
* core.Bundle.ReadAt is removed and replaced by its private implementation in pkg/fuse
* introduced new Mountable interface (not in use yet)
* added RO mount unit test to cover cases with and without streaming


fredbi:
@ransomw1c the key change here is that bundle no more control the "streaming" setting: fuse mount objects do.

PR #390, opened 2020-02-03
[v1] fix download name filter

ransomw1c:
Signed-off-by: Ransom Williams <rwilliams@oneconcern.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,5 +33,7 @@ linters:
     - unparam
     - wsl
     - gocognit
+    - gomnd
+    - goprintffuncname
 
 max-same-issues: 0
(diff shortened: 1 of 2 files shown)


kerneltime:
@ransomw1c can you test this manually and cut a release for v1?

kerneltime:
Do fix the linting issue either via linter config or code change.

ransomw1c:
> why the filter may return an error?

firstly, i'd prefer to call this "the predicate" since a filter typically operations on collections (slices of strings), usually according to a predicate that operates on individual items but not always.

one way or another, it's a design decision:  predicate/filter operations certainly might include error-producing computations.  returning an error allows _the implementation to decide_ what actions to take, such as
* stop the entire download
* silent skip on error
* logged skip on error
* no skip on error
otherwise, it's up to _the predicate to decide_ what to do on errors.

so my read on the design decision here is that we want to retain more specification within the internal implementation rather than the IoC-mechanism where it's up to the predicate to recover for computation errors.

PR #391, opened 2020-02-04
Always let SDK try to get principal

kerneltime:
@fredbi the SDK searches in more places than the logic you put in there. If nothing else let SDK try.

### DIFF
--- a/cmd/datamon/cmd/flags.go
+++ b/cmd/datamon/cmd/flags.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"strings"
 
 	context2 "github.com/oneconcern/datamon/pkg/context"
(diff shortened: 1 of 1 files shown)


cmd/datamon/cmd/flags.go
```diff
@@ -454,10 +454,6 @@ func paramsToContributor(flags flagsT) (model.Contributor, error) {
 	default:
 		credentials = os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
 	}
-	if credentials == "" {
-		return model.Contributor{},
-			fmt.Errorf("could not resolve credentials: must be present as --credential flag, or in local config or as GOOGLE_APPLICATION_CREDENTIALS environment")
-	}
```
fredbi: @ritesh yes I guess you are right. I am curious about what the other ways are.

fredbi:
@kerneltime please PGP sign your commits

cmd/datamon/cmd/flags.go
kerneltime: SDK also looks in the ~/.config folder. So the cred file setup is not required if you have setup application default creds file in ~/.config. I will drop the environment check from the function as well.

PR #393, opened 2020-02-04
WIP: Feat python bindings

fredbi:
First stab at this feature and POC

HOW-TO
* see pkg/sdk/python
* make
* ./test-sdk.sh

Let's discuss how we go from here...


pkg/sdk/python/Makefile
```diff
@@ -0,0 +1,17 @@
+default: all
+.PHONY: clean all default
+# TODO(fred): version lib
+all: datamon.so
+# Define this in build environment (2.7 is default)
```
ransomw1c: data science has upgraded to python3, so i suggest we default to 3 as well if/when it's time to move this beyond PoC, ideally 3.6+ in order to allow some recent-ish niceties available in Python's `asyncio` stdlib.

ransomw1c:
whoahkay, i see this is going the [pure C extension](https://docs.python.org/3/extending/building.html#building) route rather than [ctypes](https://docs.python.org/3/library/ctypes.html), which is what i probably would've used based on familiarity.

i agree that, since datamon is primarily a golang project that's exporting a python api rather than a python project that needs some optimization in C, the pure C extension route is more appropriate.

the very most immediate priorities aside, i _am_ quite glad to see this boilerplate appear and suppose we can productively land this as a marker for decision-making on the pure C or ctypes question.

fredbi:
here some sample output:
```
from datamon import error, listRepos
import json
from dumper import dumpdef Repos():
  config = '{}'
  return json.loads(listRepos(config))if __name__ == '__main__':
  repos = Repos()
  dump(repos);

(edited)
5:15

<list at 0x7f0661573a50>:
  0: <dictionary at 0x7f065c2ed4d0>
    contributor: <str at 0x7f065c2efe30>: "{u'name': u'amarkovits@oneconcern.com', u'email': u'amarkovits@oneconcern.com'}"
    timestamp: u'2020-01-19T12:34:45.06118-08:00'
    name: u'alexm-test-repo'
    description: u'Create a test repo'
  1: <dictionary at 0x7f065c2ed710>
    contributor: <str at 0x7f065c2e65e0>: "{u'name': u'Fr\\xe9d\\xe9ric Bidon', u'email': u'frederic@oneconcern.com'}"
    timestamp: u'2020-01-24T16:25:58.006466117+01:00'
    name: u'fred-test-big'
    description: u'golang distribution'
  2: <dictionary at 0x7f065c2ed950>
    contributor: <str at 0x7f065c2e65e0>: "{u'name': u'Fr\\xe9d\\xe9ric Bidon', u'email': u'frederic@oneconcern.com'}"
    timestamp: u'2020-01-02T15:30:34.176863892+01:00'
    name: u'fred-test-repo'
    description: u'Create a test repo for fred'
  3: <dictionary at 0x7f065c2edb90>
    contributor: <str at 0x7f065c2e65e0>: "{u'name': u'Fr\\xe9d\\xe9ric Bidon', u'email': u'frederic@oneconcern.com'}"
    timestamp: u'2020-01-13T10:50:56.005799247+01:00'
    name: u'fred-test-repo-beta6'
    description: u'another round of tests'
  4: <dictionary at 0x7f065c2eddd0>
    contributor: <str at 0x7f065c2efe30>: "{u'name': u'jmachadosilva@oneconcern.com', u'email': u'jmachadosilva@oneconcern.com'}"
    timestamp: u'2020-01-27T12:36:02.099869+01:00'
    name: u'josh-test-repo'
    description: u'create test repo'
  5: <dictionary at 0x7f065c2f4050>
```

5:15
so the proof of concept is established. Now for the complete API...
5:18
@rwilliams in order to complete this, I'll dearly need that we (you, me, both?) complete the CLI config delegation thread we have started to discuss over your latest PR (my POC is still hardcoding all flags...)
5:20
NOTE(1): if some are interested I have all the material at hand now to make the very same wrapper for ruby and nodeJS  :slightly_smiling_face:
NOTE(2): this works for python2 and python3 just as well, although the binary distributions are separate (edited)

pkg/sdk/python/Makefile
fredbi: yeah it also works with python3, but this needs a different build

fredbi:
@ransomw1c I did not want to go ctypes since I need a way to free the objects allocated by the go runtime and copy them to objects managed by python. The flip side to embedding go is that you end up doing quite a bit of memory shuffling to avoid leakage.

rwilliams:
I got subscribed to this somehow.

jakedsouza:
thanks for this prototype @fredbi 

My main use case when I discussed with Ritesh were fairly simple
- be able to download, upload, possibly list bundles via python (either in local jupyter notebooks or in production)
- to use instead of the sidecar directly within the application itsself
- be able to pip install it and upgrade if and when datamon updates

fredbi:
ok @jakedsouza i shall prioritize the port  in that order, then.

My main objective here is to showcase the SDK use case, which means, eventually that:
- the main functionality resides in a datamon API, accessible either from CLI or SDK
- this API may be called directly from python programs, or (my take here) from some lightweight wrapper that takes care of the json marshalling/unmarshalling of parameters and results. My intention is thus to expose 2 modules: e.g. `from datamon import *` (params & results as python objects) and `from datamonraw import *` for the raw bindings. I am still conducting some research on the python ABI to figure out how I could reduce this to 1 single layer.
- this API may be supplemented by some extra helper porcelain features, which may be implemented/contributed either in go or in python (ex: display labels in line with bundles)
- all configuration (equivalent to the CLI flags) is accessible from the SDK

To attain that goal, a big shift is needed in the way datamon is handling parameters. That is on the way...

Also please note that the go bindings that I made may be exposed to any language with support for C/C++ ABI: we are by no means limited to python.
I've used previously the very same mechanism to demonstrably convey a go lib to ruby and nodeJS (in some older experiment, I also did that for perl and before that for COBOL...).

kerneltime:
please add a note in the read me for the SDK and say it is a PoC and if anyone wants they can leave feedback or request change or contribute...

kerneltime:
You can merge it in once you update the readme

PR #394, opened 2020-02-04
[wip] config cli test

ransomw1c:
...

### DIFF
--- a/cmd/datamon/cmd/cli_test.go
+++ b/cmd/datamon/cmd/cli_test.go
@@ -30,6 +30,8 @@ const (
 	destinationDir    = "../../../testdata/cli"
 	sourceData        = destinationDir + "/data"
 	consumedData      = destinationDir + "/downloads"
+	configDirectory = destinationDir + "/config"
+	configPath = configDirectory + "/" + configFile + ".yaml"
 	repo1             = "test-repo1"
 	repo2             = "test-repo2"
 	timeForm          = "2006-01-02 15:04:05.999999999 -0700 MST"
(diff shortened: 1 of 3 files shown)


cmd/datamon/cmd/cli_test.go
```diff
@@ -30,6 +30,8 @@ const (
 	destinationDir    = "../../../testdata/cli"
 	sourceData        = destinationDir + "/data"
 	consumedData      = destinationDir + "/downloads"
+	configDirectory = destinationDir + "/config"
+	configPath = configDirectory + "/" + configFile + ".yaml"
```
fredbi: ok. Just for you to know, this I believe better captures the intent (for future refactoring actions):
```
configPath = filepath.Join(configDirectory, configFile+".yaml")
```

cmd/datamon/cmd/config.go
```diff
@@ -26,8 +26,9 @@ const (
 	configFile = "datamon"
 )
 
+/* ??? how to override with viper and/or what does overriding a function with viper mean? */
```
fredbi: please remove this question and add a note with the answer:
> override is managed here explicitly from env (DATAMON_CONFIG).
> viper config resolver holds alternate locations when this environment is not set:
> https://github.com/oneconcern/datamon/blob/aa68e94a8a53f7136007762eaaf2e733d1208b5b/cmd/datamon/cmd/root.go#L92

cmd/datamon/cmd/cli_test.go
ransomw1c: this is, in part, me being periodically confused about `const` in golang as distinct from C

ransomw1c:
the main thing missing here is the `GOOGLE_APPLICATION_CREDENTIALS` environment variable setting escaping for the additional test and causing other tests to fail.

skimming over the commit history in the [`j1595-contextconfigclitest--wip-04`](https://github.com/oneconcern/datamon/commits/j1595-contextconfigclitest--wip-04) branch.  at 293dc0d66156ec15f86802c37de8386ff380ad54 , we see `go_test` pass in ci, showing that with the additional test skipped, the existing tests pass.  by 7655bc3b138c5942c6ab6a2ef4e85a7337a9d605 , [we see](https://app.circleci.com/jobs/github/oneconcern/datamon/4713) 

> failed to create config store: dialing: cannot read credentials file: open /Users/datamonuser/.config/gcloud/application_default_credentials.json: no such file or directory

this is the fixture data https://github.com/oneconcern/datamon/pull/394/files#diff-e376c74929cee3e31f542c8986ae00b5R92 from the additional test appearing in an existing test.

ransomw1c:
pretty sure this was subsumed by the workshop test?

do re-open if this is a test case we still want to cover

PR #396, opened 2020-02-05
Remove flag for indefinite hang for debug

kerneltime:


kerneltime:
@fredbi @ransomw1c @bluelion this should slave the pod stuck in terminating state.

ransomw1c:
here's at least one additional location for change given that the `-s` flag is removed...

https://github.com/oneconcern/datamon/blob/master/hack/fuse-demo/follow_coord_logs.sh#L103

... in fact, that's almost certainly what's causing the `fuse_sidecar_test` fail.

less clear on the `go_test` fail

fredbi:
could you please explain to me what's the problem we are facing here?

PR #397, opened 2020-02-05
init workshop web test

ransomw1c:
update the `WorkshopTest` to include a smoke test of the `web` command at the same level of abstraction (driving the cli as a binary rather than using Cobra's api) as the remainder of that test.

`routes_test.go` remains the place for more detailed specification- and documentation-by-test of the `web` package.

### DIFF
--- a/cmd/datamon/cmd/cli_fuse_test.go
+++ b/cmd/datamon/cmd/cli_fuse_test.go
@@ -206,7 +206,7 @@ func testMountCmdParams(t testing.TB, testType, repo, pathToMount, pathBackingFs
 	}
 }
 
-func testMountReady(t testing.TB, token string, defaultWait time.Duration) func(*zap.Logger, io.Reader) {
+func testWaitForReader(t testing.TB, token string, defaultWait time.Duration) func(*zap.Logger, io.Reader) {
 	return func(l *zap.Logger, output io.Reader) {
 		// awaits expected ready message
 		// TODO(fred): we may use that to time out on waiting if needed
@@ -245,15 +245,15 @@ func testMountReady(t testing.TB, token string, defaultWait time.Duration) func(
 }
 
 func TestBundleMount(t *testing.T) {
-	testBundleMount(t, "stream-dest", testMountReady(t, `"mounting"`, 5*time.Second))
+	testBundleMount(t, "stream-dest", testWaitForReader(t, `"mounting"`, 5*time.Second))
 }
 
 func TestBundleMountNoStream(t *testing.T) {
-	testBundleMount(t, "nostream-dest", testMountReady(t, `"mounting"`, 5*time.Second))
+	testBundleMount(t, "nostream-dest", testWaitForReader(t, `"mounting"`, 5*time.Second))
 }
 
 func TestBundleMountNoStreamNoDest(t *testing.T) {
-	testBundleMount(t, "nostream-nodest", testMountReady(t, `"mounting"`, 5*time.Second))
+	testBundleMount(t, "nostream-nodest", testWaitForReader(t, `"mounting"`, 5*time.Second))
 }
 
 func captureOutputProgress(p io.Reader) (*bytes.Buffer, *sync.WaitGroup) {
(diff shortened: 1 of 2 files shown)


ransomw1c:
these wound up being straightforward thanks to what Fred has built for the fuse and workshop tests

ransomw1c:
needs rebase and lint, out of draft mode.

ransomw1c:
current test fail is out of `TestSelfUpgrade`, occurs in `master`

cmd/datamon/cmd/cli_fuse_test.go
```diff
@@ -206,7 +206,7 @@ func testMountCmdParams(t testing.TB, testType, repo, pathToMount, pathBackingFs
 	}
 }
 
-func testMountReady(t testing.TB, token string, defaultWait time.Duration) func(*zap.Logger, io.Reader) {
+func makeBlockingTestWaiterForReader(t testing.TB, token string, defaultWait time.Duration) func(*zap.Logger, io.Reader) {
```
fredbi: I think we are in a contention here. Please keep the short names. Since you come with a more general use case for that func, I am suggesting `testWaitForReader` as a compromise.

At any rate  `makeMyselfJustTryOnceToTestIfWaitingForSomeReaderEventuallyWorksKnowingThatSaidReaderMightSometimesNotBeFunctionningProperly` is not something I would like to generalize.

cmd/datamon/cmd/cli_workshop_test.go
```diff
@@ -460,6 +506,23 @@ func deleteMeta(t testing.TB, bucket, repo string) (err error) {
 	return err
 }
 
+// lifted from `github.com/phayes/freeport`
+// GetFreePort asks the kernel for a free open port that is ready to use.
+func GetFreePort() (int, error) {
```
fredbi:
```suggestion
func testGetFreePort() (int, error) {
```

cmd/datamon/cmd/cli_workshop_test.go
```diff
@@ -343,8 +346,51 @@ func TestWorkshop(t *testing.T) {
 	})
 	require.NoError(t, err)
 
-	// no test here for web (see pkg/web)
 	// no test here for upgrade (see self_upgrade_test.go)
+
+	logger.Info("test web")
+	freePort, err := GetFreePort()
```
fredbi:
```suggestion
	freePort, err := testGetFreePort()
```

cmd/datamon/cmd/cli_fuse_test.go
fredbi:
```suggestion
func testWaitForReader(t testing.TB, token string, defaultWait time.Duration) func(*zap.Logger, io.Reader) {
```

cmd/datamon/cmd/cli_workshop_test.go
```diff
@@ -343,8 +346,51 @@ func TestWorkshop(t *testing.T) {
...
+
+	defer resp.Body.Close()
+	doc, err := goquery.NewDocumentFromReader(resp.Body)
+	require.NoError(t, err, "parse body html")
+
+	require.NotNil(t, doc)
+
+	logger.Info(fmt.Sprintf("home page:\n%v", doc.Text()))
+
+	// ??? more response content validation?
```
fredbi: no need to `???` here, since you already gave the answer in the body of this PR.
Suggesting instead:
```
// smoke test only: no content validation
```

cmd/datamon/cmd/cli_fuse_test.go
ransomw1c: actually, i recalled your comment on my request for a multi-word name on the response to Kevin's perf. improvement request, and chose a name you'd definitely object to ... probably should've left a comment.

of course the name needed to be changed (not only about `Mount`), so i'll go with the `testWaitForReader`, np.

cmd/datamon/cmd/cli_workshop_test.go
ransomw1c: 👍

PR #399, opened 2020-02-06
Diamonds #1: Detailed design for diamond workflow

fredbi:
Detailed design matching the proposal for diamond workflow.
This started from the original proposal doc, with substantial details.
There some changes, though, to some points in the proposal. Best to discuss all this.

1. Focus on usability: I've tried to reduce the number of commands and to keep subcommands and options as consistent as possible with already existing flag layout
2. Focus on usability/practical implementation: the `init` stage is opt-in, since any unique ID would do. It is not necessary to set up the metadata stage yet at this time.
3. Focus on usability: detailed conflict management, arbitration, inspection and resolution by users. Added a command for that.
4. Focus on transparency: added metadata and commands to report back on the diamond process, once completed, supplemented some commands with reporting options just for that.
5. Focus on data consistency: detailed (in length) the commit process and how concurrency, failures and job restarts are managed

Hope this helps. 

Next, I'll split the work into 3 PRs:
1. About metadata layout only (pkg/model)
2. About the `split add` data packing itself (mostly pkg/core)
3. Eventually about the CLI


fredbi:
Rendered link: https://github.com/oneconcern/datamon/blob/d70869699aaadef65d646ef565e30e318ab1d18b/docs/proposals/diamond-design.md

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,380 @@
...
+  /bundles/{repo}/{bundle-id}/conflicts/conflict-files-{index}.yaml
+  ```
+* The metadata to describe conflicts encompass:
+  * the reference to the originating diamond and split identifiers and timings
+  * for each conflicting path, all references (hash) to the conflicting versions (including the winner)
+
+### Handling failures, retries and cancellations
+
+Premises:
+* Some `add` operation might fail: the client has the opportunity to either _retry_ or _cancel_ the whole operation
```
ransomw1c: why would `add` fail?

ransomw1c: recall from discussion:  some confusion (likely on my part) about why `split add` needs _retry_ and _cancel_ notions in ui when `bundle upload` doesn't

fredbi: Will add some precision here. 
Cancellation is about the COMMIT operation, for the whole diamond, not individual split.
Retry is not a change in UI, but a change in user behaviour. I need to be more accurate/clarify these notions that are intrinsic to the new use case

fredbi:
@kerneltime  please stop merging every time from the github UI. This just produces noise. I shall anyway rebase any branch I am pushing on origin.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
+# Diamond workflow support
+
+Diamond workflows allow for a pipeline to scale out the generation of a single bundle.
```
fredbi: The initial "A diamond workflow allows for" was better...

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+      datamon bundle diamond initialize --repo {repo}
+      {diamond-id}
+      ```
+   5. Backend model: the following path is created in the vmetadata bucket.
+      ```
+      /splits/{repo}/{diamond-id}/diamond.yaml
+      ```
+
+> **Trade-off**: this requirement is set on purpose. It is not guided by implementation constraints but by the desire to protect users against
+> unintentional, silent data corruption to could occur if the diamond ID is left as a user-controlled parameter.
```
fredbi: "which could occur"

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+      /splits/{repo}/{diamond-id}/{split-id}/bundle-files-{index}.yaml
+      ```
+   4. The split ID is internal to datamon, and automatically generated with every `add` command.
+      The user is normally not required to know about this ID. Using KSUIDs to identify split allows for the ordering of uploads.
+   5. Incremental uploads for a bundle or running a diamond with splits work the same way. See [below](#diamond-vs-incremental-uploads)
+
+3. Eventually, all file lists combine into a new bundle with the commit message and possibly a label.
+
+    1. The initiator of the diamond **needs to coordinate with the different clients to decide a commit** and create a new bundle.
+    2. Any `split add` operation still in progress at commit time will be ignored from the resulting bundle.
```
fredbi: "in the resulting bundle"

docs/proposals/diamond-design.md
ransomw1c: additional detail

i suppose the ui for `split add` and `bundle upload` should be as similar as possible, because users are already familiar with – or need to be familiar with, anyway, `bundle upload`, as well as diamond bundles’ `commit` command.
the only failure mode that could distinuish `add` from `upload`, as far as i can tell, is duplicate split id.  there’s also non-exist diamond id, except behavior in case of this error is, i suppose, the same failure mode as non-exist repo name for `upload` – and for `add`, for that matter.
now, i would like avoid any discrepancy between `add` and `upload` at all by avoiding `add` erroring on duplicate split ids. instead, the most recently added split id that successfully completes (about the same thing, from the user’s perspective, as completing a bundle upload successfully) is the only split id added to the bundle (in absence of any additional conflict resolution mechanisms as described elsewhere in the document).

fredbi: I've mentioned in the PR main body that seeking similarity (esp. regarding flags) between these commands _is_ my goal. I am hoping that the recap at the bottom of the doc demonstrates this point.

I am sure about what "failure mode" means to you.

I am attempting here to detail the behavior in the case of failure _during the execution of the operation_, whatever their cause.
I am making a distinction between "controlled failures" (i.e. an exception that may be caught and managed, like a network failure for instance) and "uncontrolled failures" (i.e. pod crash).

So this is definitely not about errors caused by preliminary checks (like some repo does not exist, some id is a duplicate etc).

Eventually, I removed from this spec the capability given to users to inject the split ID. This remains a debatable matter (we discussed that in part yesterday already btw).

ransomw1c: further discussion reveals i was operating under the false assumption that this wouldn't include a change to `diamond.md`, disregard these comments.  i'm okay with the idea of making `split add` different in some ways from `bundle upload` and in fact suppose this is rather a proposal to properly (which i define as "according to users' expectations") resolve the incremental upload functionality request as well as diamond workflow support for PT.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+## Workflow
+
+The diamond workflow consists of the following steps.
+
+1. For a diamond to start, datamon needs some unique ID. Datamon SDK allows for the generation of such a unique ID for a diamond.
+
+   1. Prerequisite: generate a unique KSUID and a signalling metadata object that allows for starting splits.
+   2. All clients participating to the same diamond are expected to use the same ID at each step. **This is a requirement on client coordination**
+   3. The {diamond-id} is only used once: users cannot initialize the same diamond twice.
+   4. Initialization command, returning an identifier which is guaranteed to be unique:
```
kerneltime: I thought we had decided to let the client specify an ID but we would fail initialization if the ID was already used for the repo.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+      datamon bundle diamond add --repo {repo} --diamond {diamond-id} --path {source}
+      ```
+      This command is similar to `datamon bundle upload`, the only difference being the required `--diamond` flag.
+   3. Backend model: in the vmetadata bucket the following path will be created or rewritten.
+      ```
+      /splits/{repo}/{diamond-id}/{split-id}/bundle-files-{index}.yaml
+      ```
+   4. The split ID is internal to datamon, and automatically generated with every `add` command.
+      The user is normally not required to know about this ID. Using KSUIDs to identify split allows for the ordering of uploads.
+   5. Incremental uploads for a bundle or running a diamond with splits work the same way. See [below](#diamond-vs-incremental-uploads)
```
kerneltime: ```diamond-vs-incremental-uploads``` not working in rendered viewe

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+> **Trade-off**: this requirement is set on purpose. It is not guided by implementation constraints but by the desire to protect users against
+> unintentional, silent data corruption to could occur if the diamond ID is left as a user-controlled parameter.
+
+2. Uploading files in a diamond split stores files as blobs, then generates a file list in the versioned metadata bucket.
+
+   1. Files are uploaded by clients concurrently. This operation does not result in a new bundle yet.
+      Files are in a pre-commited state, meaning that all content is persisted as blobs on the CAFS store.
+   2. Command:
+      ```
+      datamon bundle diamond add --repo {repo} --diamond {diamond-id} --path {source}
```
kerneltime: one downside of not having ```split-id``` is that we not create conflicting files when a client retires the same files due to failure.. so if the customer wants to set commit with no conflict and there is no real conflict as the splits wrote independent files we would artificially introduce conflicts, breaking their failure handling and retry mechanism for a pod in the diamond. I think we will need to reintroduce the ```split-id``` or make it an optional flag where we honor the split-id and not internally generate one.

kerneltime: Also adding the suffix for the split id to the conflicting files allows the developer to better understand the issue with their logic, we could add our own internal KSUID but that is harder to use for the developer.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+```
+or
+```
+datamon bundle get --repo {repo} --bundle {bundle-id} --with-diamonds
+```
+
+### Inspect conflicts
+
+```
+datamon bundle list files --repo {repo} --bundle {bundle-id} --with-conflicts|--with-checkpoints
```
kerneltime: Not in MVP... this might be useful or might be totally unused. For now, I still prefer the original proposal to put all the conflicting files in a special folder within the bundle.

docs/proposals/diamond-design.md
fredbi: Ok. Can do

docs/proposals/diamond-design.md
kerneltime: Handling conflicts is what we need to help developers avoid vs. spending efforts to make it easier to handle.
Also for the current use cases, we do not expect this. The conflict handling was added to at least detect it if it occurs, we do not need to solve this in an elegant manner.
We can write objects in the diamond object-store path for each file to be written and help clients get a clock on. file as an example of avoiding conflicts.. so in terms of putting in the time, I would rather avoid conflicts than making it easier to handle. Either way, we do not need to implement an extension to the bundle MD that is designed purely for conflict handling.

docs/proposals/diamond-design.md
fredbi: Ok typo forgot plural in the link

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+
+Example:
+```
+/destination/my-conflicting-file.data                                     #  <-- winner
+/destination/my-conflicting-file.data-0ujsszwN8NRY24YaXiTIE2VWDTS         #  <-- candidate
+/destination/my-conflicting-file.data-0ujsswThIGTUYm2K8FjOOfXtY1K         #  <-- candidate
+/destination/my-conflicting-file.data-0ujssxh0cECutqzMgbtXSGnjorm         #  <-- candidate
+```
+
+### Resolve conflicts
```
kerneltime: Same as above, not in MVP, if we see developers battle this often, we can then decide how to help.
My suspicion is that developers might have requirements that we do not quite understand.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+
+
+If everything is fine and you just want to make it clear for all consumers of this bundle, you may force all conflicts to be resolved at once.
+```
+datamon bundle conflicts resolve --bundle {bundle-id} --resolve-all [--message {...} --label {...}]
+```
+This creates a new bundle with the same data (no upload here) and conflicts metadata are wiped out.
+The bundle on which conflicts where initially detected remains untouched for traceability.
+
+### Cancel a diamond
```
kerneltime: Not in MVP

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+datamon bundle diamond cancel --repo {repo} --diamond {diamond-id}
+```
+
+This waits for ongoing running splits to complete then marks vmetadata ready for cleanup
+
+**NOTE**: in the future, we may interrupt running splits rather than wait for them to complete _then_ cancel the dimaond
+
+### Identify splits with some custom node identifier
+```
+datamon bundle diamond split add --repo {repo} --path {source} --split-tag {my-bespoke-distinctive-tag}
```
kerneltime: why not split-id? I mentioned this before, this might the only way to deal with pod restarts for diamonds that do not generate conflicts and want to no conflicts during commit but have pods that restart. This should be the prescribed way to run a diamond workflow. We can make it easier by making it optional.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+              |_  cancel --repo {repo} --diamond {diamond-id}
+              |
+              |_  split
+                    |
+                    |_ add --repo {repo} --diamond {diamond-id} --path {source} [--split-tag {my-bespoke-distinctive-tag}] [--files {list of files}]
+                    |
+                    |_ list --repo {repo} --diamond {diamond-id}
+
+
+datamon bundle conflicts resolve --bundle {bundle-id} {--files {file-list}|--resolve-all} [--message {...} --label {...}]
```
kerneltime: Drop

kerneltime: Drop

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+
+datamon bundle get --repo {repo} --bundle {bundle-id} [--with-conflicts]
```
kerneltime: drop

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+
+datamon bundle get --repo {repo} --bundle {bundle-id} [--with-diamonds]
```
kerneltime: drop

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+
+datamon bundle list files --repo {repo} --bundle {bundle-id} [--with-conflicts|--with-checkpoints]
```
kerneltime: drop. Once a bundle it should be a normal bundle. I do not want to introduce types of bundles.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+
+datamon bundle download --repo {repo} --bundle {bundle-id} [--with-conflicts|--with-checkoints] --destination {local path}
```
kerneltime: Drop. Bundles should be normal bundles.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+```
+
+
+## Recap of the extended metadata layout
+
+### metadata
+```
+bundles/{repo}/{bundle-id}/bundle.yaml            # <- additional audit trail information
```
kerneltime: I do not see the need for additional MD for a bundle

docs/proposals/diamond-design.md
fredbi: Retrying same files without changing their content shall not trigger a conflict.

fredbi: Retrying same files without changing their content shall not trigger a conflict.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,458 @@
...
+```
+
+### vmetadata
+
+```
+splits/{repo}/{diamond-id}/diamond.yaml            # <- captures the diamond state: initialized, committing, done
+splits/{repo}/{diamond-id}/{split-id}/split.yaml   # <- captures the split state, plus holds information about the running split, eventually merged into metadata for bundle
+splits/{repo}/{diamond-id}/{split-id}/bundle-files-{index}.yaml
+
+conflicts/{repo}/{bundle-id}/conflict-files-{index}.yaml
```
kerneltime: Drop the conflicts and checkpoint, I want to keep the implementation for incremental and split as close as possible.. just allowing a ```split-id```` and not creating conflicts within a split-id will solve incremental uploads.

docs/proposals/diamond-design.md
fredbi: Agree with the priority to release the full feature later. However we need to prepare this: if we lose the references to the hashes we won t be able to deliver it without yet another migration.
Storing all intermediary hashes (comflicts or checkpoints) will s what i am after

fredbi: Agree with the priority to release the full feature later. However we need to prepare this: if we lose the references to the hashes we won t be able to deliver it without yet another migration.
Storing all intermediary hashes (comflicts or checkpoints) will s what i am after

fredbi: Ok not to extend bundle metadata. Will provision another location on vmetadata for that

fredbi: Ok not to extend bundle metadata. Will provision another location on vmetadata for that

docs/proposals/diamond-design.md
fredbi: An interesting case i haven t checked is when you remove a file from the dataset. Todo

docs/proposals/diamond-design.md
fredbi: Tag is a decorator for monitoring/inspecting results and bears no functional impact

fredbi: Thia is not true that conflicts are generated when restarting a pod with unchanged data

docs/proposals/diamond-design.md
fredbi: Was intended for audit trail inspection not typing bumdles

docs/proposals/diamond-design.md
fredbi: Intended as porcelain akin to what i recently proposed for —with-labels

docs/proposals/diamond-design.md
fredbi: I understand the militancy for —split-id but the undue creation of conflicts with my proposal is a false premise

fredbi: Implementation is thr same.conflicts and checkpoints only differ in name

docs/proposals/diamond-design.md
fredbi: I ve pit some background to that like inspecting ex post the timings of the different splits

docs/proposals/diamond-design.md
kerneltime: In case a pod that does not use split-id restarts and retries upload, at commit time there will be no way to know that the second file list is not in conflict. Also, the developer might want commit with no conflict in which case we will introduce a conflict where there is none.

docs/proposals/diamond-design.md
kerneltime: If a pod restarts it can regenerate the files and the files can have timestamps, so there is no guarantee that the data as computed by hashing will be the same.

docs/proposals/diamond-design.md
fredbi: now split ID may be injected to cover the discussed use-case (now detailed).
didn't provide details about removing files, although providing explicit split ID should cover that case.

docs/proposals/diamond-design.md
fredbi: Agreed with everything. Now conflicts are stored as special paths (note: filtered from uploads, like `.datamon` metadata files). Mentioned many things as "nice to have".
Kept however the distinction between conflict and checkpoints, which remains only matter of terminology.

docs/proposals/diamond-design.md
fredbi: Removed everything related to resolving conflicts.

docs/proposals/diamond-design.md
fredbi: ok put that one (cancel) as nice to have.

docs/proposals/diamond-design.md
fredbi: the use case with pod restart with changed dataset content is now addressed and examplified.
Notice that although split may be specified, I've kept it auto generated (KSUID).

fredbi: "taggging" splits for future inspection/timing analysis moved as a nice to have

docs/proposals/diamond-design.md
fredbi: ok dropped. Does make sense any longer with conflicting files stored in a special path on the bundle.

docs/proposals/diamond-design.md
fredbi: ok dropped. Does make sense any longer with conflicting files stored in a special path on the bundle. Bundle commands remain now unchanged.

docs/proposals/diamond-design.md
fredbi: ok dropped.
Notice that in the future, inspecting / reporting about diamonds & splits will cause the same kind of headache as we have now to join labels with bundles (typically the join operation is delegated to some - slower - porcelain func).

docs/proposals/diamond-design.md
fredbi: ok. Bundle commands now unaffected by diamonds.

docs/proposals/diamond-design.md
fredbi: ok. Reverted on that: bundle metadata now unaffected by diamonds.

docs/proposals/diamond-design.md
fredbi: ok dropped the specific metadata. conflicting files are moved to a special "hidden" path at commit time.

docs/proposals/diamond-design.md
fredbi: explicit split id supported

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
+# Diamond workflow support
+
+A diamond workflow allows for a pipeline to scale out the generation of a single bundle.
+
+A diamond is composed of multiple splits; each split tries to write a set of files.
+Ideally, 2 splits do not write the same file.
```
ransomw1c: So that this serves as documentation PEP-style, I would embed a definition here by saying

> If 2 splits write the same file, it's called a conflict, and ideally conflicts do not occur.

or something to that effect.  That way, when the commit logic uses the term "conflict" it's totally clear what's being described.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+
+To elaborate, a bundle consists of a filesystem namespace with files in folders; each split can write files to any
+folder but should not try to write the same file.
+
+The individual nodes add files to splits without any co-ordination between them.
```
ransomw1c: more commentary for clarity rather than fundamental idea:

could this be clarified as

> The individual nodes _in the DAG generated by the Argo runtime_ add files to splits without any co-ordination between them.

otherwise, it's not totally clear to me what vertex set is referred to here

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+
+Commit logic handles any conflicts that exist. Conflicts can only occur if the same file in the namespace of
+the bundle is written to by more than one split with different data.
+
+Diamonds and splits are essentially transitory objects, which live in the metadata namespace of a datamon repo,
```
ransomw1c: according to `defaultDiamond` in #403 , diamonds exist in `context.Stores.VMetadata`, the version metadata, rather than `context.Stores.Metadata`.

i probably am finicky about homonyms and that sort of thing lately, yet also it probably is important to be very clear about whether we're talking about `VMetadata` or `Metadata` in conversation since these two names are nearly the same and mean different things (i am still not totally clear on `VMetadata`'s meaning as distinct from `Metadata` other than the summary here about it being safe to delete on a GC-like basis).

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+   This disables all conflict detection on files coming from that split. See more about conflicts [below](#handling-conflicts).
+   ```
+   datamon diamond add --repo {repo} --diamond {diamond-id} --split {split-id} --path {source}
+   {split-id}
+   ```
+   This command fails when when the split has not been already created.
+6. Incremental uploads for a bundle and diamonds work the same way. See [below](#diamonds-vs-incremental-uploads)
+
+
+> **Trade-off**: the possibility given to the user to opt-in for a specific `{split-id}` reflects the desire
```
ransomw1c: this comment is an anchor for reference from pipelines#33

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+```
+
+
+## Detailed design
+
+### Diamonds vs incremental uploads
+
+Diamonds and incremental uploads are very much alike, the use-cases come with some nuances.
+
+Use-cases:
```
ransomw1c: i want to note here for thoroughness that adding some additional verbiage to the cli for the incremental upload from a single client use-case in particular rather than expecting desktop users to be familiar with all concepts exposed by the diamond cli additions.

docs/proposals/diamond-design.md
fredbi: what is PEP style?

docs/proposals/diamond-design.md
fredbi: I am just mentioning "nodes" here. There is no reason to introduce DAG, ARGO, vertices or other complex notions. A node simply executes an instance of the datamon program.

docs/proposals/diamond-design.md
fredbi: The way I seed it is that versioned metadata is a specialization of metadata.

docs/proposals/diamond-design.md
ransomw1c: [Python Enhancement Proposals](https://www.python.org/dev/peps/) (it turns out, had to refer to python.org to unacronymize) are the process for deciding things about Python.  even after decision-making, PEPs are often referred to by number, so i've said, "refer to PEP8" more times than i can remember due to the absence of `pyfmt`

ransomw1c: [572](https://hub.packtpub.com/why-guido-van-rossum-quit/) is another PEP, probably to live on in infamy

kerneltime:
Looks pretty good, will read it once more and approve.

fredbi:
Sure. I've tried hard to go back to the basics of immutable metadata (which is vastly superior). Although this led me to some implementation complexities for ancillary tasks such as listing objects and merging their various state-key combinations, the commit implementation is more straightforward.

Regarding split-id specification, I want to support both cases (externally supplied or not) since
it still remains pretty much unclear to me which path users will follow here.
What makes this possible is the introduction of the extra {generation-id}.

This was a nice dev to accomplish because it stressed a lot my initial micro-design with such a big change, which in the end did not change much code in the pkg/core I initially made.

PR #401, opened 2020-02-06
Bucket deletion

kerneltime:


ransomw1c:
additional note to self on intent:

need to ensure cli_test.go doesn’t delete incorrect buckets by creating a separate “project” (gg doc link?) for the test buckets.

PR #402, opened 2020-02-07
Diamonds #2: Diamond metadata  models and paths

fredbi:
~~design doc in this branch is not up-to-date: please ignore that one.~~

### DIFF
--- a/pkg/model/bundle.go
+++ b/pkg/model/bundle.go
@@ -15,11 +15,6 @@ import (
 	"github.com/oneconcern/datamon/pkg/cafs"
 )
 
-const (
-	// CurrentBundleVersion indicates the version of the bundle model
-	CurrentBundleVersion = 1.0
-)
-
 // BundleDescriptor represents a commit which is a file tree with the changes to the repository.
 type BundleDescriptor struct {
 	LeafSize               uint32        `json:"leafSize" yaml:"leafSize"`                               // Bundles blobs are independently generated
@@ -28,9 +23,9 @@ type BundleDescriptor struct {
 	Parents                []string      `json:"parents,omitempty" yaml:"parents,omitempty"`             // Bundles with parent child relation
 	Timestamp              time.Time     `json:"timestamp,omitempty" yaml:"timestamp,omitempty"`         // Local wall clock time
 	Contributors           []Contributor `json:"contributors" yaml:"contributors"`                       // Contributor for the bundle
-	BundleEntriesFileCount uint64        `json:"count" yaml:"count"`                                     // Number of files which have BundleDescriptor Entries
-	Version                uint64        `json:"version,omitempty" yaml:"version,omitempty"`             // Version for the bundle
-	Deduplication          string        `json:"deduplication,omitempty" yaml:"deduplication,omitempty"` // Type of deduplication used
+	BundleEntriesFileCount uint64        `json:"count" yaml:"count"`                                     // Number of file index files in this bundle
+	Version                uint64        `json:"version,omitempty" yaml:"version,omitempty"`             // Version for the metadata model used for this bundle
+	Deduplication          string        `json:"deduplication,omitempty" yaml:"deduplication,omitempty"` // Deduplication scheme used
 	RunStage               string        `json:"runstage,omitempty" yaml:"runstage,omitempty"`           // Path to the run stage
 	_                      struct{}
 }
@@ -83,6 +78,7 @@ type BundleEntry struct {
 	NameWithPath string      `json:"name" yaml:"name"`
 	FileMode     os.FileMode `json:"mode" yaml:"mode"`
 	Size         uint64      `json:"size" yaml:"size"`
+	Timestamp    time.Time   `json:"timestamp,omitempty" yaml:"timestamp,omitempty"` // time the file was uploaded. Only serialized with entries uploaded by splits (not bundles)
 	_            struct{}
 }
 
@@ -206,6 +202,8 @@ func GetBundleTimeStamp() time.Time {
 }
 
 // IsGeneratedFile indicates if some file comes from auto-generation (e.g. .datamon files)
+//
+// More generally, it indicates any special path on a consumable store not to be uploaded with a bundle.
 func IsGeneratedFile(file string) bool {
 	// TODO: Need to find a way for afero.Fs to convert to abs patch while honoring the fake root
 	//path, err := filepath.Abs(file)
@@ -215,5 +213,5 @@ func IsGeneratedFile(file string) bool {
 func init() {
 	metaRe = regexp.MustCompile(`^\.datamon/(.*)\.yaml$`)
 	flRe = regexp.MustCompile(`^(.*)-` + bundleFilesIndexPrefix + `(.*)$`)
-	genFileRe = regexp.MustCompile("^.datamon/*|^/.datamon/*|^/.datamon$|^.datamon$|^./.datamon/*|^./.datamon$")
+	genFileRe = regexp.MustCompile(`^\.datamon/.*|^/\.datamon/.*|^/\.datamon$|^\.datamon$|^\./\.datamon/.*|^\./\.datamon$|^\.?/?\.conflicts(/.*|$)|^\.?/?\.checkpoints(/.*|$)`)
 }
--- a/pkg/model/diamond.go
+++ b/pkg/model/diamond.go
@@ -1,0 +1,306 @@
+package model
+
+import (
+	"fmt"
+	"path"
+	"time"
+
+	"github.com/segmentio/ksuid"
+)
+
+// ConflictMode indicates the conflict detection mode defined for a diamond
+type ConflictMode string
+
+const (
+	// IgnoreConflicts is the diamond mode with which conflicts are not handled (latest win, no track kept of clobbered files)
+	IgnoreConflicts ConflictMode = "ignored"
+
+	// EnableCheckpoints is the diamond mode with which conflicts are explicitly handled and saved as "checkpoints" (incremental upload)
+	EnableCheckpoints ConflictMode = "enable-checkpoints"
+
+	// EnableConflicts is the diamond mode with which conflicts are detected and saved as "conflicts"
+	EnableConflicts ConflictMode = "enable-conflicts"
+
+	// ForbidConflicts is the diamond mode with which conflicts result in failure to commit
+	ForbidConflicts ConflictMode = "forbids-conflicts"
+)
+
+// DiamondState models the running status of an ungoing diamond workflow
+type DiamondState string
+
+const (
+	// DiamondInitialized is the state of an initialized diamond
+	DiamondInitialized DiamondState = "initialized"
+
+	// DiamondDone indicates the diamond has completed with a successful commit. This is a terminal state.
+	DiamondDone DiamondState = "done"
+
+	// DiamondCanceled indicates the diamond has completed with a cancel. This is a terminal state.
+	DiamondCanceled DiamondState = "canceled"
+)
+
+const (
+	done                         = "done"
+	running                      = "running"
+	ext                          = ".yaml"
+	diamondPrefix                = "diamond-"
+	splitPrefix                  = "split-"
+	diamondFinalDescriptorFile   = diamondPrefix + done + ext
+	splitFinalDescriptorFile     = splitPrefix + done + ext
+	diamondInitialDescriptorFile = diamondPrefix + running + ext
+	splitInitialDescriptorFile   = splitPrefix + running + ext
+)
+
+// IsValid checks the value of a diamond state
+func (s DiamondState) IsValid() bool {
+	switch s {
+	case DiamondInitialized, DiamondDone, DiamondCanceled:
+		return true
+	default:
+		return false
+	}
+}
+
+func (s DiamondState) String() string {
+	return string(s)
+}
+
+// SplitState models the running status of an ungoing diamond split
+type SplitState string
+
+const (
+	// SplitDone is the state of a completed split. This is a terminal state.
+	SplitDone SplitState = done
+
+	// SplitRunning is the state of running split
+	SplitRunning SplitState = running
+)
+
+// IsValid checks the value of a diamond state
+func (s SplitState) IsValid() bool {
+	switch s {
+	case SplitDone, SplitRunning:
+		return true
+	default:
+		return false
+	}
+}
+
+func (s SplitState) String() string {
+	return string(s)
+}
+
+// DiamondDescriptor models a diamond's metadata
+type DiamondDescriptor struct {
+	DiamondID      string            `json:"diamondID" yaml:"diamondID"`
+	StartTime      time.Time         `json:"startTime" yaml:"startTime"`                           // documentary: when the diamond was initialized
+	EndTime        time.Time         `json:"endTime,omitempty" yaml:"endTime,omitempty"`           // documentary: the diamond completion time (i.e. when the diamond reached a terminal state)
+	State          DiamondState      `json:"state" yaml:"state"`                                   // the captured state of the diamond
+	Mode           ConflictMode      `json:"mode" yaml:"mode"`                                     // the conflict handling mode defined at commit time. The default value is "enable-conflicts"
+	HasConflicts   bool              `json:"hasConflicts,omitempty" yaml:"hasConflicts,omitempty"` // documentary snapshot of the outcome of conflict handling after commit
+	HasCheckpoints bool              `json:"hasCheckpoints,omitempty" yaml:"hasCheckpoints,omitempty"`
+	Tag            string            `json:"tag,omitempty" yaml:"tag,omitempty"`           // user-defined tag to taint logs
+	BundleID       string            `json:"bundleID,omitempty" yaml:"bundleID,omitempty"` // keeps track of the produced bundle, when the diamond is successfully committed
+	Splits         []SplitDescriptor `json:"splits,omitempty" yaml:"splits,omitempty"`     // documentary snapshot of the splits collected after a successful commit
+	_              struct{}
+}
+
+// DiamondDescriptors is a sortable slice of DiamondDescriptor
+type DiamondDescriptors []DiamondDescriptor
+
+func (b DiamondDescriptors) Swap(i, j int) {
+	b[i], b[j] = b[j], b[i]
+}
+func (b DiamondDescriptors) Len() int {
+	return len(b)
+}
+func (b DiamondDescriptors) Less(i, j int) bool {
+	if !b[i].StartTime.IsZero() && !b[j].StartTime.IsZero() {
+		return b[i].StartTime.Before(b[j].StartTime)
+	}
+	return b[i].DiamondID < b[j].DiamondID
+}
+
+// Last bundle descriptor in slice
+func (b DiamondDescriptors) Last() DiamondDescriptor {
+	return b[len(b)-1]
+}
+
+func defaultDiamondDescriptor() *DiamondDescriptor {
+	diamondID, err := ksuid.NewRandom()
+	if err != nil {
+		panic(fmt.Sprintf("cannot generate random ksuid: %v", err))
+	}
+	return &DiamondDescriptor{
+		DiamondID: diamondID.String(),
+		StartTime: GetBundleTimeStamp(),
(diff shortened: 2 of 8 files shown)


pkg/model/diamond.go
```diff
@@ -0,0 +1,215 @@
...
+	return d
+}
+
+// SplitDescriptor models the metadata about a given split within a diamond
+type SplitDescriptor struct {
+	SplitID      string        `json:"splitID" yaml:"splitID"`
+	DiamondID    string        `json:"diamondID" yaml:"splitID"`
+	StartTime    time.Time     `json:"startTime" yaml:"starTime"`
+	EndTime      time.Time     `json:"endTime,omitempty" yaml:"endTime,omitempty"`
+	State        SplitState    `json:"state" yaml:"state"`
```
ransomw1c: swooping in kind of early with comments here in order to understand design proposal as much as talk about implementation details .. was i was trying to ask in my comments about `split add` is why `State` is appearing as a struct field rather than a method.

related but not equivalent question is why there are distinct `running` and `failed` states.  and this definitely veers to implementation detail:  don't even understand how we could distinguish in-progress `add`s from `failed` adds in all cases, such as if the internet connection cuts out during the 2KiB slices' upload. 

i suspect this'll be clarified by actual implementation and is best discussed after more source is written.

fredbi: > why State is appearing as a struct field rather than a method.

state is persisted

>  why there are distinct running and failed states. and this definitely veers to implementation detail: don't even understand how we could distinguish in-progress adds from failed adds in all cases, 

Design

    Diamonds abide by the following state-transition model:
        initialized
        committing
        done|cancelled

    Splits:
        running
        done|failed

    NOTE: the cancelled and failed states are not critical to the functionality: they merely provide better monitoring capabilities (""what's going on with my splits?") as well as safe cleaning up.


### Failed add operation

A controlled failure on such an operation would mark the split as explicitly failed (for further reporting and monitoring).

/splits/{repo}/{diamond-id}/{split-id}/split.yaml   #  <- State pass to "failed"

Please mark that it is not possible to distinguish an aborted or hanging add operation from a merely long running one.

ransomw1c: since there's a "controlled failure", there's also an "uncontrolled failure" that also can be distinguished from aborted, hanging, or long-running adds?

fredbi: yes indeed

ransomw1c: ok, so i guess this was my fundamental qualm with having the failure statuses:

how much of the space of all possible failures (both controlled and uncontrolled) needs to be covered by controlled failures in order for controlled failures to be meaningful?

by now, the decision-making is veering toward the philosophical, so it's probably time to let the overall design play out in source while keeping the more abstruse considerations in mind for future revisions?

marking this as resolved in any case.

pkg/model/bundle.go
```diff
@@ -63,6 +58,7 @@ type BundleEntry struct {
 	NameWithPath string      `json:"name" yaml:"name"`
 	FileMode     os.FileMode `json:"mode" yaml:"mode"`
 	Size         uint64      `json:"size" yaml:"size"`
+	SnapshotTime time.Time   `json:"snapshotTime,omitempty" yaml:"SnapshotTime,omitempty"` // time the file was uploaded
```
kerneltime: Why not just call it timestamp?

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+
+func (s SplitState) String() string {
+	return string(s)
+}
+
+// DiamondDescriptor models a diamond's metadata
+type DiamondDescriptor struct {
+	DiamondID       string            `json:"diamondID" yaml:"diamondID"`
+	StartTime       time.Time         `json:"startTime" yaml:"startTime"` // when the diamond was initialized
+	EndTime         time.Time         `json:"endTime,omitempty" yaml:"endTime,omitempty"`
```
kerneltime: Having startime and endtime in the same model? This is probably introducing complicated tracking of metadata RMW to be correct.

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	StateChangeTime time.Time         `json:"stateChangeTime" yaml:"stateChangeTime"` // last state change
+	State           DiamondState      `json:"state" yaml:"state"`
+	Mode            ConflictMode      `json:"mode" yaml:"mode"`
+	HasConflicts    bool              `json:"hasConflicts,omitempty" yaml:"hasConflicts,omitempty"`
```
kerneltime: I thought we had decided that we will defer adding conflicts are as a metadata entity but leave to the logic at commit time to create the special files within the bundle that indicate conflict. 
I am very hesitant to introduce this into the model before we have time to understand how developers want this to evolve. We can add it later.

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	HasCheckpoints  bool              `json:"hasCheckpoints,omitempty" yaml:"hasCheckpoints,omitempty"`
```
kerneltime: Let's differ incremental upload.

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	Tag             string            `json:"tag,omitempty" yaml:"tag,omitempty"`
```
kerneltime: What is tag?

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	BundleID        string            `json:"bundleID,omitempty" yaml:"bundleID,omitempty"`
```
kerneltime: I am a little concerned about having metadata that needs to be continuously updated during the lifecycle. We can write a new file in the path that indicates which bundle was written for a diamond. 
Object stores offer put if not present which is a very powerful primitive and allows us to only have one bundle be reported to the client as having been created for a diamond.

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+		apply(d)
+	}
+	return d
+}
+
+// SplitDescriptor models the metadata about a given split within a diamond
+type SplitDescriptor struct {
+	SplitID               string        `json:"splitID" yaml:"splitID"`
+	StartTime             time.Time     `json:"startTime" yaml:"startTime"`
+	EndTime               time.Time     `json:"endTime,omitempty" yaml:"endTime,omitempty"`
```
kerneltime: End time and start time in the same file..

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	State                 SplitState    `json:"state" yaml:"state"`
```
kerneltime: What is state?

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	Contributors          []Contributor `json:"contributors" yaml:"contributors"` // Contributor for the bundle
+	SplitEntriesFileCount uint64        `json:"count" yaml:"count"`               // Number of index files in this bundle
+	Tag                   string        `json:"tag,omitempty" yaml:"tag,omitempty"`
```
kerneltime: What is tag?

pkg/model/diamond_options.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+// SplitContributor defines a single contributor for a SplitDescriptor
+func SplitContributor(c Contributor) SplitDescriptorOption {
+	return SplitContributors([]Contributor{c})
+}
+
+// SplitTag sets an informative tag on the split
+func SplitTag(tag string) SplitDescriptorOption {
```
kerneltime: If tag is a message a client attaches then use the same nomenclature.

pkg/model/bundle.go
fredbi: yes will do.

pkg/model/diamond.go
fredbi: This is indeed working as you describe (and as agreed).
I've added this flag as a short reminder that the commit has detected something, I am not seeing this as harmful.

pkg/model/diamond.go
fredbi: StartTime is set at initialization time, end time is updated when the diamond state transitions to a terminal state.

pkg/model/diamond.go
fredbi: It is done and works, basically at a low extra complexity cost.
I'd suggest we observe how users react to that and possibly simplify later.

pkg/model/diamond.go
fredbi: Tag is just an informative string to be kept in the logs and in list commands.
I am basically using this to triage log from many splits and coordinator pods

fredbi: It has no impact on the functionality and is essentially here to make log more readable.

pkg/model/diamond.go
fredbi: The diamond is updated to reflect its final state at commit time. This is where I stitch in the bundle.
I understand your concern about updates, which indeed exposes us to some race conditions at the moment (this point is my last todo in the `core` package). In its present state, the race condition remains rather narrow and I'd say, quite unlikely. Still, I have some leads to better use the gcs primitive here.

pkg/model/diamond.go
fredbi: Yes and I don't see what's wrong with that, given that the whole thing about diamond is that they are updated (which is, I agree, a novelty when compared to the previous usage of metadata).

pkg/model/diamond.go
fredbi: state reflects the state of the split and is used at commit time to collect only  completed splits in the resulting bundle.

pkg/model/diamond.go
fredbi: same intent as explained above

> Tag is just an informative string to be kept in the logs and in list commands.
> I am basically using this to triage log from many splits and coordinator pods

pkg/model/diamond_options.go
fredbi: this has nothing to do with the commit message, which does not pertain to the split object.

pkg/model/paths.go
```diff
@@ -29,24 +34,24 @@ type ArchivePathComponents struct {
...
+	DiamondID       string
+	SplitID         string
 }
 
-// GetArchivePathComponents yields all components from an archive path.
-//
-// NOTE: this function's design is converging on being able to return something meaningful
-// given any path in the metadata archive, not just those corresponding to bundles.
-//
-// The return value might be changed to an interface type in later iterations.
```
ransomw1c: with the last iteration of this comment being elided from the source and [yet another](https://github.com/oneconcern/datamon/pull/403#discussion_r379583010) instance of "let's use `interfaces` to represent sum/union types!" enthusiasm appearing in my commentary elsewhere, i'd like to have the vestiges of this link trail (into github comments, into git history) referenced from the issue tracker.

so i am doing just that!  creating a spin-off issue here and resolving this comment immediately.

fredbi: My point was that:
>  this function's design is converging on being able to return something meaningful
has no more justification, since this function _has converged_ on being able to return something meaningful.

pkg/model/bundle.go
ransomw1c: was going to comment that we might need to take precautions for backwards-compatibility by adding some nil checks on this additional struct field.  except then i guess the `omitempty` token in the serialization dsl-in-annotations pretty much covers nil-checking?

while here, are the `omitempty` tags more or less current with those metadata fields that might be nil-able – like, are there any `omitempty` tags that can be removed supposing all metadata has been created by v2+?

pkg/model/bundle.go
```diff
@@ -195,5 +191,5 @@ func IsGeneratedFile(file string) bool {
 func init() {
 	metaRe = regexp.MustCompile(`^\.datamon/(.*)\.yaml$`)
 	flRe = regexp.MustCompile(`^(.*)-` + bundleFilesIndexPrefix + `(.*)$`)
-	genFileRe = regexp.MustCompile("^.datamon/*|^/.datamon/*|^/.datamon$|^.datamon$|^./.datamon/*|^./.datamon$")
+	genFileRe = regexp.MustCompile(`^\.datamon/.*|^/\.datamon/.*|^/\.datamon$|^\.datamon$|^\./\.datamon/.*|^\./\.datamon$|^\.?/?\.conflicts(/.*|$)|^\.?/?\.checkpoints(/.*|$)`)
```
ransomw1c: recall:  `.conflicts` and `.checkpoints` are part of the diamond conflict resolution mechanism described in the design document

pkg/model/diamond.go
ransomw1c: note to self that this is in the "nice to have" [section at the end](https://github.com/oneconcern/datamon/blob/diamond-metadata-paths/docs/proposals/diamond-design.md#nice-to-have--future) of the design document

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	// ForbidConflicts is the diamond mode with which conflicts result in failure to commit
+	ForbidConflicts ConflictMode = "forbids-conflicts"
+)
+
+// DiamondState models the running status of an ungoing diamond workflow
+type DiamondState string
+
+const (
+	// DiamondInitialized is the state of an initialized diamond
+	DiamondInitialized DiamondState = "initialized"
```
ransomw1c: more on expressing language constructs in golang, `enum` edition:

the only thought i've put into `enum` types so far is that we use `iota` to make `enum`s in golang a lot more like `enum`s in C wherein we're declaring integer constants rather than `enum`s in Java where we have something that's a lot more like an algebraic data type except without parameters on any of the unioned record types.

**succinctly**
since the preceding paragraph is probably meaningless, what i'm saying specifically is to `DiamondInitialized DiamondState = itoa` here, then `switch` from C-style enum (just an `int`!) to string value for serialization.

this is not based on anything other than preference and may be safely disregarded if it matters too much one way or another, possibly from a perspective i'm not considering.

pkg/model/bundle.go
ransomw1c: actually, i could also ask:  was remembering to change this regexp a gotcha?  or another way:  was the existing considered technical debt to resolve sooner rather than later?

would it be better to encode the guarantee about the bundles' "dotfiles" having the same naming convention here and wherever the bundle's dotfiles are written out, or could it be that as long as this regexp is append only (+ bugfixes) things will work out okay?

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// DiamondDone indicates the diamond has completed with a commit
+	DiamondDone DiamondState = "done"
+
+	// DiamondCancelled indicates the diamond has completed with a cancel
+	DiamondCancelled DiamondState = "cancelled"
+)
+
+const (
+	diamondDescriptorFile = "diamond.yaml"
```
ransomw1c: nit: to `const` expression with `bundleDescriptorFile`, etc. in `paths.go`

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	splitDescriptorFile   = "split.yaml"
```
ransomw1c: nit: to `const` expression with `bundleDescriptorFile`, etc. in `paths.go`

pkg/model/diamond.go
ransomw1c: i think RMW means "Read Mutate Write", and although i didn't plan on getting into whether to use such a RMW mechanism as part of the review of this initial sequence of commits, i did make [a comment](https://github.com/oneconcern/datamon/pull/403#discussion_r379651519) on an alternative (Write Once Read Many -type) mechanism

pkg/model/diamond.go
ransomw1c: i'm not currently considering this harmful either, except in general i do find that paring down data structures can save complexity, making for a more codebase, even though it sometimes seem like having some additional statefulness solves particular problems when in the thick of developer procedural code.

the first alternative to removing this altogether that occurs to me is to encode the conflict status in the `State` pallet:  expand `done` to `done-conflict` and `done-noconflict`, similarly expand `failed`.  this way, we can determine conflict status from metadata _and_ we know that the conflict status doesn't get set until commit (or attempted commit) time.

pkg/model/diamond.go
```diff
@@ -0,0 +1,265 @@
...
+	StartTime       time.Time         `json:"startTime" yaml:"startTime"` // when the diamond was initialized
+	EndTime         time.Time         `json:"endTime,omitempty" yaml:"endTime,omitempty"`
+	StateChangeTime time.Time         `json:"stateChangeTime" yaml:"stateChangeTime"` // last state change
+	State           DiamondState      `json:"state" yaml:"state"`
+	Mode            ConflictMode      `json:"mode" yaml:"mode"`
+	HasConflicts    bool              `json:"hasConflicts,omitempty" yaml:"hasConflicts,omitempty"`
+	HasCheckpoints  bool              `json:"hasCheckpoints,omitempty" yaml:"hasCheckpoints,omitempty"`
+	Tag             string            `json:"tag,omitempty" yaml:"tag,omitempty"`
+	BundleID        string            `json:"bundleID,omitempty" yaml:"bundleID,omitempty"`
+	Splits          []SplitDescriptor `json:"splits,omitempty" yaml:"splits,omitempty"`
```
ransomw1c: aside from the general motif of separating out data (struct fields) according to mutability expectations/guarantees , the `Splits` field is the most surprising part of the `DiamondDescriptor` to me, because we'll also be writing

```
/diamonds/{repo}/{diamond-id}/splits/{split-id}/split.yaml
/diamonds/{repo}/{diamond-id}/splits/{split-id}/bundle-files-{index}.yaml
```

on `split add`.

reading ahead, i can see this field is set only upon `diamond commit`, not `split add` (i'm pretty sure?), and going further, i'm pretty sure that it's only accessed during that operation as well such that it needn't necessarily be part of this struct?

this is all happening in vmetadata in any case, so the possibility for inconsistency by writing the same data in two places isn't particularly disturbing, particularly if we get the GC-like pod going sooner rather that later, since that program can also do things like check consistency / validate vmetadata.

pkg/model/bundle.go
fredbi: This extra field must only be present for "file list files" (aka index) produced by the split.
It is removed from the final bundle index files as I do not want to introduce any change with existing metadata.
Hence: `omitempty` does matter.

pkg/model/diamond.go
fredbi: iota is just a syntactic sugar to avoid repeating numerical indices.
Both idioms exist in goland and this doesn't change the way switch statements look like.

pkg/model/bundle.go
fredbi: this was a bug and as I added tests specifically to check against generated files for conflicts, I had to change that regexp. While doing so, I also fixed the ".datamon" metadata part.

pkg/model/diamond.go
fredbi: the whole thing should move to a subpackage called something like `paths` or `metapaths`

ransomw1c:
ok, i've made more comments than this patch probably warrants, so here's a summary of those things that're perhaps actually mildly worrying about merging this in

* separation of immutable and mutable fields and the implicit RMW mechanism when no such separation exists (likely resolution:  address with backwards compatible changes in later additions)
* duplicate data in `vmeta` bucket between split descriptors and `Splits` field of diamond descriptor (not too much of a worry supposing we tack consistency checks onto the forthcoming GC pod's responsibilities)
* nil checks / nil validation

with the last being the most worrying, because it will be the most difficult for me to keep track of after the contents of these patches are out of working memory for a while.

specifically, i'd like to have a validation routine to encode the intent

> [`BundleEntry`'s `time.Time` field] is removed from the final bundle index files

and possibly other validation steps performed wherever these structs are serialized and put to GCS (currently `uploadBundleEntriesFileList` + #403 contents).

i can add this as an addendum after #403 merges if that's more convenient.

----

so all worries aside, i don't see anything that absolutely need block this merge request, presupposing there's also approval from Ritesh.

pkg/model/diamond.go
fredbi: Objects are now immutable, although they share the same description, states are handled in diferent files.

pkg/model/bundle.go
fredbi: ok

PR #403, opened 2020-02-13
Diamonds #4: implemented core operations for diamonds and splits - READY

fredbi:
~~Still very much WIP.
Those interested may take a look out of curiosity.~~
Feature complete, ready for review

NOTE to reviewers:  this PR gets on board the content of PRs Diamond #1, #2 and #3. 
You might want to focus here on the content of `pkg/core`, other locations being separated in the other PRs.

### DIFF
--- a/pkg/core/bundle_unpack.go
+++ b/pkg/core/bundle_unpack.go
@@ -155,6 +155,7 @@ func unpackBundleDescriptor(ctx context.Context, bundle *Bundle, publish bool) e
 type bundleEntriesRes struct {
 	bundleEntries model.BundleEntries
 	idx           uint64
+	id            string // the splitID when used in a diamond split context
 }
 
 type downloadBundleFileListChans struct {
--- a/pkg/core/diamond_commit.go
+++ b/pkg/core/diamond_commit.go
@@ -1,0 +1,397 @@
+package core
+
+import (
+	"sync"
+	"time"
+
+	iradix "github.com/hashicorp/go-immutable-radix"
+	"github.com/oneconcern/datamon/pkg/core/status"
+	"github.com/oneconcern/datamon/pkg/errors"
+	"github.com/oneconcern/datamon/pkg/model"
+	"github.com/segmentio/ksuid"
+	"go.uber.org/zap"
+)
+
+// bufferingFactor determines the size of the channel buffers as a multiple of entriesPerFile
+//
+// NOTE: we can significantly speed up the process by allocating a larger buffer here. However, this
+// stresses the implementation of Store, e.g. not closing descriptors properly (localfs) makes a high
+// buferred process rapidly overcome system limits (i.e. max open files).
+const bufferingFactor = 100
+
+// Commit a diamond
+func (d *Diamond) Commit(opts ...Option) error {
+	return d.implCommit(opts...)
+}
+
+// Cancel a diamond
+func (d *Diamond) Cancel() error {
+	err := d.downloadDescriptor()
+	if err != nil {
+		return err
+	}
+
+	switch d.DiamondDescriptor.State {
+	case model.DiamondCanceled, model.DiamondDone:
+		return errors.New("diamond is already terminated").
+			WrapMessage("diamond state: %v", d.DiamondDescriptor.State)
+	default:
+		err = d.WithState(model.DiamondCanceled).uploadDescriptor()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// WithState returns the diamond with its state modified and transition timestamp updated
+func (d *Diamond) WithState(state model.DiamondState) *Diamond {
+	if state == d.DiamondDescriptor.State {
+		return d
+	}
+	d.DiamondDescriptor.State = state
+
+	switch state {
+	case model.DiamondDone, model.DiamondCanceled:
+		// writes down the time when reaching a terminal state
+		d.DiamondDescriptor.EndTime = model.GetBundleTimeStamp()
+	}
+	return d
+}
+
+// collectSplits which are done, sorted by completion time
+func (d *Diamond) collectSplits(opts ...Option) (model.SplitDescriptors, error) {
+	splits := make(model.SplitDescriptors, 0, typicalSplitsNum)
+
+	err := ListSplitsApply(
+		d.RepoID,
+		d.DiamondDescriptor.DiamondID,
+		d.contextStores,
+		func(sd model.SplitDescriptor) error {
+			if sd.State == model.SplitDone {
+				splits = append(splits, sd)
+			}
+			return nil
+		},
+		opts...,
+	)
+	if err != nil {
+		return nil, errors.New("couldn't collect splits").Wrap(err)
+	}
+	return splits, nil
+}
+
+func (d *Diamond) checkBundleID() error {
+	if !enableBundlePreserve {
+		// feature guard: require specific build flag
+		return nil
+	}
+	if d.BundleID == "" {
+		// regular case: bundle ID is generated on the fly
+		return nil
+	}
+
+	// case of bundleID preservation (e.g. data migration use case)
+	id, err := ksuid.Parse(d.BundleID)
+	if err != nil {
+		return status.ErrInvalidKsuid.Wrap(err)
+	}
+	d.setBundleID(id.String())
+	exists, err := d.Exists(d.contexter())
+	if err != nil {
+		return errors.New("could not check for bundle existence").Wrap(err)
+	}
+	if exists {
+		return status.ErrBundleIDExists.WrapMessage("bundleID: %v", id)
+	}
+	return nil
+}
+
+func (d *Diamond) implCommit(opts ...Option) (err error) {
+	// check if repo exists
+	if err = RepoExists(d.RepoID, d.contextStores); err != nil {
+		return err
+	}
+
+	// check when bundle is already specified (e.g. data migration use case, with bundleID preservation)
+	if err = d.checkBundleID(); err != nil {
+		return err
+	}
+
+	logger := d.l.With(zap.String("diamond_id", d.DiamondDescriptor.DiamondID))
+
+	// critical section #1
+
+	// check that the diamond is ready to accept splits (e.g. not already done or canceled)
+	if err = diamondReady(d.RepoID, d.DiamondDescriptor.DiamondID, d.contextStores); err != nil {
+		return errors.New("cannot proceed with diamond commit").WrapWithLog(logger, err)
+	}
+
+	// before exiting, save the diamond state either as done or rollbacked to initialized
+	defer func() {
+		if err != nil {
+			// TODO(fred): nice - last ditch check done automatically with nooverwrite: problem is error qualification...
+			logger.Error("diamond commit failed", zap.Error(err))
+			return
+		}
+		err = d.WithState(model.DiamondDone).uploadDescriptor()
+		if err != nil {
+			logger.Warn("failed to complete diamond commit, but bundle is available", zap.String("bundle_id", d.BundleID), zap.Error(err))
+		}
+	}()
+
+	// walk all completed splits
+	// TODO(fred): nice - performances - should be piped to next stage asynchronously - at the moment, we start collecting index files
+	// only after metadata about all splits have been collected.
+	splits, err := d.collectSplits(opts...)
+	if err != nil {
+		return err
+	}
+
+	if len(splits) == 0 {
+		return errors.New("no split to commit")
+	}
+
+	// capture splits summarized metadata in the final diamond descriptor
+	d.DiamondDescriptor.Splits = splits
+	d.l.Info("splits for this commit", zap.Int("num_splits", len(splits)))
+
+	// prepare indexer to walk over all index files for all splits
+	d.splitIndexer = d.makeDownloadIndexer()
+
+	// merge contributors
+	d.BundleDescriptor.Contributors = d.BundleDescriptor.Contributors[:0]
+	for _, split := range d.DiamondDescriptor.Splits {
+		d.BundleDescriptor.Contributors = append(d.BundleDescriptor.Contributors, split.Contributors...)
+	}
+
+	if d.BundleID == "" {
+		err = d.InitializeBundleID()
+		if err != nil {
+			return err
+		}
+	}
+
+	// prepare index to upload bundle index files
+	d.bundleIndexer = d.makeUploadIndexer()
+
+	// upload file lists to the target bundle
+	filePackedC := make(chan filePacked, bufferingFactor*d.bundleIndexer.entriesPerFile)
+	errorC := make(chan errorHit)
+	doneOkC := make(chan struct{})
+
+	// start merging split file lists
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go d.mergeSplits(filePackedC, errorC, doneOkC, &wg)
+	defer wg.Wait()
+
+	// hand out all files to merger goroutine
+	count, err := d.bundleIndexer.Upload(filePackedC, errorC, doneOkC)
+	if err != nil {
+		return err
+	}
+
+	// finalize bundle metadata
+	d.BundleDescriptor.BundleEntriesFileCount = count
+
+	err = uploadBundleDescriptor(d.contexter(), d.Bundle)
+	if err != nil {
+		return err
+	}
+
+	d.l.Info("uploaded bundle id", zap.String("BundleID", d.BundleID))
+	d.DiamondDescriptor.BundleID = d.BundleID
+	return nil
+}
+
+type mergeEntry struct {
+	model.BundleEntry        // single file
+	ID                string // split ID which uploaded this file
+}
+
+// mergeSplits merges all files from splits and resolves conflicts
+// then produces the output to the given channels.
+//
+// TODO(fred): nice - ok so the channel stuff here might look irrelevant for that use case.
+// I am reusing Download, which is asynchronous, but everything is buffered in the loop below.
+// I am also using buffered channels here...
+// I am thinking channel might be useful when we need to offload part of this to disk (asynchronously).
+//
+// TODO(fred): scalability - at the moment, the merger is carried out in memory. Use local storage to merge very large file lists
+// TODO(fred): nice - a more careful choice of type in lieu of filePacked could avoid some extra data copy => will tend to that when refactoring with bundle
+func (d *Diamond) mergeSplits(filePackedC chan<- filePacked, errorC chan<- errorHit, doneOkC chan<- struct{}, wgg *sync.WaitGroup) {
+	defer wgg.Done()
+
+	var (
+		wg sync.WaitGroup
+
+		// metrics
+		conflicts, merged, bundleEntries uint64
+	)
+
+	mode := d.DiamondDescriptor.Mode
+	t0 := time.Now()
+	interrupt := make(chan struct{}, 1)
+
+	// proceed with merger
+	wg.Add(1)
+	go func(input <-chan bundleEntriesRes, output chan<- filePacked, interrupt <-chan struct{}, wg *sync.WaitGroup) {
+		defer wg.Done()
+
+		mergeIndex := iradix.New()
+		for res := range input {
+			splitID := res.id
+			d.l.Debug("merge received batch", zap.String("from split", splitID), zap.Int("num_entries", len(res.bundleEntries.BundleEntries)))
+			for _, file := range res.bundleEntries.BundleEntries {
+				select {
+				case <-interrupt:
+					return
+				default:
+				}
+				merged++
+				d.l.Debug("merge received file entry", zap.String("from split", splitID), zap.String("entry", file.NameWithPath))
+				key := []byte(file.NameWithPath)
+				obj, found := mergeIndex.Get(key)
+				if !found {
+					mergeIndex, _, _ = mergeIndex.Insert(key, mergeEntry{BundleEntry: file, ID: splitID})
+					continue
+				}
+
+				existing := obj.(mergeEntry)
+				if file.Hash == existing.Hash {
+					continue
+				}
+
+				if file.Timestamp.IsZero() {
+					panic("dev error: files should have a timing") // internal safeguard
+				}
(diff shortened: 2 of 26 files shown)


pkg/core/index.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	chans.bundleEntries <- bundleEntriesRes{bundleEntries: entries, idx: index, id: id}
+}
+
+// downloadAll downloads all index files in parallel, given an iterator to point to the file location
+func (f *fileIndex) downloadAll(chans downloadBundleFileListChans) {
+	f.indexPather.Reset()
+
+	throttle := make(chan struct{}, f.concurrency)
```
ransomw1c: i think this is a semaphore?

i'm not clear what "throttle" means as a noun in terms of programming.  i usually think throttle is a verb?

not s'th to get too delve into too much either way.

fredbi: > A throttle is the mechanism by which fluid flow is managed by constriction or obstruction. An engine's power can be increased or decreased by the restriction of inlet gases, but usually decreased.

ransomw1c:
reading this some.  my preliminary feedback can be backgrounded until it's out of draft mode.

one question is about the `patherIterator` api design, and there are at least two ways to phrase the question:

why does the indexer return a `func(int) string` instead of an iterator?  doesn't the `func` always get used by iterating from zero to the max value, the second element of the tuple returned by `patherIterator.Next`?

the other way to phrase it is, "what do we mean by indexing, exactly, and is the design conflating indexing as a concept with adjacent ideas like batching?"  like, when i think, "indexing," i'm usually thinking about ways to make database queries more efficient, and that could be what's going on here if we need different ways to slice and dice batches of files, using the same `func(int) string` over different sequences of integers.  if we're always using the same sequence of integers, why not encode that convention as described in the previous paragraph?

----

in the more general setting, this question is returning to the theme of how much to abstract with functions, and i get the sense that some subtle affinities about design decisions that aren't in agreement, where i typically want to abstract out a lot with first order functions and function closures but not quite as much as your designs typically do.

when i ask questions in on this theme, i'm mostly checking that my understanding is correct by asking if the alternate design _could_ suffice rather than suggesting that we _should_ follow the alternate design.

ransomw1c:
still unreasonably preoccupied with the states for diamonds and splits, another quick idea is to represent the state for diamonds, for instance, as a list/slice with values among some enum/typed constants rather than a single enum/type constant value.

so for instance, rather than `DiamondCancelled`, we might have have `{DiamondInitialized, DiamondCancelling, DiamondDone}` and instead of `DiamondDone`, we'd have  {DiamondInitialized, DiamondCommitting, DiamondDone}`.

phrasing this another way, i'm pretty sure the transition matrix between the states as they currently exist is reasonably sparse:  there are some states that don't lead to other states, and it's a programming error if such a state transition occurs.

perhaps, if the states are encoded differently, we can prevent some class of errors, because the state transitions that these errors correspond to cannot possibly occur given the way state is encoded?

fredbi:
@ransomw1c  about the iterator pattern - you rightly pointed out a difficult point.

- my main objective was to reuse code, not so much to reach perfect abstraction.
- "what do we mean by indexing": well this is not quite new, I am reusing the "bundle-files-{index}" stuff
-  code to download or upload the "files list file" is complex and ubiquitous. It was not possible to reuse directly the bundle code (I managed it for a bit, but these functions where needed in many places) 
- so i did it the other way: let's produce things that could be reused by bundles by refactoring bundle download/upload on our next spare cycle
- I am calling "files list file" an index file, mainly because "files list file" kinda stutters, and yes, I've a feeling that this is akin to a database index: you get your hash from there, otherwise, you get lost searching at random for your keys and hashes in the blob storage)

Eventually I came with this "iterator", which is more an "iterator giving an iterator" (but you now know that I wouldn't ever call things like "iteratorOfIterators"...).

It provides the ability to scan over all "bundle-files-{index}.yaml" files when iterated ONCE (typical examples: split upload, bundle upload part in diamond_commit)

In the most general case (collecting all files from all splits), it iterates over all splits which in turn iterate over all their "bundle-files-{index}.yaml" files.

fredbi:
@ransomw1c the transitions indeed follow a very linear pattern.

I'd say that anything that would make this thing resemble a configurable automaton would be harder to read and understand. Anyhow, this is an interesting idea. If you have an example on how this pattern could be coded in a readable way, I am all in. But I am against the typical petri-net lib.


The state stuff is essentially here to protect against re-starting things unattended, or to commit before some splits are completed etc. Checks are implemented at the start of operations at the moment.

To be totally transparent, I am still struggling with some race conditions on the commit part, though... But we can deliver a _workable_ solution with still some small windows where there is a race condition.

fredbi:
> why does the indexer return a func(int) string instead of an iterator or just another Next function that's a closure over an iterator? doesn't the func always get used by iterating from zero to the max value, the second element of the tuple returned by patherIterator.Next?

Mmh yes that is an interesting way of doing it. That would definitely clarify the many returned values. Adopting that. Thanks.

ransomw1c:
even before reading the replies, i'd decided that the state representation stuff suggestion is potentially [cleverness](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=14m35s) [considered harmful](https://go-proverbs.github.io/), although perhaps not.  it could alternately be viewed as [choosing the correct data structures](http://users.ece.utexas.edu/~adnan/pike.html), and i don't have a clear idea what perspective is more accurate.

will avoid additional contemplation of proverbs for awhile.

pkg/core/index.go
```diff
@@ -146,20 +106,22 @@ func (f *fileIndex) getFileIndex(pth string) (model.BundleEntries, error) {
...
+	entries, err := f.getIndexFile(pth)
 	if err != nil {
 		chans.error <- err
 		return
 	}
 
-	// TODO(fred): to be checked out but I don't think the index value in this output is used beyond logging
+	// TODO(fred): nice - diamond does not use the idx field, but bundle_unpack does,
+	// we keep it for this method to be reusable by bundle_unpack at a later time.
+	// Ideally, we shouldn't need that index numbering.
```
ransomw1c: the `idx` struct field was added in case it's necessary to preserve order between requests and results – i.e. whether we need index numbering in the affirmative case is a consequence of preserving sort order.

so i suggest determining whether preserving sort order is desirable, making decision about circulating indices through the channels based on that determination.

pkg/core/index.go
```diff
@@ -11,50 +9,11 @@ import (
...
-	// This iterator may be called concurrently.
-	Next() (string, indexIterator)
-}
-
-// fileIndex builds an index of files from file lists in metadata.
+// fileIndex manages an index of files based on file lists stored in metadata.
+//
+// Indexing is critical to be able to retrieve files from the blob storage.
+//
+// File lists are split into manageable chunks. By default each single index file holds up to 1000 file entries.
```
ransomw1c: based on this comment, i now understand that "index" throughout this patch is a synonym (or near-synonym?) for "bundle filelist" aka `BundleEntries`.  mildly distressed about about having three synonyms, now, for this entity.  meanwhile, "context" could either mean the golang type that essentially amounts to immutable maps or the datamon type that's currently a collection of GCS stores.

i don't have such an issue with these naming conventions when reading the source, yet i worry that conversationally these naming conventions will eventually have a negative impact on the user experience.

this isn't specific to this patch, so i'm not saying this comment need block merge.  feel free to resolve after reading.

pkg/core/bundle_unpack.go
```diff
@@ -155,6 +155,7 @@ func unpackBundleDescriptor(ctx context.Context, bundle *Bundle, publish bool) e
 type bundleEntriesRes struct {
 	bundleEntries model.BundleEntries
 	idx           uint64
+	id            string // the splitID when used in this context
```
ransomw1c: when we start talking about using a structure in multiple contexts, practically, what we have is a [sum type](https://github.com/fantasyland/daggy#daggytaggedsumtypename-constructors) type, also called an [algebraic data type](http://learnyouahaskell.com/making-our-own-types-and-typeclasses#algebraic-data-types), or [tagged union](https://en.wikipedia.org/wiki/Tagged_union), i guess is the wikipedia-official term.

i often find myself reaching for such a paradigm in golang only to recall that it doesn't provide tagged unions as a first-order concept, and [the standard reply](https://github.com/golang/go/issues/19412) about why not is that `interface`s are supposed to be used as functional replacements for tagged unions.

so i suggest doing just that:  create an interface type and use it instead of `bundleEntriesRes` when "context" in the sense of this comment cannot be determined from variables' types.  at the simplest, it could provide a [catamorphism](https://github.com/ransomw/kvsjs/blob/master/src/dsl.js#L26) (i.e. fan-out) from the `interface` type to the original `bundleEntriesRes` and an additional type similar to `bundleEntriesRes` except in contexts where `splitID` is used, although there are other interface designs available.

what i'm saying here, most generally, is, "let's try to encode structure of the source first, then use comments for clarification and generating prose-like documentation."

pkg/model/diamond.go
```diff
@@ -0,0 +1,257 @@
...
+	return &DiamondDescriptor{
+		DiamondID: ksuid.String(),
+		StartTime: GetBundleTimeStamp(),
+		State:     DiamondInitialized,
+		Mode:      EnableConflicts,
+	}
+}
+
+// NewDiamondDescriptor builds a new DiamondDescriptor
+func NewDiamondDescriptor(opts ...DiamondDescriptorOption) *DiamondDescriptor {
```
ransomw1c: for consistency [:zombie:  ... err, ⚡️ , i guess?] , i suggest either placing this in `core` or moving `core.NewBDescriptor` to `model`.

consistency has a very early precedence for me by all indications.  i'm considering consistency more important that separation of ideas by patch here.

fredbi: yes this is a refactoring I have already prepared, but will push that PR _after_ diamonds.

ransomw1c: order of operations considered a nitpick, then.

pkg/core/diamond_commit.go
```diff
@@ -0,0 +1,326 @@
...
+// Cancel a diamond
+func (d *Diamond) Cancel() error {
+	// TODO(fred)
+	return status.ErrNotImplemented.WrapMessage("diamond cancel")
+}
+
+// WithState returns the diamond with its state modified and transition timestamp updated
+func (d *Diamond) WithState(state model.DiamondState) *Diamond {
+	d.DiamondDescriptor.State = state
+	d.DiamondDescriptor.StateChangeTime = time.Now()
```
ransomw1c: realizing i might be entering into the discussion too late on this implementation detail since i wasn't present for all of the design phase:

is there any way we can avoid storing times from the machines where the datamon binary runs and instead use times provided by the GCS service?  that way, we'll have the exceedingly nice property of synchronization.  [`ObjectAttrs`](https://godoc.org/cloud.google.com/go/storage#ObjectAttrs) keeps track of times.  i'd be happier with a patch that tears the `storage.Store` up a little (as we're already doing with conditionals CRC) in order to avoid unsynchronized timestamps appearing in the data sets that it's this program's responsibility to organize and track.

fredbi: I understand the sync issue, but I don't know of any such service specifically for GCP. You might otherwise subscribe to time.google.com as your NTP server. I am guessing that's how our pods are subscribed to.

ransomw1c: > subscribe to time.google.com as your NTP server. I am guessing that's how our pods

to iron out the user story somewhat, can we _either_ make sure

* pods are time-synced (@bluelion ?)

and

* users know (via the markdown in this readme, confluence, or just direct correspondence) that the `diamond` and `split` functionality is mainly intended for in-k8s programs, and we should be notified in #datamon or otherwise if `diamond` and `split` are used from local machines, at least for data sets that might find their way into production

_or_

* include a ± time constant to encode the uncertainty, used along with (or to derive) `diamondCommitLockExpiry` and any other future compairsions with `StateChangeTime`?

pkg/core/diamond_commit.go
```diff
@@ -0,0 +1,326 @@
...
+	if d.BundleID == "" {
+		return nil
+	}
+
+	// case of bundleID preservation (e.g. data migration use case)
+	id, err := ksuid.Parse(d.BundleID)
+	if err != nil {
+		return status.ErrInvalidKsuid.Wrap(err)
+	}
+	d.setBundleID(id.String())
```
ransomw1c: specifically, this line could be removed, except for the data migration?

is it worth feature-flagging this use case?

pkg/core/diamond_commit.go
```diff
@@ -0,0 +1,326 @@
...
+
+	// check when bundle is already specified (e.g. data migration use case)
+	if err = d.checkBundleID(); err != nil {
+		return err
+	}
+
+	logger := d.l.With(zap.String("diamond_id", d.DiamondDescriptor.DiamondID))
+
+	// critical section #1
+	// TODO(fred): better protect critical section
```
ransomw1c: recording consideration of state-management metaphors for GCS while here for future reference:

do lockfiles translate to gcs?

"mutations" on immutable objects are implemented by creating a new object, and immutability has some desirable properties.  so state could be tracked via a directory-like prefix `.../<diamond-prefix>/<state-name>/` and then use some lexicographical ordering (e.g. zero-padded integers) on objects to distinguish the current state from previous states.

fredbi: yes, using such resources as "lock-files" was my initial design.
However, since we can't delete the files, and since state is not always going forward but may revert (from running to initialized, from failed to running, etc), I gave up that way.
The downside of this is now I have a race condition on updating gcs files.

pkg/core/diamond_commit.go
fredbi: indeed you have a point here - I don't have a good solution for it right now. Mentioned it in the detailed design document. See https://github.com/oneconcern/datamon/pull/399

pkg/core/diamond_commit.go
fredbi: added feature guard

pkg/core/bundle_unpack.go
fredbi: With this PR, I am trying to reuse what I can from existing bundle stuff.
A lot of such interesting theoretical points indeed come up to mind when doing so.
The practical outcome is what I did. The comment stuff is left behind so I may refactor this later on.

pkg/core/index.go
fredbi: index stands for "file list file"

pkg/core/index.go
fredbi: This investigation shall be conducted. From what I understand from the existing bundle functionality, this order preserving idea doesn't mean much, but I'll make sure later on.

Indeed, I recently drafted some experimental refactoring of this bundle download/upload stuff, which comes clearer now that I have to implement a more general functionality with 90% overlap. This draft is not published yet.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
+# Diamond workflow support
+
+A diamond workflow allows for a pipeline to scale out the generation of a single bundle.
```
kerneltime: Should we close https://github.com/oneconcern/datamon/pull/399

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+                      [--with-conflicts]   # <- this is the default - overlaps are reported as conflicts detected when merging splits
+
+datamon diamond commit --diamond {diamond-id} \
+                       --with-checkpoints   # <- indicates an incremental upload: conflicts are subsequent versions of uploaded files
+
+datamon diamond commit --diamond {diamond-id} \
+                       --no-conflicts       # <- forbids any overlap to take place at commit time
+
+datamon diamond commit --diamond {diamond-id} \
+                       --ignore-conflicts   # <- do not report about any conflict: intermediate versions
```
kerneltime: This is scope creep, this is also unsafe as a client can lose data and we will allow it.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+with conflicts. See also [§ Handling Conflicts below](#handling-conflicts).
+
+
+Available options at commit time:
+```
+datamon diamond commit --diamond {diamond-id} \
+                      [--with-conflicts]   # <- this is the default - overlaps are reported as conflicts detected when merging splits
+
+datamon diamond commit --diamond {diamond-id} \
+                       --with-checkpoints   # <- indicates an incremental upload: conflicts are subsequent versions of uploaded files
```
kerneltime: I still prefer the client reusing the same split-id and letting the most recent file win. I don't yet see the reason to have this flag.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+   This is a consequence of the blob deduplication strategy (the same blob might be claimed by another file, so we can't
+   mark it as something to be deleted).
+
+
+Design:
+
+* Diamonds abide by the following state-transition model:
+  1. `initialized`
+  2. `committing`
+  3. `done`
```
kerneltime: Instead of tracking this as fields in a single yaml file which will require a Read-Modify-Write cycle which under concurrent load is hard to get right. Use separate files and use the put if not present primitive. Also we only need init and done.

kerneltime: I don't see the need for a 2 phase commit verbiage.

docs/proposals/diamond-design.md
```diff
@@ -0,0 +1,562 @@
...
+
+2. Concurrent splits.
+  * Each `split add` command runs with its uniquely generated internal ID: metadata can be safely modified within this namespace.
+
+3. Commit consistency.
+  * The diamond is in effect concluded with the first commit associated with it. No further action is allowed on a committed diamond.
+  * Any attempt to commit while another commit is in progress is rejected (the operation should normally be fast
+    since only metadata is affected).
+  * For the duration of the commit operation, the diamond is in effect locked in "committing" status.
+  * To recover from uncontrolled failures during a commit (e.g. pod crash,...) this lock is considered stale after some
```
kerneltime: This is not needed, we only need a single done.yaml file in the path with the bundle ID in it to mark commit. The put if not present primitive of cloud storage APIs will provide the necessary protection.

kerneltime:
Not completed the review, have some concerns about state tracking which seems redundant and more than what is needed without serving a purpose.

docs/proposals/diamond-design.md
fredbi: I made (and sync'ed) all these PRs to produce a _reviewable_ ensemble.
I'd prefer that we merge them one by one following informed and understood agreement.

docs/proposals/diamond-design.md
fredbi: I can disable this if you prefer.

docs/proposals/diamond-design.md
fredbi: _verbiage_ is perhaps not carrying the same meaning. Certainly not when compared with some _actual_ verbiage that clutters the other reviews on this series of PR.
Best to discuss that part separately.

fredbi: moved back to fewer states for diamond: initialized (aka running), done, canceled

and for splits: running, done

fredbi: Even though the key name now holds the state semantics, I am keeping the state encoded in the metadata for easier retrieval (it is a bit of a pain to query the state of things when all is encoded as keys)

docs/proposals/diamond-design.md
fredbi: All that part about locking is indeed no more needed since the file no more needs to be locked.

docs/proposals/diamond-design.md
fredbi: --ignore-conflict is not unsafe: the result is the same with all flags. The only difference lies in the presence of /.conflicts, /.checkpoints "hidden" folders.

pkg/core/split.go
```diff
@@ -0,0 +1,248 @@
...
+
+	if s.SplitDescriptor.Tag != "" {
+		s.l = s.l.With(zap.String("tag", s.SplitDescriptor.Tag))
+		s.Bundle.l = s.l
+	}
+	return s
+}
+
+// Upload a dataset as a split, without committing as a bundle
+func (s *Split) Upload(opts ...Option) error {
```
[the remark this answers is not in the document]
fredbi: yes, it is part of the interface: it is called by the CLI (next PR)

pkg/core/split.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/oneconcern/datamon/pkg/model"
+	"github.com/oneconcern/datamon/pkg/storage"
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
+)
+
+// Split is a generalization of a Bundle, to support diamond operations
+type Split struct {
+	metaObject
+	*Bundle
```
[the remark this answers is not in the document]
fredbi: this is type composition: it is quite different from `metaObject *Bundle`
Basically this means that a split is composed of a `metaObject` and a `*Bundle`
which also means a split inherits from all their methods.

pkg/core/split.go
```diff
@@ -0,0 +1,248 @@
...
+	uploadIndexer   *fileIndex
+	downloadIndexer *fileIndex
+	mustExist       bool // used to check replayed splits with splitID forced
+}
+
+func defaultSplit(repo, diamondID string, stores context2.Stores) *Split {
+	empty := NewBundle(NewBDescriptor(), Repo(repo), ContextStores(stores))
+	split := model.NewSplitDescriptor()
+	return &Split{
+		metaObject:      defaultMetaObject(stores.VMetadata()), // splits live on vmetadata
```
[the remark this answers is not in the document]
fredbi: yep sure. My bad here.

pkg/core/diamond_commit.go
```diff
@@ -0,0 +1,418 @@
...
+					continue
+				}
+
+				existing := obj.(mergeEntry)
+				if file.Hash == existing.Hash {
+					continue
+				}
+
+				if file.SnapshotTime.IsZero() {
+					panic("dev error: files should have a timing") // internal safeguard
```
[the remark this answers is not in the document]
fredbi: yep sure

Issue #53, opened 2018-11-26
Improvements to Key enumeration of store.go


kerneltime:
The localfs implementation needs to

- [ ] More efficiently walk the space for localfs (faster is better)

- [ ] The API should allow for pagination to walk an arbitrarily large number of files.

- [ ] Support for prefix key enumeration

fredbi:
localfs case addressed by #404 

We might still worry about the overall efficiency of key walking on a localfs. I don't see any better solution for now.

fredbi:
TODO: in core, replace calls to `store.Keys` by iterators.
Plus: should deprecate `Keys` in storage.Store

PR #404, opened 2020-02-14
Diamonds #3: fix(localfs)

fredbi:
fix(localfs): fixed Put flags: added O_TRUNC (avoids file corruption when updating a file), removed 0600
fix(localfs): implemented KeysPrefix method (cursory implementation, primarily intended to use localfs as a mocked up Store)
fix(localfs): added the WithLock option to guard localfs against concurrent R/W and W/W options

**Motivation:**
I am building up some elaborate testcases for the diamonds feature, based on mocked localfs.
This brought up several issues with concurrent updates and required an implementation of KeysPrefix.

The new KeyPrefix is certainly not very scalable, but at least it
supports my testcases with core. I am not sure there is a better
way to go with afero.

### DIFF
--- a/pkg/storage/localfs/store.go
+++ b/pkg/storage/localfs/store.go
@@ -8,7 +8,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
+	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -18,17 +21,36 @@ import (
 )
 
 // New creates a new local file system backed storage model
-func New(fs afero.Fs) storage.Store {
+func New(fs afero.Fs, opts ...Option) storage.Store {
 	if fs == nil {
 		fs = afero.NewBasePathFs(afero.NewOsFs(), filepath.Join(".datamon", "objects"))
 	}
-	return &localFS{
-		fs: fs,
+	local := &localFS{
+		fs:   fs,
+		glob: make(map[string][]string),
+	}
+	for _, apply := range opts {
+		apply(local)
+	}
+	return local
+}
+
+// Option for the local FS store
+type Option func(*localFS)
+
+// WithLock prevents concurrent writes or concurrent read/writes on this local FS
+func WithLock(flag bool) Option {
+	return func(fs *localFS) {
+		fs.lock = flag
 	}
 }
 
 type localFS struct {
-	fs afero.Fs
+	fs        afero.Fs
+	glob      map[string][]string // current state of KeyPrefix matches
+	exclusive sync.Mutex          // mutex on glob access
+	lock      bool
+	rw        sync.RWMutex
 }
 
 func (l *localFS) Has(ctx context.Context, key string) (bool, error) {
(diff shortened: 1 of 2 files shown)


pkg/storage/localfs/store.go
```diff
@@ -155,9 +158,70 @@ func (l *localFS) Keys(ctx context.Context) ([]string, error) {
...
+// KeyPrefix provides a paginated key iterator using "pageToken" as the next starting point
+//
+// NOTE: "delimiter" is ignored (always set to "/").
+//
+// TODO: This implementation does not really scale up, but it is quite workable for our testcases using locafs.
+func (l *localFS) KeysPrefix(_ context.Context, token, prefix, _ string, count int) ([]string, string, error) {
+	l.exclusive.Lock()
+	defer l.exclusive.Unlock()
+
+	prefix = path.Clean("/" + prefix)
```
ransomw1c: Afero api reminders:

* even in case of `BasePathFS`, `afero.Walk` can accept an absolute path as `root`
* when `Walk` is provided an absolute path, the `WalkFunc`'s `pth` parameter is also an absolute path

ransomw1c:
well, i got the notification and started reviewing this before noticing that i'm not added as a reviewer.

anyway, this'd be a good addition, imo.

only caveat to mention is slight mismatch with gcs regarding the meaning of `count`:  in this local fs, `count` means "this is the most that could be returned per call" whereas in the gcs implementation, i'm pretty sure `count` means "precisely this many entries will be returned per call, provided that many exists after the given page token".  i've traced this to [`NewPager`](https://godoc.org/google.golang.org/api/iterator#NewPager) where 

> `NextPage` method will obtain exactly `pageSize` items

and the `count` value is what's being passed as `pageSize`.

i'm okay with having the `storage.Store` api encompass both meanings of `KeysPrefix` such that the only action to suggest is updating the godoc strings.

fredbi:
yes there are still a bug or two in here. I am proceeding with more thorough testing.

fredbi:
I am also contemplating an implementation based on radixTree, without the cache (which helps performance but is also causing concurrency issues)

pkg/storage/localfs/store.go
fredbi: yes I've tried to normalize things here: input always goes absolute, output always go relative...

pkg/storage/localfs/store.go
```diff
@@ -123,13 +154,21 @@ func (l *localFS) Put(ctx context.Context, key string, source io.Reader, exclusi
...
+		defer l.rw.Unlock()
+	}
 	if err := l.fs.Remove(key); err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("removing %q: %v", key, err)
 	}
 	return nil
 }
 
 func (l *localFS) Keys(ctx context.Context) ([]string, error) {
+	if l.lock {
```
kerneltime: Why lock when listing keys?

pkg/storage/localfs/store.go
```diff
@@ -123,13 +154,21 @@ func (l *localFS) Put(ctx context.Context, key string, source io.Reader, exclusi
 }
 
 func (l *localFS) Delete(ctx context.Context, key string) error {
+	if l.lock {
```
kerneltime: unlink is atomic

pkg/storage/localfs/store.go
fredbi: consistent read

pkg/storage/localfs/store.go
fredbi: true. shall remove the lock here.

fredbi:
fixes #53

PR #405, opened 2020-02-17
Diamond #5: CLI - READY

fredbi:
### NOTE to reviewers
This PR integrates all the work from diamond PRs #1, #2, #3, and #4 
You might want to view the last commits only, starting at https://github.com/oneconcern/datamon/pull/405/commits/1762acf5213b9ae3cb3402dd1124e4dbf608c36d

### DIFF
--- a/cmd/datamon/cmd/cli_fuse_test.go
+++ b/cmd/datamon/cmd/cli_fuse_test.go
@@ -16,7 +16,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/oneconcern/datamon/internal/rand"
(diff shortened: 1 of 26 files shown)


ransomw1c:
i can already comment here that i'd probably like to have an optional parameter to `diamond initialize` that takes a path on the local system and writes the diamond id used to initialize to a file at that location.

i want something similar for `bundle upload` (among Jira issues for sidecar post-Zsh prototyping), so perhaps these both can be added as a separate request.

kerneltime:
I looked into the basic code path for init add and commit and it looks good.
We can merge it in and start testing before cutting a release.

PR #409, opened 2020-02-20
refact(core): factorized object listing options

fredbi:
## refact(core): factorized object listing options

* single algorithm to list metadata objects

TODOs deferred to forthcoming refactoring action:
* refact listing tests (comprehensive mocked up test may be done one, in the new core/lister package)
* use errGroup pattern to simplify error handling on goroutines


ransomw1c:
i think this patch is mostly abstracting out a concurrency pattern, one that could be used to download lists of metadata – and other things as well.

what about using the same concurrency pattern (fan out on buffered channel semaphores + collect results using `count` – cf. WAL implementation) for listing that other operations (including the forthcoming diamond functionality) use as a way to use a single concurrency pattern where applicable throughout?

iirc, the reason the buffered channel semaphore pattern wasn't used here was the need to preserve the sort order in results (and because we were interested in getting a couple of sample concurrency patterns in the source tree for comparison).  there was also a comment about removing the index field on the struct used to report bundle and diamond upload results.  that index was added in case it's necessary to preserve sort order, this was to be used in the results collection thread.

as is, this patch strikes me mostly as dsl-ification of the internal apis.  it's not something that i have i problem with, exactly, except the order of operations is slightly eyebrow-raising.

the missing magic here, i think, is not standardization via dsl.  it's standardization via reasoned consideration, _then_ dsl-ifying the reasoned consideration.  i'm afraid if that if we dsl-ify first it could result in a bias to use the dslified thing for convenience rather than pushing hard on deduplicating procedural encodings throughout.

fredbi:
> iirc, the reason the buffered channel semaphore pattern wasn't used here 

This is a different topic: changing the algorithm is not within the scope of this PR.

If you want to change this pattern later on, this PR makes this easier by putting the algorithm in one single place.

> was the need to preserve the sort order in result

yes. And this requirement did make all the difference.

> what about using the same concurrency pattern (fan out on buffered channel semaphores + collect results using count – cf. WAL implementation
Maybe. Not in my current plans anyhow.

> as is, this patch strikes me mostly as dsl-ification of the internal apis.

I don't understand this

> ... is not standardization via dsl. it's standardization via reasoned consideration, then dsl-ifying the reasoned consideration

I don't understand this either. I fail to establish any connection with the proposed code.

### My objective here is simple and clear: to factorize an algorithm that currently works into one single piece of code.

### Please either approve or request pragmatic changes

fredbi:
> iirc, the reason the buffered channel semaphore pattern wasn't used here 

This is a different topic: changing the algorithm is not within the scope of this PR.

If you want to change this pattern later on, this PR makes this easier by putting the algorithm in one single place.

> was the need to preserve the sort order in result

yes. And this requirement did make all the difference.

> what about using the same concurrency pattern (fan out on buffered channel semaphores + collect results using count – cf. WAL implementation
Maybe. Not in my current plans anyhow.

> as is, this patch strikes me mostly as dsl-ification of the internal apis.

I don't understand this

> ... is not standardization via dsl. it's standardization via reasoned consideration, then dsl-ifying the reasoned consideration

I don't understand this either. I fail to establish any connection with the proposed code.

### My objective here is simple and clear: to factorize an algorithm that currently works into one single piece of code.

### Please either approve or request pragmatic changes

ransomw1c:
recording clarified review from call:  i'd prefer not to merge this now.  i do not think it's _currently_ a maintainability improvement.  i'd prefer to keep the branch on the remote to rebase against and merge in sooner rather than later.

PR #412, opened 2020-02-24
feat(ops): [RES-2173] Cleaned up circleci details

joaoh821C:
**Description:**
Datamon being an Open Source Project we don’t want any One Concern details revealed to the public anywhere in the code base. And we noticed that on the Circle CI yaml file there are a couple of parameters that are revealed and should not be.

Got rid of:
* Google Project ID
* Cluster Name

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,6 @@ executors:
   docker_minio:
     working_directory: ~/project
     environment:
-      GOOGLE_PROJECT_ID: "onec-co"
       GOOGLE_COMPUTE_ZONE: us-west2-c
     docker:
       - image: circleci/golang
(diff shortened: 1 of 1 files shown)


fredbi:
ok. @joaoh821C could you please explain which variables are set from context?
I believe we have a few k8-aware CI jobs there (fuse demo, etc) and it would be nice to know more about what is available from context.

fredbi:
@joaoh821C don't worry about the linting: linter has moved up its version unattended. I'll fix this today so you may rebase

fredbi:
@joaoh821C I fixed the issue and checked that your work could be merged...
Unfortunately, there is a DCO issue, which prevents me from merging.
Could you please rebase this with a single PGP-signed commit?
Thanks for your contribution.

Cheers

Frederic

fredbi:
@joaoh821C yay! merging now

PR #413, opened 2020-02-24
init web uploader proposal

ransomw1c:
quick idea motivated by how to smooth out the "upload from local" user story.

the idea is that, supposing 

* we have the functionality described in this proposal
* users can upload data from their local to somewhere in the [k8s] cluster accessible from a pod's filesystem (via Persistent Volume Claim or otherwise)

then we can have a managed "ad-hoc cloud upload" user story that covers everything except the last mile of the "upload from local" story and also provides similar affordances to the "Jupyter as a service" story that we've discussed a few times in standup.

### DIFF
--- a/docs/proposals/web_uploads.md
+++ b/docs/proposals/web_uploads.md
@@ -1,0 +1,34 @@
+# Add upload scheduler to webservice
+
+This is a proposal to facet `bundle upload` via the Datamon webservice (i.e. the `web` command).
+
+Additionally, so that the web service -- currently a frontend for `repo list`, `bundle list`, and `bundle list files`, all operations that typically complete in per-HTTP-request goroutines in about the expected timeframe for page loads in a 1.0-ish webservice -- can continue to be useful for a possible deploy to Kubernetes, faceting `bundle upload` necessarily includes implementing some at least rudimentary scheduling functionality:  The webservice will need to keep track of several ongoing uploads, each triggered by an HTTP-request, but with fire-and-forget (i.e. event) semantics rather than those of request-response (i.e. RPC).
+
+### data sources
+
+This proposal calls for an additional, optional flag to the `web` command, `--upload-source`, say, which can be present in arbitrary arities (to allow arbitrary numbers of upload sources -- is this possible with pflags?).  When this flag isn't present, the webservice will continue to function as before.  When a non-empty list of directories is passed in via this flag, an additional page will be available in the web ui that, at first, provides a simple file browser rooted at each of these directories.
+
+From within any navigable directory of the filebrowser-as-a-webservice, it will additionally be possible to click an "upload" button to schedule a `bundle upload` of the tree rooted at that directory.  The "upload" button will be attached to a form faceting a few additional `bundle upload` parameters:
+
+* `--concurrency-factor`
+* `--message`
+* `--repo` (dropdown)
+* `--label`
+
+### scheduler
+
+Round-robin scheduling with a heartbeat
+[as described here](https://github.com/ransomw/tubing/blob/master/src/core.clj)
+is expected to be as complicated as a first pass of the scheduler internals will be.  That is to say, every effort will be made to decomplect scheduler design at the possible expense of performance.
+
+In addition to the filebrowser views described in the data sources section, the web UI will provide a flat list of all uploads according to source directory and destination description (repo, message, label), including simple status (complete, error, ongoing).
+
+## nice-to-haves
+
+### downloadable logs
+
+The web UI will provide a flat list of all uploads, including logging  the output from `pkg/core` and beyond for download.
+
+### build filelists
+
+currently, bundle structure is expected to be described entirely by filesystem layout.


docs/proposals/web_uploads.md
```diff
@@ -0,0 +1,34 @@
+# Add upload scheduler to webservice
+
+This is a proposal to facet `bundle upload` via the Datamon webservice (i.e. the `web` command).
+
+Additionally, so that the web service -- currently a frontend for `repo list`, `bundle list`, and `bundle list files`, all operations that typically complete in per-HTTP-request goroutines in about the expected timeframe for page loads in a 1.0-ish webservice -- can continue to be useful for a possible deploy to Kubernetes, faceting `bundle upload` necessarily includes implementing some at least rudimentary scheduling functionality:  The webservice will need to keep track of several ongoing uploads, each triggered by an HTTP-request, but with fire-and-forget (i.e. event) semantics rather than those of request-response (i.e. RPC).
```
fredbi: An alternative could be python-based SDK "async bundle upload"

docs/proposals/web_uploads.md
```diff
@@ -0,0 +1,34 @@
...
+
+### data sources
+
+This proposal calls for an additional, optional flag to the `web` command, `--upload-source`, say, which can be present in arbitrary arities (to allow arbitrary numbers of upload sources -- is this possible with pflags?).  When this flag isn't present, the webservice will continue to function as before.  When a non-empty list of directories is passed in via this flag, an additional page will be available in the web ui that, at first, provides a simple file browser rooted at each of these directories.
+
+From within any navigable directory of the filebrowser-as-a-webservice, it will additionally be possible to click an "upload" button to schedule a `bundle upload` of the tree rooted at that directory.  The "upload" button will be attached to a form faceting a few additional `bundle upload` parameters:
```
fredbi: I think a similar result, with more powerful generalization, could be achieved by upgrading our RW mount,
making it capable of _incremental uploads_ (newly available capability with diamond workflow).

The browser use case would be covered by simply adding a "commit" button and possibly some nice to have visual hint to figure out new/changed files on a view of the local file system (i.e. URL like file://home/my-dataset).

There would be no "upload" button: just use native browser or possibly some pretty standard file-dropping JS widget, but no interaction with `web serve`. There would be no async logic or scheduling whatsoever.

docs/proposals/web_uploads.md
```diff
@@ -0,0 +1,34 @@
...
+* `--concurrency-factor`
+* `--message`
+* `--repo` (dropdown)
+* `--label`
+
+### scheduler
+
+Round-robin scheduling with a heartbeat
+[as described here](https://github.com/ransomw/tubing/blob/master/src/core.clj)
+is expected to be as complicated as a first pass of the scheduler internals will be.  That is to say, every effort will be made to decomplect scheduler design at the possible expense of performance.
```
fredbi: My alternative proposal above gets rid of this requirement.

docs/proposals/web_uploads.md
```diff
@@ -0,0 +1,34 @@
...
+
+In addition to the filebrowser views described in the data sources section, the web UI will provide a flat list of all uploads according to source directory and destination description (repo, message, label), including simple status (complete, error, ongoing).
```
fredbi: My alternative proposal above transforms this in a client-side only affair, possibly handled by a local js gadget.
I am suggesting the upload / commit status on a RW mount to be reflected by some extra attributes of the file.

docs/proposals/web_uploads.md
fredbi: Another way, that I am developing a bit below, could be an enhanced mutable mount feature.
At the moment, the mutable mount is only experimental, still has some bugs, and overall probably not used at all.

This could change by building on top of it some more user-friendly features like the one described in this paper.

docs/proposals/web_uploads.md
fredbi: Proposed sample command: `datamon bundle mount mutable --incremental`: starts a a fuse mount with a diamond, which ID is dumped in some local dot file and may be reused (and renewed) by subsequent commit signals.

Your idea of the local web service could come in handy here to signal the commit by a HTTP command (e.g. POST http://localhost:1234/commit) rather than by an OS signal.

docs/proposals/web_uploads.md
```diff
@@ -0,0 +1,34 @@
...
+
+## nice-to-haves
+
+### downloadable logs
+
+The web UI will provide a flat list of all uploads, including logging  the output from `pkg/core` and beyond for download.
```
fredbi: The mention of this here makes me think I might not have understood the goal _at all_. Please clarify the intended goal or goals.

1. The functionality described here _could be_ a web UI to replace _local_ `bundle upload` actions by async uploads.
2. On the other hand, there are some hints in the text showing that _it could be_ just as well a proxy to manage a datamon instance running on a remote pod.

fredbi:
How the proposed functionality covers the mentioned use-cases is not quite obvious to me.

I think you are on something here, but use-cases should definitely be separated and clarified.

>    users can upload data from their local to somewhere in the [k8s] cluster accessible from a pod's filesystem (via Persistent Volume Claim or otherwise)

Are we talking about some remote controlling service running on the pod? I.e, is this about providing a UI to a datamon running on a pod? The log access proposed feature advocates for this interpretation, although the "async upload" feature seems to be a purely local thing.

> then we can have a managed "ad-hoc cloud upload" user story 

What does this imply, that the CLI does not already handle?

> that covers everything except the last mile of the "upload from local" story 

Isn't precisely that last mile this document is about? This sentence confuses me.

> and also provides similar affordances to the "Jupyter as a service" story that we've discussed a few times in standup.

Indeed, the Jupyter keyword comes back in talks from time to time. This might be a good selling point, but it is not clear to me how what is proposed here relates to that topic.

Could you please share your knowledge on that specific part, so we are in a better position to assess how what is proposed matches the use case of a data scientist using Jupyter with datamon.

docs/proposals/web_uploads.md
ransomw1c: > async bundle upload

whether we implement in Python or not, i am indeed viewing this more in the async/coroutine style: heartbeats are conceptually adjacent to the event loop that coroutines usually are based on.

> could be an enhanced mutable mount feature

yerp, there are lots of ways to translate the motivation in the description of this merge request to functionality to functionality that might or might not eventually get used to assuage the essential pain point (real or perceived).

docs/proposals/web_uploads.md
ransomw1c: `datamon bundle mount mutable --incremental` i agree:  incremental upload functionality would better be faceted behind a ui other than diamond uploads.

if the faceted incremental upload can be used a starting point for the sort web service described here, so much the better.  i'm viewing this comment as supplementary information about how the implementation details here could differ, not as a total replacement for the webservice being described, which is not envisioned a local service on users' laptops but rather a service that we could run in any or all of the various stages' clusters to provide something more like the Jupyter "cleanroom" discussed in standup.

docs/proposals/web_uploads.md
ransomw1c: indeed, it is confusing b/c the user story is about local uploads, but the web service is intended to run on a remote pod, not users' laptops.

the heart of the matter is that, real or imagined, there's sometimes a perception that "upload to service XYZ is super-easy, why can't datamon uploads be just as easy?"  with the functionality in this proposal, we can address such comments by saying "ok, continue to use the XYZ uploads your familiar with as a matter of getting the bytes from here to Iowa, then please additionally click a button over here to upload the data into datamon."

there's a feeling that, independent of technical merits, answering "can i haz?" questions in the affirmative with an "if" is usually better received than replying in the negative with a "but" as a social matter -- needless to say, i could easily be incorrect about these ui design considerations.

ransomw1c:
it's like the "Jupyter as a service" story insomuch as one idea to address that story was rephrased as "a timesharing service within k8s," and when we start talking about a program with a scheduler, we're talking, essentially, about a timesharing setup, even if it's faceted such that the service doesn't share arbitrary cpu time but instead the cpu time required for the functional equivalent of `bundle upload`.

ransomw1c:
would suggest this be reviewed for unspecified future implementation if the time cost isn't too great?

i can also summarily close the request if there's no particular interest.

ransomw1c:
closing this out per recent-ish decision to background datamon features except those specifically requested by data-science

ransomw1c:
pls keep the branch on the remote, tho

PR #415, opened 2020-02-27
test: replaced test random generator

fredbi:


ransomw1c:
paging @joaoh821C here as well.  FYI, this ci job
https://app.circleci.com/jobs/github/oneconcern/datamon/6036
required a manual re-run.

suspect that this error is due to an error out of `apk` (the Alpine distro's package manager) due to some downtime from the Alpine servers, unlike the pipelines linter error, yet want to mention it for completeness.

PR #419, opened 2020-03-04
storage.Store versions

ransomw1c:
##### background on GCS versioning

since the WAL (#271), metadata (the stuff that bridges the CLI and CAFS) is stored in two GCS buckets, one with and one without [object versioning](https://cloud.google.com/storage/docs/object-versioning) activated.  what object versioning means is that GCS keeps track of everything (every object, a near synonym for file) that ever gets stored on it (unless the object is deleted).   that is, overwriting an object in a bucket with versioning activated doesn't delete the previous object.

i don't fully understand all the details of GCS versioning.  the upshot is that with versioning activated, every object _can_ be addressed by `(name, version)` instead of just `(name)`.  objects can still be addressed by name, and in this case the latest version is retrieved.  also instead of "version", the value used to address previous versions is called (by GCS), the "generation", a [non-negative] `int64` (`<rant>`following a common golang tendency, also not fully understood, to use signed integers when unsigned integers would not only suffice but also provide additional space.  neither @kerneltime nor myself grok this.`</rant>`)

----

anyway, the `label`s already familiar to datamon users wind up in the versioned metadata, not the unversioned metadata.  what this means is that we have track of every bundle a label has ever referred to already because GCS is storing that information.  what we need is a way to deliver a label's history to the user.

this patch updates `storage.Store`, datamon's internal interface to GCS, to support versions.  allowing datamon to support versions is a necessary (and obviously _not_ sufficient) condition for implementing something like

```
label list --repo <repo-name> --name <label-name>
```

at the command line interface.

### DIFF
--- a/hack/build_artifacts.sh
+++ b/hack/build_artifacts.sh
@@ -1,0 +1,44 @@
+#! /bin/zsh
+
+setopt ERR_EXIT
+setopt PIPE_FAIL
+
+perhaps_module_dirs=(pkg/)
+
+for perhaps_module_dir in $perhaps_module_dirs; do
+    find $perhaps_module_dir -type d |while read pkg_dir; do
+        if find $pkg_dir -maxdepth 1 \
+                |grep -q '_test.go$'; then
+            print -- "building $pkg_dir test binary"
+            (cd $pkg_dir \
+                 && go test -c)
+        fi
+    done
+done
+
+# ./deploy ???
+perhaps_main_root_dirs=(cmd/ internal/ hack/ pkg/)
+
+typeset -a mains
+
+for perhaps_main_root_dir in $perhaps_main_root_dirs; do
+    find $perhaps_main_root_dir -type d \
+        |while read perhaps_main_dir; do
+        (cd $perhaps_main_dir \
+             && find $ -type f \
+                 |while read perhaps_main; do
+                 if cat $perhaps_main \
+                         |grep -q '^package\s*main'; then
+                     mains=($mains $perhaps_main)
+                 fi
+                 if [[ $#mains -eq 1 ]]; then
+                     print -- "attempting $$perhaps_main build"
+                     go build
+                 else
+                     if [[ $#mains -eq 0 ]]; then continue; fi
+                     print -- "unsure how to build multiple mains" \
+                           "in same directory $perhaps_main_dir build"
+                 fi
+             done)
+    done
+done
--- a/pkg/storage/gcs/store.go
+++ b/pkg/storage/gcs/store.go
@@ -13,6 +13,7 @@ import (
 	gcsStorage "cloud.google.com/go/storage"
 
 	"github.com/oneconcern/datamon/pkg/dlogger"
+	"github.com/oneconcern/datamon/pkg/errors"
 	"github.com/oneconcern/datamon/pkg/storage"
 	storagestatus "github.com/oneconcern/datamon/pkg/storage/status"
 	"google.golang.org/api/option"
@@ -24,6 +25,7 @@ type gcs struct {
 	bucket         string
 	ctx            context.Context
 	l              *zap.Logger
+	s              storage.Settings
 }
 
 func clientOpts(readOnly bool, credentialFile string) []option.ClientOption {
@@ -39,11 +41,20 @@ func clientOpts(readOnly bool, credentialFile string) []option.ClientOption {
 	return opts
 }
 
-// New builds a new storage object from a bucket string
-func New(ctx context.Context, bucket string, credentialFile string, opts ...Option) (storage.Store, error) {
+// implNew provides stricter typing than New for package internal tests.
+// Akin to the impl* functions in pkg/core/bundle.go, this function is in place
+// in order to be extended with additional parameters and/or return values
+// independently of the public interface provided by New.
+func implNew(ctx context.Context, bucket string, credentialFile string, opts ...Option) (*gcs, error) {
 	googleStore := new(gcs)
 	for _, apply := range opts {
 		apply(googleStore)
+	}
+	{
+		generationNumber := googleStore.s.Version.GcsVersion()
+		if generationNumber < 0 && generationNumber != storage.GcsSentinelVersion {
+			return nil, errors.New("invalid version")
+		}
 	}
 	if googleStore.l == nil {
 		googleStore.l, _ = dlogger.GetLogger("info")
@@ -62,6 +73,11 @@ func New(ctx context.Context, bucket string, credentialFile string, opts ...Opti
 		return nil, toSentinelErrors(err)
 	}
 	return googleStore, nil
+}
+
+// New builds a new storage object from a bucket string
+func New(ctx context.Context, bucket string, credentialFile string, opts ...Option) (storage.Store, error) {
+	return implNew(ctx, bucket, credentialFile, opts...)
 }
 
 func (g *gcs) String() string {
@@ -110,8 +126,8 @@ func (r *gcsReader) ReadAt(p []byte, offset int64) (n int, err error) {
 	defer func() {
 		r.l.Debug("End ReadAt", zap.Int("chunk size", len(p)), zap.Int64("offset", offset), zap.Int("bytes read", n), zap.Error(err))
 	}()
-	objectReader, err := r.g.readOnlyClient.Bucket(r.g.bucket).Object(r.objectName).NewRangeReader(
-		r.g.ctx, offset, int64(len(p)))
+	objectReader, err := r.g.readOnlyClient.Bucket(r.g.bucket).Object(r.objectName).
+		NewRangeReader(r.g.ctx, offset, int64(len(p)))
 	if err != nil {
 		return 0, toSentinelErrors(err)
 	}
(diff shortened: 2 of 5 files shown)


pkg/core/options.go
```diff
@@ -30,49 +32,56 @@ var (
 )
 
 // ConcurrentList sets the max level of concurrency to retrieve core objects. It defaults to 2 x #cpus.
-func ConcurrentList(concurrentList int) Option {
+func ConcurrentList(v int) Option {
 	return func(s *Settings) {
-		if concurrentList == 0 {
+		if v == 0 {
```
fredbi: the general idea when renaming identifiers was to be brief, not terse. This is just code obfuscation. I sense some irony behind this change...

pkg/core/options.go
```diff
@@ -30,49 +32,56 @@ var (
...
-func BatchSize(batchSize int) Option {
+func BatchSize(v int) Option {
 	return func(s *Settings) {
-		if batchSize == 0 {
+		if v == 0 {
 			s.batchSize = defaultBatchSize
 			return
 		}
-		s.batchSize = batchSize
+		s.batchSize = v
```
fredbi: the rule of thumb to find a good identifier is that is should remain short and expressive. `v` is short, but not expressive. `theBatchSizeThatTheUserMayDefineFromAFlag` could be slightly more expressive, but definitely not short.
Please refrain here.

pkg/storage/store.go
```diff
@@ -55,6 +55,14 @@ type StoreCRC interface {
 	PutCRC(context.Context, string, io.Reader, bool, uint32) error
 }
 
+type StoreVersioned interface {
```
fredbi: `VersionedStore` ?

pkg/core/options.go
```diff
@@ -17,6 +17,8 @@ type Settings struct {
 	profilingEnabled bool
 	memProfDir       string
 
+	label string
```
fredbi: Is this param here to indicate that we want to retrieve all versions of the listed?
Could it mean something like `wantsLabelVersions`?
This works, but why don't we simply add this as an ordinary parameter to the `List` functions? (I believe the change on the CLI pkg is pretty much localized - there is probably one single call).

pkg/core/label_list.go
```diff
@@ -133,7 +139,15 @@ func listLabelsChan(repo string, stores context2.Stores, prefix string, opts ...
 	keysChan := make(chan keyBatchEvent, 1)
 
 	iterator := func(next string) ([]string, string, error) {
-		return GetLabelStore(stores).KeysPrefix(context.Background(), next, model.GetArchivePathPrefixToLabels(repo, prefix), "", settings.batchSize)
+		if settings.label != "" {
+			return GetLabelStore(stores).KeysPrefix(context.Background(), next,
+				model.GetArchivePathPrefixToLabels(repo, prefix),
+				"", settings.batchSize)
+		} else {
+			versions, versionsErr := GetLabelStore(stores).(storage.StoreVersioned).
```
fredbi: why do we force an error here, since the underlying store may be legit not implementing this interface?
I'd suggest falling back on the previous implementation when `KeyVersions` is not available.

pkg/storage/store.go
fredbi: otherwise, adding this new interface is the way to go

pkg/core/options.go
ransomw1c: no gaslighting on this end of the wire:

this _was_ motivated by my understanding of your preferences for terser names.  it is _not_ intended to be ironic.  the `v` for "value" name is rather an attempt for consistency with [the convention for type assertions](https://github.com/ransomw/exercises/blob/master/gotour/mni/main.go#L231) in the golang tour.

pkg/core/options.go
ransomw1c: i will revert the renaming, nonetheless, of course.

pkg/storage/store.go
ransomw1c: > adding this new interface is the way to go

👍 ty for the sanity check

pkg/core/options.go
ransomw1c: > Could it mean something like `wantsLabelVersions`?

yep, that's what this means.  i'm proceeding with "the same `struct` can have slightly different meanings in different places, and that's okay," idea.  note also, the

```
todo: validation method on settings (specificify settings type per use-case).
```

comment. i'm still slightly convinced that distinguishing the semantics of the `struct` by how and where its instances appear in the source is a good idea, although it's not the focus of this request.

----

about whether to use `... Options` -> `Settings` or not, i was actually thinking in the opposite direction.

what about passing what is currently the `prefix` parameter to `listLabelsChan` as one of the varadic `Option` parameters?

pkg/core/label_list.go
ransomw1c: i am not sure what you mean by "force an error".  if `versionsErr` is `nil`, then `listLabelsChan` also returns `nil` error.

pkg/core/options.go
fredbi: ok we are good then

pkg/core/options.go
fredbi: ok. Either way makes sense. The only inconsistent situation is when we have some args via func (like the current `prefix`) and some args via opts (like your `label` aka `wantsVersions bool`). Either consistent approach is fine by me.

pkg/core/label_list.go
fredbi: `versionErr!= nil` here means that the type assertion failed. This means that the underlying storage does not support versioning. I am not seeing this as a sufficient cause for returning an error and stopping the process.

Another way of seeing this is to keep the error (as you do), and document with the option (or func arg, see discussion above) that 
_this feature is not supported for non-versioned stores_ (that is localfs and our current implementation of the S3 store)

pkg/core/options.go
```diff
@@ -76,6 +78,13 @@ func WithMetrics(enabled bool) Option {
 	}
 }
 
+// WithLabel toggles metrics for the core package and its dependencies (e.g. cafs)
```
fredbi: doc string probably results from a copy/paste

ransomw1c:
previous changes (core+storage) at `j2240-labelversions--wip-03-initialsketch` branch while the scope of this patch is lessened to `storage.VersionedStore`/`StoreVersioned`.

from conference:  version is passed to existing `storage.Store` as vararg using usual functional option pattern.

types for the option _and version_ will be added to `storage` – so object names are object names and object versions are object versions.  will not immediately match [`gsutil`'s naming convention](https://cloud.google.com/storage/docs/naming#objectnames).

pkg/core/options.go
ransomw1c: will attempt to make a note of this in the forthcoming change at layers after `storage`

pkg/core/label_list.go
ransomw1c: this was changed.

my mistake:  type assertions look like method chaining, but they aren't.

pkg/core/options.go
ransomw1c: ty for being permissive about prefs here.  i will push for consistency upon starting changes outside of `storage`.

ransomw1c:
dunno what's up with gh <> circ.  there's a pass on the 75e5222 tooltip/popover gh ui el, yet the inline summary box is indicating fail.

i'm toggling this as ready for review toward merge in any case.

cmd/metrics/cmd/mock_dest_store.go
```diff
@@ -69,7 +69,7 @@ func (mds *mockDestStore) Put(ctx context.Context, key string, source io.Reader,
 	return nil
 }
 
-func (mds *mockDestStore) Delete(ctx context.Context, key string) error {
+func (mds *mockDestStore) Delete(ctx context.Context, key string, opts ...storage.Option) error {
```
fredbi: Adding options to a Store should follow a different pattern.
Rather than cluttering the interface with `opts ...storage.Option`, this is made once in `New()`

hack/build_artifacts.sh
```diff
@@ -0,0 +1,44 @@
+#! /bin/zsh
+
+setopt ERR_EXIT
```
fredbi: ok so I am taking it that this is an experiment to emulate `go test $(go list ./...|grep pkg)`, isn't it?

ransomw1c: yes, this is cruft.

cmd/metrics/cmd/mock_dest_store.go
ransomw1c: resolved at cfae952 , currently in `j2240-labelversions--wip-05-rev` after squash and rebase of this branch.

note: cfae952 passes `go_test` while `fuse_sidecar_test` fails, hence the rebase and try-again.

ransomw1c:
@rbecker1c

rich, i've taken the initiative to add you to the reviewers' list in order to allow some measure of administration here.  it would still be most preferable to allow fred and ritesh's reviews.  i'm adding your review in the spirit of allowing more options rather than anything suggestive beyond that.

pkg/storage/gcs/store.go
```diff
@@ -39,12 +41,30 @@ func clientOpts(readOnly bool, credentialFile string) []option.ClientOption {
 	return opts
 }
 
-// New builds a new storage object from a bucket string
-func New(ctx context.Context, bucket string, credentialFile string, opts ...Option) (storage.Store, error) {
+// versionedObject wraps *BucketHandle.Object
+// towards canonicalization of object names.
+func versionedObject(
```
ransomw1c: from #429 , canonicalization of object names is not anywhere on the horizon and `versionObject`'s body need be inlined

kerneltime:
I am closing this PR for now. Given the realignment and what @ransomw1c has to focus on.

PR #421, opened 2020-03-04
Upgrade CI sizing

fredbi:
This PR upgrades the resource class used to run our tests.
The default instance ("medium") is often failing on memory shortage or time outs.
Upgrading to "medium+" beefs up a bit the workflows, at marginal cost in circleci credits.
Testing with "large" instances did not improve things since the next bottleneck is the latency to gcs (this story shall also be told).


fredbi:
Thanks @joaoh821C

PR #428, opened 2020-03-09
Refactored Dockerfiles and sidecar demo CI

fredbi:
This PR addresses several known issues with datamon sidecars:
1. No tests for Postgres sidecar
2. Tests for fuse sidecar frequently run forever and leave unrelinquished k8s resources
3. Very long CI runs for sidecar tests
4. Intricate parameterization of the Postgres sidecar

## Tests for Postgres sidecar
This brings some evidence that the Postgres sidecar is working.

I removed support for multiple database servers within a single container.
Conversely, I've added support for multiple such sidecars on the same pod.

NOTE: a sidecar exposes a database _server_  (i.e. a port). Each server may publish several logical databases.

## Forerever running CI tasks
This is often caused by several CI jobs running in parallel: sidecar test suite refers to one single image ("latest") and rebuilds it in place. Further, pod detection and log tracking are misled by the existence of another workflow doing the same thing on another branch.

This is fixed by identifying each k8s resource with a commit hash and no more pulling images by tag "latest".

I also added internal timeout checks and a trailing CI step to relinquish all created k8 resources.

I also wanted tests (Postgres or fuse sidecar) to use the _actually released_ images and not some derived image build during tests (these images used to achieve what we do now with config maps: baking config files in a container image).

All CI tests are now corraled into a dedicated namespace.

## CI build time
All docker files have been largely revamped or even deleted.

All docker files are now located in `./dockerfiles` or `./hack/fuse-demo` for the few ones that are specific to the demo. Following suit, the `Makefile` did undergo many small changes.

The outcome of this is greatly reduced repetition of install steps in Docker files and CI steps:
- we build base images for CI runs and sidecar builds every week
- most docker files refer to these base images (exception: the alpine build is not worth it)
- the CI workflow has  been modified: images used by the demo are now only built once and correspond to the ones that would be released by a merge
- we now release many stable docker images: datamon, migrate tool, fuse sidecar, Postgres sidecar, and the standalone wrapper bourne shell

**Before**
Workflow   datamon 1987 c8e2dc69 2991 47cb b56a a933cb1f5525

**After**
Workflow   datamon 2087 3d7e9ac9 488d 497a a0d4 42def8dcd213

**NOTE**: this has been painstakingly tested and benchmarked... I eventually resolved to construct a Ci executor out of the newest generation of CircleCI convenience images (cimg/go), which led to the smallest download times (many layers are often already cached).
 

## Postgres sidecar parameterization
I've dropped most of the parameterization code in this sidecar and replaced this by a simplistic yaml parser. Parameters, including the datamon config file, are declared as k8s config maps mounted into the container.

Read-only Postgres can be achieved by _not_ specifying a destination for upload: upload is just skipped. The database is not really read-only: the app may modify it, but the outcome is not saved.

It remains possible to use env vars with a straightforward prefixing rule. The `sidecar_param` binary is not used (I've tried to keep this, but it just yielded corrupted output).

## Code simplification
Such an in-depth walk-through the sidecar functionality was an opportunity to remove some unused code, reminiscent of past experiments and tests.

This gave me a lot of ideas about how to deal with the problem at hand in novel ways, but I refrained from a complete overhaul and essentially kept the original approach untouched.
## Other changes
* Postgres sidecar has migrated from zsh to bash. All bash scripts are checked by shellcheck. I've left the existing zsh for the fuse demo scripts
* fuse sidecar parameterization no more needs to generate param files - the go script to do just that has been removed
* I removed a lot of the "demo parameterization" stuff, which makes it hard to follow
* I replaced flaky, non-portable timestamp parsing utility by a go script

## TODO/Overlooked issues
* ~~need more testing/understanding of the role of tini (e.g. SIGTERM catching)~~  tini is not used
* need more testing to remove the need for projecting secrets
* ~~now that this part is clearer to me (and hopefully, to reviewers as just), I can start writing some "hands-on" documentation, taking the new pod template as a case study.~~ see PR #433
* ~~I don't know  what is this `datamover.sh` utility. It wasn't part of the tests and is still not. It is probably broken now with all that reshuffling around. Please advise.~~ removed with PR #432 
* ~~I don't know what are the `nfs_xxx` utilities about. They look like unfinished work. Please advise.~~ removed with PR #432
* continue simplification by removing all this bespoke "shell on the pod" thing; if a dev needs to run a shell in a pod, there is no need for all that
* binaries.Dockerfile build could be sped up further by transferring the content of the module cache to the building container
* Postgres test could go further by running another pod to consume the results saved by the first pod


kerneltime:
We can remove `datamover.sh` and nfs related scripts.. they have been replaced by a different process followed by ops.

fredbi:
> We can remove datamover.sh and nfs related scripts.. they have been replaced by a different process followed by ops.

OK. Will do that

.circleci/config.yml
```diff
@@ -490,8 +530,20 @@ workflows:
             branches:
               only: /.*/
 
+      - go_cross_compile:
+          context: "OC Common"
+          requires:
+            - go_lint
+          filters:
+            tags:
+              only: /.*/
```
ransomw1c: so this is effectively a no-op?

.circleci/config.yml
```diff
@@ -2,12 +2,16 @@ version: 2.1
...
+    # An executor to run S3 compatibility unit tests
     working_directory: ~/project
     environment:
       GOOGLE_COMPUTE_ZONE: us-west2-c
     resource_class: 'medium+'
     docker:
-      - image: circleci/golang
+      - image: gcr.io/onec-co/datamon-builder:latest
+        auth:
+          username: _json_key
```
ransomw1c: `_json_key` resolves to a value in Circle CI, `_json_key` as a value is passed to docker, or ... ?

dockerfiles/builder.Dockerfile
```diff
@@ -0,0 +1,25 @@
+# A base container to build datamon and run CI jobs
+#
+# This image is updated every week on our CI
```
ransomw1c: can there be a reference dockerfile responsible for the image  and the repo that contains it here, or for infosec reasons, would such a link better be added to Confluence somewhere?

one way or another, what i'm missing here is why this comment is just saying "such an image exists" rather than "such an image exists, and this is where it came from."

.circleci/config.yml
fredbi: @ransomw1c not quite. You're right that this change tries to keep things as much as they used to be, only faster and with less code. However, the devil is in the details and _a lot_ of small details changed, such as harmonizing released images (all debian-based, with a lot less built-in clutter), reformulating the build dependencies for sidecars, ...

.circleci/config.yml
fredbi: `_json_key` is a special tag used by `docker login` to interpret the password as an `OAuth2` token when connecting to the `gcr.io` login. If you are interested in that part, take a look at this repo for more details: https://github.com/ivanilves/lstags/issues/211

dockerfiles/builder.Dockerfile
fredbi: @ransomw1c that is a very good idea. Extracting this as a separate issue. I will proceed with writing more documentation, specifically for the docker part.

hack/k8s/flood-pipeline.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: extensions/v1beta1
```
kerneltime: This should not be here..

fredbi:
@ransomw1c Re your comments;

> the main thing I wanted to notice in the shell scripting is whether it's still long-polling or not – and it still is, which is okay

Yes, it is. I've added some timeout logic, though

 >  eventually, I'd like to use a different transport-layer mechanism for the eventing. Jake mentioned an existing tool for inter-OS-level application messaging

I'll push today a new design proposal for sidecars.
Yes, eventing should probably be supported by a different technology, either network-based (e.g. some lightweight embedded message queue) or IPC based - requires specific pod config, though). I've been thinking about that a lot, lately. I'd like to equip the app wrapper with some automatic discovery of the various containers to be coordinated.

> also, let me note here that I'm biting my tongue somewhat on the virtues of bash versus zsh.

And I thank you for your restrain. Indeed `zsh` and `bash` are very much equivalent.

The reason why I moved to bash is: since I had to rewrite a sizeable chunk of the Postgres bit, I chose the shell I am most familiar with. Simple as that.

My personal take on all this is, that _eventually_ (not now) all the shell stuff should go because it is simply ill-suited for what we want it to do (yaml parsing and event handling).

The big benefit I see from moving to `bash` is that it is supported by `shellcheck`. And `shellcheck` helps _a lot_ figure out the many quirks and code smells in shell scripts.
Notice that I've added a new make target to make an attempt at linting zsh _as bash_.

hack/k8s/flood-pipeline.yaml
fredbi: yeah I'd considered that too. I implicitly put that one together with the bag of unused files.

kerneltime:
I am about 40% done.. will continue tomm.

hack/k8s/flood-pipeline.yaml
fredbi: see #432

dockerfiles/builder.Dockerfile
fredbi: @ransomw1c done. See PR #435

fredbi:
@bluelion @joaoh821C you are welcome to a look at the screenshot above, re CI optimization (and the associated reduction of circleCI credits)

.circleci/config.yml
```diff
@@ -560,12 +625,24 @@ workflows:
...
+  weekly:
+    triggers:
+      - schedule:
+          cron: "5 4 * * 6"
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - push_base:
```
kerneltime: Might be better to build the base image only on merge into master instead of a cron job.

fredbi: Done as discussed

.circleci/config.yml
```diff
@@ -527,45 +624,88 @@ workflows:
...
+          requires:
+            - go_lint
+            - push_base
+          filters:
+            tags:
+              only: /.*/
+            branches:
+              only: /.*/
+
+      - push_base:
```
kerneltime: @fredbi If the base images' tag `latest` are based only on master, should it not be triggered only when the master is updated?

PR #429, opened 2020-03-18
core pkg label versions

ransomw1c:
apply gcs version functionality from #419 to datamon labels, specifically, at the `pkg/core` level of abstraction, the layer nearest to and distinct from the ui

### DIFF
--- a/hack/build_artifacts.sh
+++ b/hack/build_artifacts.sh
@@ -1,0 +1,44 @@
+#! /bin/zsh
+
+setopt ERR_EXIT
+setopt PIPE_FAIL
+
+perhaps_module_dirs=(pkg/)
+
+for perhaps_module_dir in $perhaps_module_dirs; do
+    find $perhaps_module_dir -type d |while read pkg_dir; do
+        if find $pkg_dir -maxdepth 1 \
+                |grep -q '_test.go$'; then
+            print -- "building $pkg_dir test binary"
+            (cd $pkg_dir \
+                 && go test -c)
+        fi
+    done
+done
+
+# ./deploy ???
+perhaps_main_root_dirs=(cmd/ internal/ hack/ pkg/)
+
+typeset -a mains
+
+for perhaps_main_root_dir in $perhaps_main_root_dirs; do
+    find $perhaps_main_root_dir -type d \
+        |while read perhaps_main_dir; do
+        (cd $perhaps_main_dir \
+             && find $ -type f \
+                 |while read perhaps_main; do
+                 if cat $perhaps_main \
+                         |grep -q '^package\s*main'; then
+                     mains=($mains $perhaps_main)
+                 fi
+                 if [[ $#mains -eq 1 ]]; then
+                     print -- "attempting $$perhaps_main build"
+                     go build
+                 else
+                     if [[ $#mains -eq 0 ]]; then continue; fi
+                     print -- "unsure how to build multiple mains" \
+                           "in same directory $perhaps_main_dir build"
+                 fi
+             done)
+    done
+done
--- a/pkg/core/label_list.go
+++ b/pkg/core/label_list.go
@@ -8,7 +8,7 @@ import (
 
 	context2 "github.com/oneconcern/datamon/pkg/context"
 	"github.com/oneconcern/datamon/pkg/core/status"
-
+	"github.com/oneconcern/datamon/pkg/errors"
 	"github.com/oneconcern/datamon/pkg/model"
 	"github.com/oneconcern/datamon/pkg/storage"
 )
@@ -21,7 +21,7 @@ const (
 func ListLabels(repo string, stores context2.Stores, prefix string, opts ...Option) ([]model.LabelDescriptor, error) {
 	labels := make(model.LabelDescriptors, 0, typicalLabelsNum)
 
-	labelsChan, workers := listLabelsChan(repo, stores, prefix, opts...)
+	labelsChan, workers := listLabelsChan(repo, stores, append(opts, WithPrefix(prefix))...)
 
 	// consume batches of ordered labels
 	err := doSelectLabels(labelsChan, func(labelBatch model.LabelDescriptors) {
@@ -52,17 +52,20 @@ func ListLabelsApply(repo string, store context2.Stores, apply ApplyLabelFunc, o
 		isVersionsList bool
 	)
 
+	if !apply.valid() {
+		return errors.New("invalid ApplyLabelFunc")
+	}
+
 	{
 		settings := defaultSettings()
 		for _, bApply := range opts {
 			bApply(&settings)
 		}
-		if settings.label == "" && settings.prefix == "" || settings.label != "" && settings.prefix != "" {
-			// ???
-			//			return errors.New("either prefix or label must be supplied")
+		if !settings.listLabelsValid() {
+			return errors.New("options translate to invalid settings")
 		}
 		prefix = settings.prefix
-		isVersionsList = settings.label != ""
+		isVersionsList = settings.vLabel != ""
 	}
 
 	labelChan := make(chan model.LabelDescriptor)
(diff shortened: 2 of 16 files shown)


pkg/storage/gcs/store.go
```diff
@@ -39,6 +39,15 @@ func clientOpts(readOnly bool, credentialFile string) []option.ClientOption {
 	return opts
 }
 
+// versionedObject wraps *BucketHandle.Object
+// towards canonicalization of object names.
+func versionedObject(
```
fredbi: @ransomw1c  when we last talked about this, we all agreed that this strategy was regarded as bringing too much change compared to what was really needed. TL, DR: just drop this and keep your original idea (adding KeyVersions only).

hack/build_artifacts.sh
```diff
@@ -0,0 +1,44 @@
+#! /bin/zsh
+
+setopt ERR_EXIT
+setopt PIPE_FAIL
+
+perhaps_module_dirs=(pkg/)
+
+for perhaps_module_dir in $perhaps_module_dirs; do
```
fredbi: @ransomw1c could you please introduce that one? what is this about? running tests on dependencies?
I am not even sure that `go mod download` picks up tests in dependency packages (since the introduction of the GO MOD PROXY feature, I'd say).

pkg/core/label_list.go
```diff
@@ -35,6 +36,7 @@ func ListLabels(repo string, stores context2.Stores, prefix string, opts ...Opti
 // ApplyLabelFunc is a function to be applied on a label
 type ApplyLabelFunc func(model.LabelDescriptor) error
 
+// todo: prefix to opts
```
fredbi: :+1:  Definitely a todo for another PR

pkg/core/label_list.go
```diff
@@ -133,7 +139,23 @@ func listLabelsChan(repo string, stores context2.Stores, prefix string, opts ...
...
+		if settings.label != "" {
+			return GetLabelStore(stores).KeysPrefix(context.Background(), next,
+				model.GetArchivePathPrefixToLabels(repo, prefix),
+				"", settings.batchSize)
+		} else {
+			// tbd: sandbox o.(type).method() type assert fail compared to o.(type).method() error
+			versionedStore := GetLabelStore(stores).(storage.StoreVersioned)
+			versions, versionsErr := versionedStore.KeyVersions(context.Background(), settings.label)
+			if versionsErr != nil {
+				return nil, "", versionsErr
```
fredbi:
```suggestion
				return nil, "", fmt.Errorf("a backend store that supports versionning is required when using this option: %w" versionsErr)
```

pkg/core/label_list.go
```diff
@@ -133,7 +139,23 @@ func listLabelsChan(repo string, stores context2.Stores, prefix string, opts ...
 	keysChan := make(chan keyBatchEvent, 1)
 
 	iterator := func(next string) ([]string, string, error) {
-		return GetLabelStore(stores).KeysPrefix(context.Background(), next, model.GetArchivePathPrefixToLabels(repo, prefix), "", settings.batchSize)
+		if settings.label != "" {
+			return GetLabelStore(stores).KeysPrefix(context.Background(), next,
+				model.GetArchivePathPrefixToLabels(repo, prefix),
+				"", settings.batchSize)
+		} else {
+			// tbd: sandbox o.(type).method() type assert fail compared to o.(type).method() error
```
fredbi: there is nothing "tbd" left here. No need to comment on it

pkg/core/label_list_test.go
```diff
@@ -157,3 +157,7 @@ func TestListLabels(t *testing.T) {
 		}
 	}
 }
+
+func TestListLabelVersions(t *testing.T) {
+	t.Skip("tbd")
```
fredbi: ok. So here is a todo, isn't it?

pkg/core/options.go
```diff
@@ -17,6 +17,8 @@ type Settings struct {
 	profilingEnabled bool
 	memProfDir       string
 
+	label string
```
fredbi: Not very important, but I am still thinking a more explicit name would help here. Some suggestions: `versionedLabels`, `withVersions`, `vLabel`...

pkg/core/options.go
```diff
@@ -76,6 +78,13 @@ func WithMetrics(enabled bool) Option {
 	}
 }
 
+// WithLabel toggles metrics for the core package and its dependencies (e.g. cafs)
```
fredbi: Comment comes from copy/paste

pkg/storage/gcs/store.go
```diff
@@ -282,6 +310,42 @@ func (g *gcs) KeysPrefix(ctx context.Context, pageToken string, prefix string, d
 	return
 }
 
+func (g *gcs) KeyVersions(ctx context.Context, key string) ([]storage.Version, error) {
```
fredbi: :+1: 
 
That is the one we need.

pkg/storage/gcs/store.go
```diff
@@ -226,9 +236,17 @@ func (g *gcs) PutCRC(ctx context.Context, objectName string, reader io.Reader, d
 	return toSentinelErrors(err)
 }
 
-func (g *gcs) Delete(ctx context.Context, objectName string) (err error) {
+func (g *gcs) Delete(ctx context.Context, objectName string, opts ...storage.Option) (err error) {
```
fredbi: Similarly, we discussed the point about adding options.

i) this is not needed to address the requirement
ii) we don't want to generalize support for versioning, actually quite the opposite (labels stand out as an exception)
iii) adding options to a Store follows a different pattern, injecting options with `New()` rather than with every method

ransomw1c: yes, all `storage/gcs` changes are discussed and resolved in #419

pkg/core/options.go
ransomw1c: noted and changed:  added several comments/godoc strings since this review.

hack/build_artifacts.sh
ransomw1c: > what is this about?

ensuring that the test binaries/executable build.

----

introduction for team members less familiar with golang is that (as you know) under the hood, `go test` builds an executable similar to `go build`, then runs that executable.  go, unlike python, is a typed and compiled language.  we get a reasonable amount of sanity checks by choosing appropriate types and making sure everything builds, runtime checks aside.

passing the `-c` flag to `go build` says, "build the test executable only, don't run it."  and `go test -c` can only build one executable at a time.  so this script is there to traverse all the test binaries and make sure everything is kosher thru compile time, stopping short of the runtime checks provided by `go test ./...` or somesuch.

pkg/storage/gcs/store.go
ransomw1c: i'll back out these changes momentarily.

it is an internal implementation detail that need be resolved before this request lands, not something that need block percolating finishing touches to ui.

pkg/core/label_list.go
ransomw1c: pretty sure there's a better way to chain errors than this ;)

pkg/core/label_list.go
ransomw1c: i intended to throw together an example to broadcast in Slack#engineering. not sure if there's yet time for that now.  that's what this comment was about, tho.

pkg/core/label_list_test.go
ransomw1c: this is the tail end of the wip here, yep

pkg/core/options.go
ransomw1c: :p just :p .  `vLabel` it is.

pkg/core/label_list.go
```diff
@@ -101,7 +146,22 @@ type labelEvent struct {
 // labelsEvent catches a collection of labels with possible retrieval error
 type labelsEvent struct {
 	labels model.LabelDescriptors
-	err    error
+	// fallback to string typing pending decision-making re. type system design.
```
ransomw1c: please give some consideration to the question in this comment

pkg/storage/gcs/store.go
ransomw1c: .. whoops, this and all `pkg/storage` changes probably belong in #419 .

i'll make the change in that patch, then, once it's landed, i'll be sure to catch on rebase to this branch.

kerneltime:
I am closing this PR for now. Given the realignment and what @ransomw1c has to focus on.

PR #433, opened 2020-03-18
Doc sidecar

fredbi:
* rewrote user doc on sidecars, essentially a walk-through the examples
* moved the existing material as "sidecar design", which reflects the current design -- remove pg parts that are no more relevant
* introduced a new "analysis" document (could also have called this "impediments"), preliminary to the forthcoming design proposal -- basically challenging https://oneconcern.atlassian.net/browse/RES-1826

NOTE to reviewers: since this builds up on top of other PR you may see the actual changes
by jumping to that commit https://github.com/oneconcern/datamon/pull/433/commits/e94af91c77817294442c8cfbf34db8b518c46d9f

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -241,8 +241,11 @@ commands:
       - run:
           name: Add CircleCI IP to whitelist
           command: |
-            CIRCLE_CI_IP=$(curl -sSL https://checkip.amazonaws.com)
-            gcloud container clusters update --zone ${GOOGLE_COMPUTE_ZONE} ${CLUSTER} --enable-master-authorized-networks --master-authorized-networks ${ALLOWED_NETWORKS},$CIRCLE_CI_IP/32
+            CIRCLE_CI_IP="$(curl -sSL https://checkip.amazonaws.com)"
+            ALREADY="$(gcloud container clusters describe --zone "${GOOGLE_COMPUTE_ZONE}" --format json "${CLUSTER}"|jq -c '{net: .masterAuthorizedNetworksConfig.cidrBlocks}|.net|map(.cidrBlock)|map(select(. == "'"${CIRCLE_CI_IP}"'/32"))|.[]')"
+            if [ -z "${ALREADY}" ] ; then
+              gcloud container clusters update --zone ${GOOGLE_COMPUTE_ZONE} ${CLUSTER} --enable-master-authorized-networks --master-authorized-networks ${ALLOWED_NETWORKS},$CIRCLE_CI_IP/32
+            fi
 
   remove_whitelist_ip:
     steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -560,12 +625,24 @@ workflows:
...
+            - go_test_metrics
             - build_images
           filters:
             tags:
               only: /^v.*/
             branches:
               ignore: /.*/
+
+  weekly:
+    triggers:
```
kerneltime: What's the motivation for a cron job?

docs/sidecar-design.md
```diff
@@ -0,0 +1,184 @@
...
+
+<!-- Ideas to improve this design may be found [here](proposals/sidecar-design.md) -->
+
+## Design
+
+The current design favors the sidecar container approach, with bespoke signaling between containers,
+over the CSI driver approach. Therefore,  datamon is not available as a kubernetes persistent volume plugin.
+
+This design choice stems from the current inability by Kubernetes CSI API to handle ephemeral volumes.
+Managing many short lived kubernetes volumes is at the moment not practical.
```
kerneltime: This is specific to GKE, CSI should have the support for it shortly but is far away from GKE.

Makefile
```diff
@@ -63,103 +92,122 @@ push-target:
...
+		upx ${TARGET}/*_linux_* ; \
+	else \
+	  echo "WARN:: upx not available in this environment. Binaries not compressed." ; \
+	fi
+
+.PHONY: build-and-push-datamon-builder
+## BUILD: Build new base builder image for CI jobs
+build-and-push-datamon-builder:
+	$(BUILDER) -t $(REPOSITORY)/datamon-builder:${BASEVERSION} - < dockerfiles/builder.Dockerfile
+	docker tag $(REPOSITORY)/datamon-builder:${BASEVERSION} $(REPOSITORY)/datamon-builder:latest
```
kerneltime: How does this work with parallel builds? Who gets the latest tag? Since this is based on branch, should the tag be the branch or hash of the commit? For the release, we can have a check for the tag attached to the commit?

.circleci/config.yml
fredbi: as discussed, commented this part

Makefile
fredbi: this step was not intended to be run during PR runs, but by cronjob only and from master only.
the content of this image does not depend on the repo

docs/sidecar-design.md
fredbi: Indeed you are correct. This part was added some time ago. Now updated to refer to possible future CSI implementation, when k8s v1.16 is available

kerneltime:
Needs rebasing

fredbi:
Rebased to current master

PR #435, opened 2020-03-19
Doc dockerfiles

### LINKED ISSUE #431: can there be a reference dockerfile responsible for the image  and the repo that contains it here, or for infosec reasons, would such a link better be added to Confluence somewhere?
can there be a reference dockerfile responsible for the image  and the repo that contains it here, or for infosec reasons, would such a link better be added to Confluence somewhere?

one way or another, what i'm missing here is why this comment is just saying "such an image exists" rather than "such an image exists, and this is where it came from."

_Originally posted by @ransomw1c in https://github.com/oneconcern/datamon/pull/428_

fredbi:
* fixes #431

NOTE to reviewers: please jump at commit https://github.com/oneconcern/datamon/pull/435/commits/ecef099224dc5c18f92176618166ddb88f7850b8 to see changes relative to #428 and friends which this PR is based upon

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -241,8 +241,11 @@ commands:
       - run:
           name: Add CircleCI IP to whitelist
           command: |
-            CIRCLE_CI_IP=$(curl -sSL https://checkip.amazonaws.com)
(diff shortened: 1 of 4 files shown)


kerneltime:
This needs rebasing. Will refrain from doing it online :-)

fredbi:
yes sure. I need to fix the whitelisting thing first: this is causing a lot of CI failures

fredbi:
Rebased to current master

fredbi:
@ransomw1c thanks

PR #436, opened 2020-03-25
mod tidy

ransomw1c:
in #419 , i'm noticing that some of `./.githooks/pre-commit/` aren't idempotent on `master` like i'd expect.  this is part of an effort to push some of those changes thu ci.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -205,14 +205,14 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - pkg-cache-{{ checksum "go.sum" }}
-            - pkg-cache-
+            - go-mod-v1-{{ checksum "go.mod" }}
+            - go-mod-v1-
 
   put_cache:
     description: save golang modules cache
     steps:
       - save_cache:
-          key: pkg-cache-{{ checksum "go.sum" }}
+          key: go-mod-v1-{{ checksum "go.mod" }}
           paths:
             - "/home/circleci/go/pkg"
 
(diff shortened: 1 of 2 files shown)


ransomw1c:
aaand ci is failing in this branch at the "Restoring Cache" step of `go_lint`.

i've toggled draft mode off to see if that makes a difference before attempting re-run.

ransomw1c:
if i'm understanding correctly, according to [this](https://circleci.com/docs/2.0/caching/) circCI documentation, circCI caching attaches to environment-specific package managers

> Caching is particularly useful with package dependency managers such as Yarn, Bundler, or Pip

so what was probably happening was my local golang install getting out of sync with circCI.  i've tried upgrading to see if that resolves.

ransomw1c:
a more thorough read of the above linked circCI [document](https://circleci.com/docs/2.0/caching/#clearing-cache) suggests using a version to allow clearing the cache, and doing so has resolved the "Restore cache" error on `go_lint` mentioned above.

**the version is that of the cache, not that of the [golang] package manager**

furthermore, cross-correlating [circCI docs](https://circleci.com/docs/2.0/caching/#using-a-lock-file) on caching directly with the [golang wiki](https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using) suggests that [this](https://circleci.com/blog/go-v1.11-modules-and-circleci/) blog post might be in error in suggesting `go.sum` instead of `go.mod` to compute the checksum for circCI caching.

specifically, `go.sum` is about security, not reproducibility.  when we talk about caching, we're talking about reproducibility, i think.

PR #438, opened 2020-03-25
Fix whitelisting IP race conditions

fredbi:
* Running k8s jobs in parallel (sidecar demo jobs) was fine until I added the IP whitelisting
  change to abide by our network access restrictions. This triggers are race condition on these
  2 workflows and sometimes, one fail because it cannot access our cluster.


fredbi:
@kerneltime I think this is the one to address first.
Otherwise, all our rebuilds will sometimes fail because of this IP stuff

PR #447, opened 2020-08-10
Backport delete repo

fredbi:


kerneltime:
As long delete permission is for privileged used on the backend.

PR #450, opened 2020-10-09
feat(auth): add option to skip auth

fredbi:
Enforcing authentication has proved to hamper adoption of datamon.

Besides, lately, google has introduced several breaking changes with their auth, so it is difficult for us to support users with different versions of the gcloud SDK.

In a final analysis, the auth feature is not really used when we want to run datamon from a pod, which is our main use-case.

Reminder: the auth feature was introduced to enforce lineage tracking, not security. Security remains enforced by gcs credentials.

This PR introduces a CLI option to skip auth. Username and email can be fed by config, if the user (or job) is willing to sign.


aakarshgupta97 (approved):
LGTM as long as it's been tested!

PR #457, opened 2021-05-14
fix: fix migration chart

jakedsouza:


k8s/migratev2/templates/staging-claim.yaml
```diff
@@ -4,8 +4,6 @@ metadata:
   name: "{{.Release.Name}}-staging-claim"
 spec:
   storageClassName: {{ .Values.storageClassName }}
-  persistentVolumeReclaimPolicy: Delete
-  finalizers: []
```
fredbi, line 8: yep that was a hack to force the volume to be relinquished. Was never quite successful with that (k8s bug, might be solved now)

k8s/migratev2/templates/staging-claim.yaml
```diff
@@ -4,8 +4,6 @@ metadata:
   name: "{{.Release.Name}}-staging-claim"
 spec:
   storageClassName: {{ .Values.storageClassName }}
-  persistentVolumeReclaimPolicy: Delete
```
fredbi, line 7: this one should not be deprecated. Why do you want to remove this?

k8s/migratev2/templates/job.yaml
```diff
@@ -78,7 +78,6 @@ spec:
             name: config2
           - mountPath: {{ dir .Values.credentials }}
             name: creds-secret
-            key: {{ base .Values.credentials }}
```
fredbi, line 81: this provides the key in the secret file... How can you get rid of this?
(hint: if you want to use the service account, then there are no credentials  at all, but we need a service account that knows how to  deal with gcs buckets)

k8s/migratev2/templates/staging.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
   type: .Values.storageType
 spec:
   storageClassName: {{ .Values.storageClassName }}
-  persistentVolumeReclaimPolicy: Delete
```
fredbi, line 9: this should be the default, that's right. I wanted to make it explicit that we relinquish the saving area back to the cluster after we're done.
the actual behavior is pretty  much provider-dependent, hence the explicit parameter here.
See also https://kubernetes.io/docs/concepts/storage/persistent-volumes/#delete

jakedsouza: reverted

k8s/migratev2/templates/job.yaml
jakedsouza:
```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Job.spec.template.spec.containers[0].volumeMounts[3]): unknown field "key" in io.k8s.api.core.v1.VolumeMount
```
Without it, k8s mounts all keys in the secret. I think the key param is for Volumes and not VolumeMounts.

k8s/migratev2/templates/staging-claim.yaml
jakedsouza: reverted

k8s/migratev2/templates/staging-claim.yaml
jakedsouza:
```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(PersistentVolumeClaim.spec): unknown field "finalizers" in io.k8s.api.core.v1.PersistentVolumeClaimSpec, ValidationError(PersistentVolumeClaim.spec): unknown field "persistentVolumeReclaimPolicy" in io.k8s.api.core.v1.PersistentVolumeClaimSpec]
```

fredbi (approved):
Now we have to refresh a bit the CI...

fredbi:
@jakedsouza  I've carried out some dust-off on our CI. I gave up the long & complicated runs against k8s.
Now everything runs smooth again, you may resume this line of work. Apologies for the delay...

Cheers

FRED

